PR #2, opened 2015-01-28
Use setuptools when available.

ocefpaf:
This PR makes use of setuptools, when available, allowing for easy Windows builds via
[Microsoft Visual C++ Compiler for Python 2.7](http://www.microsoft.com/en-us/download/details.aspx?id=44266) as requested by @rsignell-usgs.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,28 +1,35 @@
 import sys, os, glob, subprocess
-from distutils import ccompiler, sysconfig
-from distutils.core import setup, Extension
+
+try:
+    from setuptools import setup, Extension
+    use_setuptools = True
+except ImportError:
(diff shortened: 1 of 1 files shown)


rsignell-usgs:
I verified that this works.  Super cool.   Finally, a lightweight way to build python binaries on windows, and from Microsoft!    Setuptools >6.0 can figure out where the compiler is, but distutils cannot.

rsignell-usgs:
@jswhit  poke?

ocefpaf:
@jswhit Let me know if you want me to close this or re-write in a clearer way.

ocefpaf:
Closing this...

Issue #9, opened 2015-04-20
pip install fails


stdavis:
When I try to install via pip I get:
![image](https://cloud.githubusercontent.com/assets/1326248/7237297/d06b80a0-e757-11e4-8ef8-303a8ae58d96.png)

I get a similar error when I try to run `python setup.py build` after manually downloading the project.

What am I doing wrong?

micahcochran:
First, try installing [Microsoft Visual C++ Compiler for Python 2.7](http://www.microsoft.com/en-us/download/details.aspx?id=44266).

If that doesn't work, I'd suggest installing a different python environment such as [OSGeo4W](http://wiki.osgeo.org/wiki/OSGeo4W) which can easily install the Proj.4 library from its setup and the gcc compiler.  It is focused on software for GIS.

ChrisBarker-NOAA:
Note:  if you use the "Microsoft Visual C++ Compiler for Python 2.7.", then ``pip install`` should work. but plain old ``setup.py build`` probably won't. That compmiler needs setuptools, which the setup.py does not use, but pip monkey-patches in. I'll be raising an issue / pull request on this soon.

Also - this is using ArcGIS's Python -- is that what you want? -- just making sure. And it may not sork right with that compiler, either :-(

snowman2:
Related to #216

PR #21, opened 2015-10-02
Add setuptools to setup.py

ChrisBarker-NOAA:
I suggest using setuptools, rather than raw distutils, in the setup.py.

setuptools is pretty much ubiquitous these days, so it's not likely to be a problem. And pip requires setuptools, so anyone doing ``pip install`` is guaranteed to have it.

But it's a better option, particularly on Windows, where recent setuptools can properly find the "MS VS compiler for python2.7", whereas distutils cannot.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,6 +1,6 @@
 import sys, os, glob, subprocess
 from distutils import ccompiler, sysconfig
-from distutils.core import setup, Extension
+from setuptools import setup, Extension
 
 deps = glob.glob('src/*.c')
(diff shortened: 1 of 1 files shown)


micahcochran:
Patch for issue #9

ChrisBarker-NOAA:
Thanks Jeff!

CHB

Sent from my iPhone

On Oct 7, 2015, at 7:59 PM, Jeff Whitaker <notifications@github.com> wrote:

Merged #21 <https://github.com/jswhit/pyproj/pull/21>.

—
Reply to this email directly or view it on GitHub
<https://github.com/jswhit/pyproj/pull/21#event-429853802>.

Issue #22, opened 2015-10-05
Definition parameter +no_defs crashes pyproj


ReallyNiceGuy:
Any definition parameter without an '=' will actually crash the logic on __init__.py, line 698:
  File "/usr/local/lib/python2.7/dist-packages/pyproj/__init__.py", line 698, in __new__
    k,v = kvpair.split('=')

The definition I used was: +proj=merc +a=6378137 +b=6378137 +lat_ts=0.0 +units=m +nadgrids=@null +no_defs +x_0=5113560.997131 +y_0=2662158.879021

micahcochran:
I'm running pyproj 1.9.4.

```python
from pyproj import Proj, transform

p_grid = Proj('+proj=merc +a=6378137 +b=6378137 +lat_ts=0.0 +units=m +nadgrids=@null +no_defs +x_0=5113560.997131 +y_0=2662158.879021')
p_wgs = Proj('+init=epsg:4326')
transform(p_wgs, p_grid, -87.6, 34.5, 0)

>>>  (-4626894.447280438, 6758297.919468232, 0.0)
```
I tried transforming a point to that projection and it gave me output and didn't crash.  I couldn't replicate this.

Could you give a code sample that demonstrates the crash?

ReallyNiceGuy:
import pyproj
g=pyproj.Geod("+proj=merc +a=6378137 +b=6378137 +lat_ts=0.0 +units=m
+nadgrids=@null +no_defs +x_0=5113560.997131 +y_0=2662158.879021")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python2.7/dist-packages/pyproj/__init__.py", line
697, in __new__
    k,v = kvpair.split('=')
ValueError: need more than 1 value to unpack


Version 1.8.9, but the crashing code is the same

micahcochran:
Using that Proj.4 definition in `Geod()` on version 1.9.4 raises a ValueError exception.

With the exception of +a= and +b= parameters, everything else in that string is nonsensical input for `Geod()`.  What should pyproj when given input that is a valid Proj.4 string, but not valid for defining an ellipse?  My hunch is that it should ignore the invalid parameter, which is what it seems to do for all the other Proj.4 parameters.

ReallyNiceGuy:
I guess that the library already ignores it. We should check if the
parameter contains an '=' before splitting it. I don't know what to do with
the parameter after that, as it expects a pair for every one.

[...]

PR #23, opened 2015-10-06
Don't pass valueless parameters to Geod

ReallyNiceGuy:
As the code uses a dict to pass parameters, it tries to split
every parameter into name and value. If a valueless parameter
is given, ex: +no_defs, it breaks the split.
We can safely ignore any parameter that is not +a and +b, so
the easiest workaround for this issue is to just ignore the
problematic parameter


micahcochran:
This is fix for Issue #22 

:+1:

PR #24, opened 2015-10-09
Some unit testing fixes

micahcochran:
These commits:
1. Add a unit test for passing +no_def to Geod()  -> Pull #23 (already committed) which was reported in Issue #22
2. Add README.md documentation about new unit tests.
3. Fix for a type equals string bug in `Proj.__new__()` that was causing unittest testProjAwips221 to fail.  More info in commit message "String type test fix". (The way I did this requires the six package.)

This adds six as a required package, which may not be worth it for the single line that was fixed.  I am willing to redo this to support Python 2 & 3 without six, if preferred.  Requiring six will make Python 2.6 or greater a requirement.  There should probably be some discussion about requiring six.


micahcochran:
If i remove the requirement for six to fix  (3), would you be more apt to commit this patch?

jswhit:
yes, please remove the requirement for six

micahcochran:
I'm going to close this PR and split it into two branches "unit-testing-b" and "string-type-test" and submit 2 PRs.

PR #28, opened 2015-11-12
fix a division that is no longer accaptable for cython 0.23 while 0.2…

jdkloe:
Hi Jeff,

today I found this tiny little issue in the _proj.pyx file that actually caused the pyproj package to fail its build on fedora rawhide. It gives this error:

Error compiling Cython file:
------------------------------------------------------------
...
        raise RuntimeError('x,y and z must be same size')
    xx = <double *>xdata
    yy = <double *>ydata
    if inz is not None:
        zz = <double *>zdata
    npts = buflenx/8
                 ^
------------------------------------------------------------

This pull request fixes this division which is no longer accaptable for cython 0.23 while 0.22 still accpted it (It should be an integer division).

### DIFF
--- a/_proj.pyx
+++ b/_proj.pyx
@@ -340,7 +340,7 @@ def _transform(Proj p1, Proj p2, inx, iny, inz, radians):
     yy = <double *>ydata
     if inz is not None:
         zz = <double *>zdata
-    npts = buflenx/8
+    npts = buflenx//8
     if not radians and p1.is_latlong():
         for i from 0 <= i < npts:
             xx[i] = xx[i]*_dg2rad


jswhit:
Thanks!

jdkloe:
I'm happy I could contribute.
One related question remains though. The same file is also present in the basemap project (basemap/src/_proj.pyx), so I trust you'll update that one as well?
Do you have any plans to remove the (py)proj part from basemap and let pyproj be a dependency of basemap? As you may know the Fedora  project tries to minimize multiple copies of the same library and may ask me to remove this bundled copy in basemap. Are there any fundamental differences between the 2 copies that I should be aware of, or is it safe to remove the copy in basemap and use the standalone pyproj in stead?

micahcochran:
> Do you have any plans to remove the (py)proj part from basemap and let pyproj be a dependency of basemap? 
:+1: to that idea.   It looks like pyshp is a similar situation.

jdkloe:
Any news on the bunding issue?
or should I rather reopen that discussion as an issue on the basemap repository?

micahcochran:
Please open an issue in the basemap repository regarding this.

PR #30, opened 2015-11-13
Fix for string type test

micahcochran:
This line in `Proj.__new__()` to test if `projparms` is a string was always true because `or unicode` statement short circuited the logic. This would cause projparms as a dictionary to be passed to projstring.  That caused the projstring.count() to fail, because dictionaries don't have that function.

I recoded this so that the six package is NOT needed.

This problem was discovered because the unittest testProjAwips221 was failing.


micahcochran:
Thanks.

PR #39, opened 2015-12-11
update proj4 source code and data to latest github master

jswhit:
add PJ_hammer.c.diff patch for inverse hammer transform.


jswhit:
PJ_hammer.c now included in proj4 git master, so building against a pre-installed proj4 lib is no longer a problem.

micahcochran:
Thank you!  It looks like this should be released in the v4.9.3 release of proj4.

Issue #43, opened 2016-01-14
segmentation fault in test run after building without bundled proj sources


jdkloe:
I encountered an issue for the case when the pyproj module is build without the bundled proj sources. In this case the test run: python -c 'import pyproj;pyproj.test()'
ends with an ugly segmentation fault.

After some digging around I found that this is caused by the fact that my system (Fedora 23) has proj version 4.9.1 which does not have the inverse function for the hammer projection available. The same happens on Fedora rawhide(24) which has proj version 4.9.2
The segmentation fault happens at the call:
    projlonlatout = pj_inv(projxyin,self.projpj)
on line 204 of the file _proj.pyx
The root cause is the fact that the inverse function is called through a pointer in the self.projpj object, and the cython code does not check if this function pointer is defined or not (actually I don't even know if this is possible inside cython, maybe a little additional c-code is needed to do this).

Anyway, it would be nice if some safety check coul;d be added to pyproj to let it safely run for older external proj versions. Of course the problem will go away when proj 4.9.3 is released, as mentioned in issue #19 but I have no idea when that will happen.

jdkloe:
I have disabled the inverse hammer projection test in the rpm build for fedora, and after this all remaining tests run fine for rawhide (upcoming fedora 24). This one will have proj version 4.9.2
For fedora 23, which has proj version 4.9.1, there are 2 failing tests related to the hammer projection (both for python2 and python3). See the details in this log file:
https://kojipkgs.fedoraproject.org//work/tasks/1374/12551374/build.log
I guess this also is related to the missing inverse hammer in these proj versions.

The output for the failing tests is:
Trying:
    p = Proj(proj='hammer') # hammer proj and inverse
Expecting nothing
**********************************************************************
File "/builddir/build/BUILDROOT/pyproj-1.9.5.1-1.fc23.x86_64/usr/lib64/python3.4/site-packages/pyproj/__init__.py", line 324, in pyproj.Proj.__new__
Failed example:
    p = Proj(proj='hammer') # hammer proj and inverse
Exception raised:
    Traceback (most recent call last):
      File "/usr/lib64/python3.4/doctest.py", line 1318, in __run
        compileflags, 1), test.globs)
      File "<doctest pyproj.Proj.__new__[20]>", line 1, in <module>
        p = Proj(proj='hammer') # hammer proj and inverse
[...]

PR #44, opened 2016-01-16
fix segmentation fault

jdkloe:
add a check to ensure inverse projection is defined, before calling it (needed to prevent segfaults for proj versions 4.9.2 and before).

This solves issue #43 on my system. The inverse call now throws a proper python exception if the inverse projection is not defined, in the case that the module is build without bundled proj sources.

### DIFF
--- a/_proj.pyx
+++ b/_proj.pyx
@@ -26,6 +26,15 @@ cdef extern from "geodesic.h":
                double lat1, double lon1, double lat2, double lon2,\
                double* ps12, double* pazi1, double* pazi2)
 
+# define part of the struct PJconsts from projects.h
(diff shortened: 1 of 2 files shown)


micahcochran:
Would you mind adding a unit test for this?

jdkloe:
Sure, I just added some test code that shows the segmentation fault for the unpatched pyproj version when build with unbundled proj version 4.9.1 or 4.9.2. When the patched _proj.pyx from this pull request is applied all tests pass. Currently it does 135 tests, 1 for each projection, but if this is too much, its easy to do all tests in a single test as well.

micahcochran:
Thanks!

PR #45, opened 2016-01-17
PROJ.4 project appears to have moved

stestagg:
The links in this file point to the old trac server, whereas PROJ.4 is now maintained on github


micahcochran:
The HTML files are generated from the source code using [epydoc](http://epydoc.sourceforge.net/).  The best way to accomplish this is to replace all of the trac hyperlinks in `lib/pyproj/__init__.py`.  The docs should be regenerated during the next release.

jswhit:
I've fixed those URLs, and a few others, in github master.

micahcochran:
Thanks @stestagg and @jswhit

PR #49, opened 2016-01-27
setup.py: Parse version from _proj.pyx

micahcochran:
This makes _proj.pyx have the only copy of `__version__`.  The with statement requires Python 2.5.  If desired, I'll refactor the code for Python 2.4 support.  My intent is to remove one step from the process of updating the version number.

Here's what pip outputs:
```
C:\Users\mcochran\github\pyproj>pip install .
Processing c:\users\mcochran\github\pyproj
Installing collected packages: pyproj
  Running setup.py install for pyproj ... done
Successfully installed pyproj-1.9.5.1
```

Looks like pip recognizes the version number.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -86,15 +86,14 @@ else:
     datafiles = [os.path.join('data',os.path.basename(f)) for f in datafiles]
     package_data = {'pyproj':datafiles}
 
-# store pyproj version information (from _proj.pyx) in version variable
+# retreive pyproj version information (stored in _proj.pyx) in version variable
 # (taken from Fiona)
 with open('_proj.pyx', 'r') as f:
     for line in f:
         if line.find("__version__") >= 0:
-            version = line.split("=")[1].strip()
-            version = version.strip('"')
-            version = version.strip("'")
-            continue
+            # parse __version__ and remove surrounding " or '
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -85,12 +86,21 @@
...
     package_data = {'pyproj':datafiles}
 
+# store pyproj version information (from _proj.pyx) in version variable
+# (taken from Fiona)
+with open('_proj.pyx', 'r') as f:
+    for line in f:
+        if line.find("__version__") >= 0:
+            version = line.split("=")[1].strip()
+            version = version.strip('"')
+            version = version.strip("'")
```
ocefpaf: Instead of calling strip twice you can do `version = version.strip()[1:-1]`.  Also, I'd rather use `break` than continue, you do not need to keep looking the rest of the file :wink:

:+1: BTW

micahcochran: Thanks for reviewing.  Good suggestions.  

I must admit `version = line.split("=")[1].strip()[1:-1]` is a bit on the cryptic side.  I guess we can say that there is magic inside.  :tophat: :rabbit2:

I'll revise once jswhit comments. Thanks.

micahcochran:
Thank you.

PR #51, opened 2016-02-01
Travis-CI: Add Python version 2.6 to testing

micahcochran:
Had to install unittest2 for Python 2.6 in order to support tests in the tests in class TypeError_Transform_Issue8_Test.  `from __future__ import with_statement` was added for Python 2.5 support.


micahcochran:
Thank you.

PR #54, opened 2016-02-10
Change Proj.to_latlong in order to return an instance of pyproj.Proj

georgeouzou:
Returning a **pyproj.Proj** instance instead of the underlying base class **_proj.Proj** in **pyproj.Proj.to_latlong** function feels more intuitive.   
This patch adresses a bug as well in pyproj.transform's function check. See [Issue #53 ](https://github.com/jswhit/pyproj/issues/53)

Issue #53, opened 2016-02-09
Type error (used to work):


citibeth:
Consider the following code:
```
	xyproj = pyproj.Proj(sproj)
	llproj = xyproj.to_latlong()
	print(type(xyproj))
	print(type(llproj))
	ondata, latdata = pyproj.transform(self.xyproj, self.llproj, xdata, ydata)
```

This used to work.  But now, it gives me the following:

```
<class 'pyproj.Proj'>
<class '_proj.Proj'>
Traceback (most recent call last):
  File "plot_grid_outlines.py", line 27, in <module>
    grid1.plot(basemap)
  File "/Users/rpfische/git/icebin/pylib/icebin.py", line 125, in plot
    londata, latdata = pyproj.transform(self.xyproj, self.llproj, xdata, ydata)
  File "/Users/rpfische/macports/mpgompi-4.9.3/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/site-packages/pyproj/__init__.py", line 494, in transform
    raise TypeError("p2 must be a Proj class")
TypeError: p2 must be a Proj class
```
Why is llproj not of the same type as xyproj?  And why is transform() now giving an error?

Thanks,
-- Elizabeth

micahcochran:
I submitted a patch a couple month ago that apparently tightens typechecking down a bit too much.  _proj.Proj is the Cython code which communicates with the C Proj.4 library.  I think there is enough information for the transform() to be able to run using either class.

The way to quickly fix this is to comment out these lines (491-494) under the transform function.
```Python
    # check that p1 and p2 are from the Proj class
    if not isinstance(p1, Proj):
        raise TypeError("p1 must be a Proj class")
    if not isinstance(p2, Proj):
        raise TypeError("p2 must be a Proj class")
```

[If you can, test this patch](https://github.com/micahcochran/pyproj/tree/typechecking) where I think I have fixed the problem.

Before I submit a pull request, I would like to make a unit test out of this, so that this doesn't get broken again in the future.  The code that you gave has a couple of undefined variables, but I got the gist of it. @citibeth Can you help with creating a sample for a unit test for the problem?

georgeouzou:
@micahcochran i think you can check only for the base class in the patch like this:
 ``` 
    if not isinstance(p1, _proj.Proj):
        raise TypeError("p1 must be a Proj class")
    if not isinstance(p2, _proj.Proj):
        raise TypeError("p2 must be a Proj class")
```
[...]

PR #56, opened 2016-02-11
Unittest for latlong() TypeError

micahcochran:
This pull request should fail.

This is a unittest for issue #53, which latlong() function gives a _proj.Proj which is not a Proj object.  Pull request #54 resolves that issue.  This should be merged after PR #54, so that it will pass on Travis testing.


georgeouzou:
I have just included the fix in a fresh #72, that continues on your changes. :)

micahcochran:
Thanks George!

citibeth:
I believe this PR should be closed, since #73 supersedes it.  Then we can focus on getting #73 merged.

Issue #64, opened 2016-03-25
BUG: Geod shared memory between objects


micahcochran:
I was working on adding an \__eq__ operator (using the \__repr__ function in PR #63) for Geod, and found a shared memory bug.

For example (this is without the \__repr__ PR patch applied):
```Python
In [1]: from pyproj import Geod, pj_ellps

In [2]: pj_ellps['lerch']
Out[2]: {'a': 6378139.0, 'description': 'Lerch 1979', 'rf': 298.257}

In [3]: pj_ellps['clrk66']
Out[3]: {'a': 6378206.4, 'b': 6356583.8, 'description': 'Clarke 1866'}

# these ellipsoids are not equal
In [4]: pj_ellps['lerch']['a'] == pj_ellps['clrk66']['a']
Out[4]: False

In [5]: glerch = Geod(ellps='lerch')

In [6]: glerch.a
Out[6]: 6378139.0

In [7]: ga = glerch.a # store this

In [8]: repr(glerch)
Out[8]: '<pyproj.Geod object at 0x0208E090>'

In [9]: gclrk = Geod(ellps='clrk66')

In [10]: gclrk.a
Out[10]: 6378206.4

In [11]: repr(gclrk)
Out[11]: '<pyproj.Geod object at 0x0208E258>'

In [12]: ga == glerch.a  # compare those from #7, these should match
Out[12]: False

In [13]: glerch.a
Out[13]: 6378206.4  # this should be  6378139.0

In [14]: repr(glerch)
Out[14]: '<pyproj.Geod object at 0x0208E090>'

In [15]: repr(gclrk)
Out[15]: '<pyproj.Geod object at 0x0208E258>'

In [16]: glerch.a == gclrk.a  # should not match, see #4
Out[16]: True
```

The `.initstring` variable is unaffected by this.

There is some memory sharing happening in the Geod class between different instances.

One way to resolve this is just documenting that there can only be one Geod object.  I think it is preferable to rewriting this to allow for multiple Geod objects that do not share memory.

I tested pyproj version 1.9.5.1 and 1.9.3, both have the problem.  I'm planning to do a little more investigation.

micahcochran:
This bug does not seem to have an impact the distance geodesic calculations.

IPython console:
```Python
In [1]: >>> from pyproj import Geod

In [2]: >>> g = Geod(ellps='clrk66') # Use Clarke 1966 ellipsoid.

In [3]: g.a
Out[3]: 6378206.4

[...]

PR #65, opened 2016-03-25
Fix shared memory bug issue #64

micahcochran:
Fix for issue #64.  Geod.\__new__ rename self to cls as per [Python Docs](https://docs.python.org/2/reference/datamodel.html#object.\__new__). Move variables assigned in Geod.\__new__ to pass parameters of Cython Geod.\__cinit__. Unit tests added.

I'm not 100% sure why this patch works.  From my limited reading, I suspect that at \__new__ it is not quite a Python object.  It isn't until \__cinit__ that it can set objects.  As a sidenote, Proj.\__new__ does not store any "self" variables, so it doesn't have this problem.

Running the unittest/test.py in version 1.9.5.1 results in this and probably fails many of the asserts in the test_not_shared_memory function:
```
C:\Users\mcochran\github\pyproj\unittest>python test.py
................................................................................
..........................................................F...
======================================================================
FAIL: test_not_shared_memory (__main__.GeodSharedMemoryBugTestIssue64)
----------------------------------------------------------------------
  File "test.py", line 107, in test_not_shared_memory
    self.assertEqual(self.ga, self.g.a)
AssertionError: 6378206.4 != 2439700

----------------------------------------------------------------------
Ran 142 tests in 0.026s

FAILED (failures=1)
```


jswhit:
this pull request has conflicts with master that need to be resolved before it can be merged

micahcochran:
I had to run Cython to regenerate _proj.c, which was causing the conflict.

jswhit:
Not sure why this is needed, but I'll go ahead and merge

PR #50, opened 2016-01-31
Add pyproj.itransform function

georgeouzou:
An added function "**itransform**" similar to pyproj.transform, that enables transformation using generators and iterators. I needed that functionality in a project of mine, i wanted to read a file line by line, transform and write back to file, so a generator function seemed nice.
I feel it is more natural to process multiple points as tuples of (x,y,z) instead of separate arrays of coordinates. 
I added a stride-version "**_transform_sequence**" function in _proj.pyx, using the underlying proj4.transform to accelerate itransform.

PR #69, opened 2016-04-08
Add itransform function

georgeouzou:
An added function "itransform" similar to pyproj.transform, that enables transformation using generators. I needed that functionality in a project of mine, i wanted to read a file line by line, transform and write back to file, so a generator function seemed nice.
I feel it is more natural to process multiple points as tuples of (x,y,z) instead of separate arrays of coordinates.
I added a stride-version "_transform_sequence" function in _proj.pyx, using the underlying proj4.transform to accelerate itransform. This is the same as pull request #50 but it has no conflicts with the base branch.


micahcochran:
That version of Cython is a bit old "0.20.1" from 2014.   Cython 0.24 came out earlier this week.

itransform could be a good convenience function, but it probably isn't going to be as fast as transform.  However, there is a huge difference between projecting 1,000 points and 1,000,000 points.  If you are going after performance, you should expect to have to change your code based on benchmarking.

georgeouzou:
Thanks for the hints @micahcochran! I will update cython. As for the performance, indeed, i believe **transform** function to be faster, i thought of **itransform** as a more natural way of processing a sequence of points. If the function needs to be faster i can benchmark it.

micahcochran:
> If you are going after performance, you should expect to have to change your code based on benchmarking.

Sorry, I didn't mean you, @georgeouzou,  specifically.  I meant in general when writing code that sometimes changes have to be made to get more performance.   I've not benchmarked itransform versus transform.

georgeouzou:
If @jswhit is positive to the function as an idea, i can of course benchmark and tune it for better performance before including it in the library.

PR #66, opened 2016-03-29
Python3 Fixes.

citibeth:
Added a Python3 fix, in addition to what was fixed in PR #54.

This supercedes PR #54

PR #72, opened 2016-04-21
Change Proj.to_latlong in order to return an instance of pyproj.Proj

georgeouzou:
Returning a pyproj.Proj instance instead of the underlying base class _proj.Proj in pyproj.Proj.to_latlong function feels more intuitive.

This patch adresses a bug as well in pyproj.transform's function check. See Issue #53 
**Updated from #54, #56 and #66**


georgeouzou:
So many commits...Closing this and reopening a new "clean" one...

PR #63, opened 2016-03-23
Add __repr__ function for Proj & Geod.

micahcochran:
This is a small patch to make pyproj a little more developer friendly.

Currently repr() does this in the console:
```Python
In [14]: import pyproj

In [15]: p = pyproj.Proj(proj='latlong', preserve_units=True)

In [17]: p
Out[17]: <pyproj.Proj at 0x3028880>

In [18]: g = pyproj.Geod(ellps='clrk66')

In [20]: g
Out[20]: <pyproj.Geod at 0x230e7b0>
```

This pull request adds repr() functionality that look like this in the console:
```Python
In [1]: import pyproj

In [2]: p = pyproj.Proj(proj='latlong', preserve_units=True)

In [3]: p
Out[3]: pyproj.Proj('+proj=latlong ', preserve_units=True)

In [4]: g = pyproj.Geod(ellps='clrk66')

In [5]: g
Out[5]: pyproj.Geod(ellps='clrk66')
```

So, you can now tell what projection is in variable `p` or whatever other Proj/Geod object is.  I also added unit tests for \__repr__ function.

Issue #15, opened 2015-09-02
Pyproj test for two equal projections


fmaussion:
pyproj doesn't seem to check if projections are equal when asked to transform:

```python 
x = np.random.randn(1e7) * 60
y = np.random.randn(1e7) * 60
start_time = time.time()
xx, yy = pyproj.transform(pyproj.Proj(init='epsg:26915'), pyproj.Proj(init='epsg:26915'), x, y)
print(end_time-time.time())
```
Takes about 3 seconds on my laptop. I wonder if it would be possible for pyproj to catch these?

Would the following workaround work or is it dangerous?

```python 
def my_transform(p1, p2, x, y):
    if p1.srs == p2.srs:
        return x.copy(), y.copy() # or no copy if kwd nocopy=True
    return pyproj.transform(p1, p2, x, y)
```

Thanks,

Fabien

micahcochran:
I like the idea of a few lines of code detecting if the projections are the same.  I am unsure if the built-in `transform()` should do that checking.

Your work around will work for the above case because you defined both projections in exactly the same way.  It is not a bad way to test.  Look at my example using the   [same projection defined in two different ways.](http://epsg.io/26915)
```python
p1 = pyproj.Proj(init='epsg:26915')
p2 = pyproj.Proj('+proj=utm +zone=15 +datum=NAD83 +units=m +no_defs +ellps=GRS80 +towgs84=0,0,0')
if p1.srs == p2.srs:
  print("They are equal???")     # this shouldn't happen
else:
  print("Test fails.")
```

In order to make a projection equivalence test,  PyProj would either have to expose [`pj_get_def()`](https://github.com/OSGeo/proj.4/wiki/ProjAPI#advanced-functions) from Proj.4 or create a new function `Proj.__eq__()` that to check if the projections are equivalent and most likely use the `pr_get_def() .

I like the idea of a `Proj.__eq__()` so you can test `p1 == p2`.  Then, your code could test `==` and skip the transform.  There other useful application for that projection equivalency of function.

fmaussion:
Yes I see. I would find a `Proj.__eq__()` very useful in general, yes! How complicated would it be to implement?

micahcochran:
Pseudocode of how I would make a `Proj.__eq__()`.
 1. Run `pj_get_def()` on a projection
 2. convert Proj4 string to Fiona.crs dictionary representation, `from_string()`
 3. remove those key from dictionary that are for metadata, such as the 'init' key
[...]

PR #74, opened 2016-04-25
Add __repr__ and __eq__ functions

micahcochran:
This is a redo of PR #63 without conflicts and all the extra commits.  This adds representations for Geod and Proj classes that are a little more user friendly when using the python console.

Also, this adds a Geod.\__eq\__ function, which allows comparison between Geod objects.  I had issue #15 in mind when writing this, but I'm not really sure that this is really useful. 

In order for Geod.\__repr\__ function to work as written, issue #64 had to be fixed (which was done in PR #65).

### DIFF
--- a/_proj.pyx
+++ b/_proj.pyx
@@ -396,6 +396,11 @@ cdef class Proj:
         pj_dalloc(c_def_string)
         return py_def_string
 
+    def __repr__(self):
+        return "{modname}.{classname}({srs!r}, preserve_units=True)".format(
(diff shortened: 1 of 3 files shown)


micahcochran:
I resolved the conflicts.  I made this commit more pep8-like.  Proj.initstring was modified so that integer version of parameters `+a` and `+f` should appear in the string when those values do not have fractional portions.

micahcochran:
Self-merging.

Issue #77, opened 2016-05-05
Way to get back proj4 string?


nickeubank:
Curious: is there a way to get back a proj4 string from a Proj object? Preferably an informative one?

For example, if I passed:

    p = pyproj.Proj('+init=epsg:4326')

is there a way I could get back `"+proj=longlat +ellps=WGS84 +datum=WGS84 +no_defs"` through something like a method on `p`?

micahcochran:
p.srs is the proj4 initialization string for the Proj object.

You'd need to call proj4 pj_get_def function, which I don't think pyproj uses.

nickeubank:
@micahcochran Thanks! pj_get_def might be interesting to expose at some point, but p.srs will serve my purposes for now.

nickeubank:
@micahcochran Thinking of trying to add that, but don't have a lot of experience importing C functions (I'm familiar with cython as a tool, just usually use it to speed up python code, not to access C libraries). 

It looks like in `_proj.pyx` at line 56, there is an import of the function you're describing (pj_get_def). Would you mind providing some guidance to how I would go about exposing that to the Python api? Would it be sufficient to just write a `def` wrapper in the `_proj.pyx` file and import it in `pyproj/__init__`?

micahcochran:
#78 has been merged.  So I'm closing this.

PR #78, opened 2016-05-07
Initial commit of exposing pj_get_def

nickeubank:
Attempt to expose pj_get_def to user (related to #77). Input appreciated. 

Also, not clear to me how "tests" work in this library -- no unit testing with assert statements? Just print outputs for user to look for errors?

### DIFF
--- a/_proj.pyx
+++ b/_proj.pyx
@@ -62,6 +62,7 @@ cdef extern from "proj_api.h":
     projCtx pj_get_default_ctx()
     void pj_free(projPJ)
     void pj_set_searchpath ( int count, char **path )
+    void pj_dalloc(void *)
     cdef enum:
         PJ_VERSION
 
@@ -355,6 +356,15 @@ cdef class Proj:
             return True
         else:
             return False
+
+    def definition_string(self):
+        # Returns projection definition
+        cdef bytes py_def_string
+        cdef char* c_def_string
+        c_def_string = pj_get_def(self.projpj, 0)
+        py_def_string = c_def_string
+        pj_dalloc(c_def_string)
+        return py_def_string
 
 def _transform(Proj p1, Proj p2, inx, iny, inz, radians):
     # private function to call pj_transform
--- a/lib/pyproj/__init__.py
+++ b/lib/pyproj/__init__.py
@@ -422,6 +422,21 @@ class Proj(_proj.Proj):
         """returns True if projection in geocentric (x/y) coordinates"""
         return _proj.Proj.is_geocent(self)
 
+    def definition_string(self):
+        """Returns formal definition string for projection
+
+        >>> Proj('+init=epsg:4326').definition_string()
+        ' +units=m +init=epsg:4326 +proj=longlat +datum=WGS84 +no_defs +ellps=WGS84 +towgs84=0,0,0'
+        >>>
+        """
+
+        string_def = _proj.Proj.definition_string(self)
+
+        if not isinstance(string_def, str):
+            string_def = string_def.decode()
+
+        return string_def
+        
 def transform(p1, p2, x, y, z=None, radians=False):
     """
     x2, y2, z2 = transform(p1, p2, x1, y1, z1, radians=False)


micahcochran:
test/test.py is currently not ran by Travis, nor will it work in Python 3. Could you either add a docopt test or a test in unittest/test.py?

georgeouzou:
At _proj.pyx :
```python
return pj_get_def(self.projpj, 0)
```
There exists a nasty bug right there, as pj_get_def returns a **char*** c_string allocated by the proj4 library, and [cython automatically copies to a python string ](http://docs.cython.org/src/tutorial/strings.html#passing-byte-strings) that is returned by the function. But the c_string is never deallocated so you have a memory leak right there. 
To deallocate the c_string call proj4's void pj_dalloc(void *)

nickeubank:
@georgeouzou And that is why python programmers get in trouble with cython. :)

Would you mind advising me a little more on how to implement? My intuition is to do the following, but doesn't seem to be working:

- Add to the imports from `proj_api.h`: `void pj_dalloc(void *)`
- Change function def in _proj.pyx to:

```
    def definition_string(self):
          # Returns projection definition
          proj_def = pj_get_def(self.projpj, 0)
          pj_dalloc()
          return proj_def
```

@micahcochran Ah, I see -- was looking for tests in wrong place! Will add after fixing this bug.

nickeubank:
@micahcochran Where are docopt tests?

nickeubank:
Thanks all!

@micahcochran Not sure where docopt tests are, but found unittest and added there. 
@georgeouzou Would you mind taking a look at tweaks? I'm not a c programmer, so am muddling things together a little.

micahcochran:
Sorry, I had a brain fart.  Instead of "docopts tests", I meant to say "doctests".  Sorry for the confusion.  The tranform function has a good example of doctests.

nickeubank:
@micahcochran ahh, thanks. :)

micahcochran:
Even if it is showing the test passed. That test is still having problems (below)...   We need to look into it.

**Excerpt from Travis:**
```
Trying:
    'lon=%9.3f lat=%9.3f (degrees)' % (lon,lat)
Expecting:
    'lon=  -30.000 lat=   40.000 (degrees)'
ok
Trying:
    Proj('+init=epsg:4326').definition_string()
Expecting:
    ' +units=m +init=epsg:4326 +proj=longlat +datum=WGS84 +no_defs +ellps=WGS84 +towgs84=0,0,0'
**********************************************************************
File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/pyproj-1.9.5.1-py2.7-linux-x86_64.egg/pyproj/__init__.py", line 428, in pyproj.Proj.definition_string
Failed example:
    Proj('+init=epsg:4326').definition_string()
Exception raised:
      File "/opt/python/2.7.9/lib/python2.7/doctest.py", line 1315, in __run
        compileflags, 1) in test.globs
      File "<doctest pyproj.Proj.definition_string[0]>", line 1, in <module>
        Proj('+init=epsg:4326').definition_string()
      File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/pyproj-1.9.5.1-py2.7-linux-x86_64.egg/pyproj/__init__.py", line 433, in definition_string
        return _proj.Proj.definition_string(self).decode()
    UnicodeDecodeError: 'ascii' codec can't decode byte 0x83 in position 2: ordinal not in range(128)
Trying:
    p1 = Proj(init='epsg:4326')
Expecting nothing
ok
```

I'm not opposed to this function returning bytes.   A PROJ.4 string \*should\* only contain ASCII characters.  In Python (3.x) string is really a unicode string, which can includes ASCII and all the other characters.   Maybe I'm wrong about this.  This is the single biggest change between Python 2/3.

If `.definition_string()` were to return bytes, it could cause some oddness in `Proj(p.definition_string())` in working in Python 2.x, but not working in Python 3.x.

If using bytes for proj.4 strings was taken it is logical end, a few changes would have to be made for Python 3.x:
* Proj.\__init\__ would need to accept take byte strings.
* Proj object srs would have to be returned as byte strings.
* Proj object srs currently returns a string if given the parametrized form `Proj(init='epsg:4326').srs`  -> `'+units=m +init=epsg:4326 '`

@nickeubank   That is a much more extensive overhaul than what you trying to do with this patch, and should be discussed in a separate issue. To get a string, you might try using _strencode() in _proj.pyx.   That can be called within _proj.pyx (decode/encode should also work too).  I'm not quite sure this will fix the problem.

nickeubank:
@micahcochran yeah, good catch. It is saying the `+` is not ascii?? Weird...

I'll try `_strencode`.

nickeubank:
@micahcochran OK, this feels a little hacky but seems to work (at least on my builds)

georgeouzou:
@nickeubank, yes now it deallocates the memory of the c_string. 
You could do this in __init__.py:
```python
# Decode bytes --> Unicode String (for Python3)
if not isinstance(string_def, str):
    string_def = string_def.decode()
```
In case of python 2 the bytes are the same as str
In case of python 3 you decode to unicode.

georgeouzou:
@nickeubank, i'm sorry i just checked again (cannot be sure until you double check with c) ...
You must deallocate the c_string not the bytes:
```python
def definition_string(self):
    # Returns projection definition
    cdef bytes py_def_string
    cdef char* c_def_string
    c_def_string = pj_get_def(self.projpj, 0)
    py_def_string = c_def_string # cython copies to bytestring
    pj_dalloc(c_def_string)
    return py_def_string
```

nickeubank:
@georgeouzou thanks! :) fixed and rebased.

nickeubank:
Need anything else on this?

micahcochran:
@nickeubank  I like the idea for this.  It is exposing PROJ.4's `pj_get_def`, which is part of the public api (proj_api.h).  `pj_get_def` would be a rather awkward name for Python.

I have some additional unittests that I would like to add based on this function, which could be added with this PR or after it is committed.

 @jswhit What are your thoughts on this PR?

micahcochran:
Thank you, Nick!

PR #79, opened 2016-05-12
Doctest: add exit code for Travis of the number of failed doctests.

micahcochran:
This was found during a commit of PR #78 that a doctest failed, but passed in Travis.  This change will cause Travis to fail when a doctest fails. 

Also, remove duplicate lines in doctest for Geod class.


micahcochran:
Self-merging.

PR #82, opened 2016-05-21
Add Travis testing of PROJ.4 repo install and 4.9.2

micahcochran:
"nightly" is currently Python 3.6.0 alpha.

Compiling PROJ.4 from the repository allows a chance to catch some problems early.

Three unit test from the current PROJ.4 repository fail ~~due to a change in algorithm for the utm projection from tmerc to etmerc algorithm (see https://github.com/OSGeo/proj.4/issues/316).  Commit https://github.com/OSGeo/proj.4/commit/b0fb037ea22e2e9452feda502455ac6110c3e1e3 made on October 24, 2015.  This causes the results to shift by meters.~~ (**EDIT**: that is not the reason.)

PROJ.4 also recently had a [big PR merged that changed the macro structure](https://github.com/OSGeo/proj.4/pull/373) to make it more C-like.


micahcochran:
I added a test that compiles the PROJ.4 Release version 4.9.2 (second commit).  That version does not have the inverse of the hammer projection code.  Currently this is a widely installed version of PROJ.4.  It is the binary version for Ubuntu Linux 16.04 Long Term Service release (released April 2016).  Debian currently has this version of PROJ4 in testing.  I think it would be a good idea to try to keep some kind of support for PROJ4 v4.9.2 for a little while.

The inverse hammer doctest fails on 4.9.2 (unsurprisingly). The rest of the tests are failing similarly between it and the git version.

I'd like to figure out why these tests are failing.

micahcochran:
PR #86 was separated from this.  

Some of the coordinates 3-5 of the doctest coordinate tests differ.  I don't have a good explanation for it.  I'll try installing proj-datumgrid.

I should probably rest this commit before merging https://github.com/jswhit/pyproj/pull/82/commits/95397bb5145644fb8b839352fbfe72696f5efad5

micahcochran:
Self-merging.

PR #86, opened 2016-06-09
TEST: Move inverse hammer test from doctest to unittest.

micahcochran:
The unittests have different tests for PROJ.4 versions before and after the inverse hammer patch was integrated into the main PROJ.4 repository.  Fixes a doctest error for PROJ.4 version 4.9.2.  Retain similar testing method.

This was discovered during PR #82 and "cherry picked" from it.


micahcochran:
Self-merging.

Issue #83, opened 2016-06-06
Cannot import pyproj on Python 3.5.1 (OSX)


bsnizek:
Hey,

 I cannot import pyproj in my Python 3.5.1 virtualenv on the mac. 

```
$ python
Python 3.5.1 (v3.5.1:37a07cee5969, Dec  5 2015, 21:12:44) 
[GCC 4.2.1 (Apple Inc. build 5666) (dot 3)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import pyproj
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/besn/Dropbox/metascapes-ivs/Projects/2016-ebikes-gothenborg/elcykler-göteborg-2016/ebike-json-loader/lib/python3.5/site-packages/pyproj/__init__.py", line 251, in <module>
    set_datapath(pyproj_datadir)
  File "_proj.pyx", line 59, in _proj.set_datapath (_proj.c:887)
  File "_proj.pyx", line 370, in _proj._strencode (_proj.c:3939)
UnicodeEncodeError: 'ascii' codec can't encode character '\u0308' in position 78: ordinal not in range(128)
```

bsnizek:
Does not load when there is some utf-8 char in the path to the __init__.py

micahcochran:
I think I have fixed this.  Here is my [utf8-path branch](https://github.com/micahcochran/pyproj/tree/utf8-path).  (Note: Cython is required to installed in order to install from a pyproj repository.)  If you would, please test this patch.

In Linux, Python 2.x and 3.x gave the UnicodeEncodeError.   Here is test that I used to reproduce this problem:
```
$ python3 -c "import pyproj; pyproj.set_datapath(u'göteborg')"
```

micahcochran:
@bsnizek Did you get a chance to test this patch?

bsnizek:
@micahcochran : thanx for your answer, no not yet - I decided to rename the folder to be able to get on fast ;) Will try later...

frafra:
Same issue on Python 3.8 (trying to install pyproj with poetry).

micahcochran:
@frafra Could you post the error message?  (It might be a different line part of the code that is causing the error.)  Which version of pyproj are you trying to install?

snowman2:
Currently pyproj>=2.0.0 has binary wheels for osx. Feel free to reopen if they do not resolve your issues.

PR #88, opened 2016-06-20
Fix set_datapath to allow unicode paths to be decoded into bytes.

micahcochran:
Fix for issue #83.


micahcochran:
Self-merging.

PR #94, opened 2016-09-27
fix error when data in Chinese folder

net21:


micahcochran:
Issue #83 had some issues relating to utf-8, which I fixed in PR #88.

net21:
@micahcochran set_datapath  will failed  using string type contains unicode characters in python 3.4 when real running.

micahcochran:
1) Does the repository version have the bug?  2)  If YES, can you provide a little detail about how I can reproduce the bug on my end?

micahcochran:
Test case data needed to reproduce the bug.

micahcochran:
@net21 I'd like to close this bug.  Are there any reasons that I should not?

PR #114, opened 2017-05-06
Allow NaNs to be returned by geodesic routines; fixes #112.

### LINKED ISSUE #112: NaNs in Geod routines trigger "undefined inverse geodesic (may be an antipodal point)"
~~~
import pyproj
g=pyproj.Geod(ellps='WGS84')
g.fwd(1, 2, float('nan'), 10)
g.inv(1, float('nan'), 90, 3)
~~~
both trigger "undefined inverse geodesic (may be an antipodal point)" errors.  The Geod routines
are set up to "work" with nans (returning nans for results).  I believe allowing the nans to come
through the pyproj interface is the most useful behavior.  (Besides the reference to antipodal points
is misleading!)

cffk:


micahcochran:
Would you please add a unit test in test/unittest.py?

cffk:
Yes, I'll add a test (in a couple of days).

micahcochran:
Thanks @cffk

PR #115, opened 2017-05-06
Use geod_inverseline and geod_position to improve Geod.npts; fixes #113

### LINKED ISSUE #113: Geod.npts can be made about 2x faster
I see that Geod.npts works by calling geod_inverse and then making repeated calls to geod_direct.  Starting in 1.46, there's been a better way: call geod_inverseline and make repeated called to
geod_position on the resulting object.  A call to geod_position is about 2.3 times faster than
a call to geod_direct.  See the code fragment at

https://geographiclib.sourceforge.io/1.46/C/geodesic_8h.html#a639904c52fc74105d084ec679988ccb6

This may address issue #106.

cffk:


cffk:
The python 3.3 build has failed because it's not using the 4.9.3 version of geodesic.[ch].  But
I don't see how this can be...

micahcochran:
Proj 4.9.2 had GeographicLib (version 1.44).

`geod_inverseline` was added in GeographicLib version 1.46 (released 2016-02-15).

cffk:
Is the expectation that pyproj will work with both proj 4.9.2 and proj 4.9.3?  Is there
a preprocessor facility I can use in _proj.pyx?  If there is, I can check which version is
available.

micahcochran:
Proj 4.9.2 is still in use, that was the version released with Ubuntu 16.04 long term service release (support ending in 2021). It doesn't look like at this point in time there would be too much additional code by supporting both 4.9.2 and 4.9.3.  I would say that proj 4.9.2 and 4.9.3 should be supported.

There are multiple ways to check version numbers.

1. You can use ` PJ_VERSION >= 493` from PROJ.4's `#define` statement.
2. There is also string version of that `proj_version_str`.     In the unit tests, you'll see `proj_version_str > LooseVersion('4.9.2')`.  `LooseVersion` is from `distutils.version`.  
3.   Also, there is a `geodesic_version_str`.  `geodesic_version_str >= LooseVersion('1.46.0')`
4. `GEODESIC_VERSION_MAJOR` and `GEODESIC_VERSION_MINOR` are also exposed from C, so perhaps something like `(GEODESIC_VERSION_MAJOR, GEODESIC_VERSION_MINOR) >= (1, 46)`

Take your pick. I don't really have a preference on how the version is checked.

cffk:
Rather than checking version numbers, I reverted to the pre
proj-4.9.3 (and slightly more awkward) way of setting up the
geod_geodesicline.  The pull request should be ready to be
merged now.

micahcochran:
Charles, 
I will trust that this works.

Thank you for the PR,
Micah

PR #142, opened 2018-10-01
Sort input file list

bmwiedemann:
so that the .so file builds in a reproducible way
in spite of indeterministic filesystem readdir order
and http://bugs.python.org/issue30461

See https://reproducible-builds.org/ for why this is good.

This PR was done while working on reproducible builds for openSUSE.


micahcochran:
Thank you!

PR #143, opened 2018-10-01
Make nad2bin output reproducible

bmwiedemann:
initialize the struct CTABLE

Without this patch,
./nad2bin /tmp/x < datumgrid/stpaul.lla && md5sum /tmp/x
would differ after byte 41 because uninitialized memory was written
to the file and ASLR randomized the memory content

See https://reproducible-builds.org/ for why this matters.

### DIFF
--- a/nad2bin.c
+++ b/nad2bin.c
@@ -71,6 +71,7 @@ int main(int argc, char **argv) {
     const char *SUB_NAME = "";
     const char *CREATED  = "";
     const char *UPDATED  = "";
+    memset(&ct, 0, sizeof(ct));
 
 /* ==================================================================== */
 /*      Process arguments.                                              */


micahcochran:
`nad2bin.c` external vendor code from [PROJ.4 C library](https://github.com/OSGeo/proj.4/blob/master/src/nad2bin.c).

bmwiedemann:
Submitted the fix at https://github.com/OSGeo/proj.4/pull/1141

bmwiedemann:
Turned out that the fix is already in line 101, but our pyproj-1.9.5.1 (the latest release?) does not have it.

bmwiedemann:
Are there plans to make a new release? There are probably more fixes like this in there and I do not want to search to backport these.

jswhit:
we hope to release 2.0.0 (which required proj4 version 6.0.0) this month

Issue #136, opened 2018-07-10
Support for Python 3.7


AleksMat:
Hi,

now that Python 3.7 is officially the latest stable version, do you have any plans to support it?

At the moment `pyproj` is raising an error when I try to install it locally for Python 3.7.0 (using pip 10.0.1 and Linux Ubuntu). I get the same error as when I try to run Travis CI for my package which has `pyproj` as a dependency: [error message](https://travis-ci.org/sentinel-hub/sentinelhub-py/jobs/401845158)

AleksMat:
Thanks for adding Travis tests. I see that installing the package from GitHub repo works fine for me too. It is the latest version available on PyPI that is outdated and is raising an error during installation for Python 3.7.

Could someone also make a new release and upload it to PyPI?

micahcochran:
> Thanks for adding Travis tests. I see that installing the package from GitHub repo works fine for me too. It is the latest version available on PyPI that is outdated and is raising an error during installation for Python 3.7.

I suspect it is an issue with the C file generated by Cython.  The version requirements for Cython probably need to be updated in requirements-dev.txt... I guess Cython 0.28 should be the minimum to support Python 3.7.  ([Cython Changelog](http://docs.cython.org/en/latest/src/changes.html))

AleksMat:
When I use the latest Cython version `0.28.3` installing from PyPI still doesn't work, but from GitHub it does. So the reason must be some of the improvements that have been done since the latest release to PyPI (6th January 2016) until now.

micahcochran:
It looks like 0.28.4 was (released last week)[https://github.com/cython/cython/releases], so I guess that should now be the minimum Cython version.

valgur:
> When I use the latest Cython version 0.28.3 installing pyproj from PyPI still doesn't work, but from GitHub it does. So the reason must be some of the improvements that have been done since the latest release to PyPI (6th January 2016) until now.

The reason why the PyPI release does not work but installing from Git (with `pip install git+https://github.com/jswhit/pyproj.git`, for example) does is essentially the same that @micahcochran already mentioned: the `_proj.pyx` needs a recompilation to `_proj.c` with a newer version of Cython. The difference between the PyPI release and the Git repo is that the `_proj.c` file is not included in the latter, which triggers a recompilation of the `_proj.pyx` file and solves the problem, provided that a correct version of Cython is installed.

aleksanb:
Is this releasable now?
[...]

Issue #141, opened 2018-09-13
Pyproj Update needed for windows 10 64 and python


FrancoisMoolman51:
To Whom it May Concern:.

Please could somebody update and fix issues for pyproj as it now complete out dated !!

As i needed it for project and pip can't install it !!

![screenshot_4](https://user-images.githubusercontent.com/6871602/45488823-e87d3d80-b762-11e8-8c51-a43edafc4e30.png)

ocefpaf:
There are no binaries wheels on PyPI for `pyproj`. You'll have to use Gohlke's binaries,

https://www.lfd.uci.edu/~gohlke/pythonlibs/#pyproj

of `conda` to get a compiled version. The alternative is to compile it yourself but I do not recommend that on Windows.

snowman2:
Refer to #216

Issue #133, opened 2018-05-11
pip install pyproj


samsammurphy:
* Pip version: pip 9.0.1
* Python version: Python 3.6.4
* Operating system: Linux Mint 18.1 Serena

### Description:

I am trying to install pyproj using pip. It looks for libpython3.6m.so.1.0 but does not find it, which was not expected.

### What I've run:

```
$ mkvirtualenv pyproj --python=python3
$ pip install pyproj
$ python
>>> import pyproj
#from pyproj import _proj
#ImportError: libpython3.6m.so.1.0: cannot open shared object file: No such file or directory
```

 I was able to install from [source](https://pypi.python.org/pypi/pyproj/1.9.5.1) following the README,  reporting here in case helpful.

snowman2:
Can you try pyproj==2.0.1 and see if it works?

snowman2:
Feel free to reopen if this issue persists. Thanks for the report!

Issue #100, opened 2016-12-14
RuntimeError: b'no system list, errno: 2\n'


mohmad-null:
I've installed PyProj using the wheel from here - http://www.lfd.uci.edu/~gohlke/pythonlibs/ using pip, as I don't want to install a 3GB Visual Studio "runtime" to compile it myself.
I'm using Windows with Python 3.5 (x64).

However, when I try and run my transformation, I get this exception:
```
Traceback (most recent call last):
  File "E:/X_Dev/ogc__xml_to_es/src/xml_to_es.py", line 196, in <module>
    process_specified_project(args.index_name, args.proj_to_process)
  File "E:/X_Dev/ogc__xml_to_es/src/xml_to_es.py", line 183, in process_specified_project
    process_a_project(es, proj_to_process, PROJ_PATH, index_name)
  File "E:/X_Dev/ogc__xml_to_es/src/xml_to_es.py", line 140, in process_a_project
    post_processed = post_proc.post_process(parsed_content)
  File "E:\X_Dev\ogc__xml_to_es\src\post_process_xml.py", line 29, in post_process
    this_layer[key] = bbox_to_envelope(value, this_layer)
  File "E:\X_Dev\ogc__xml_to_es\src\post_process_xml.py", line 114, in bbox_to_envelope
    coords = change_coords(coords, start_proj, 'EPGS:4326')
  File "E:\X_Dev\ogc__xml_to_es\src\post_process_xml.py", line 181, in change_coords
    target = pyproj.Proj(init=target_proj)
  File "C:\Python35\lib\site-packages\pyproj\__init__.py", line 365, in __new__
    return _proj.Proj.__new__(self, projstring)
  File "_proj.pyx", line 117, in _proj.Proj.__cinit__ (_proj.c:1281)
RuntimeError: b'no system list, errno: 2\n'
```

Reading around suggests that this is because pyproj cannot find my /data directory. However it does exist and the `pyproj.pyproj_datadir` does have the correct path to it- 

> 'C:\\Python35\\lib\\site-packages\\pyproj\\data'

The data directory has 40 items totalling just over 5 MB.

I have tried explicitly doing:
`pyproj.set_datapath(pyproj.pyproj_datadir)`

But it still refuses to run. Am I doing something wrong, or is there something odd going on?

(reading around only found these:
https://sourceforge.net/p/py2exe/bugs/139/
http://stackoverflow.com/questions/20201532/issues-with-pyinstaller-and-pyproj
[...]

PR #154, opened 2018-12-28
1.9.6 release

jswhit:


rgommers:
looks good to me. fyi there was a recent memory leak in Cython that was fixed in 0.29.2 and looked easy to trigger, so you probably want to use that for the release. there was another fix we needed for numpy/scipy that required at least >= 0.28.5 (can't remember the details).

heitorPB:
should this close #136 #141 #133 #100 #91 #90 #89 #18 #13 #9 (and possibly others)?

Issue #152, opened 2018-12-09
Updates to support wkt with new proj version


snowman2:
Just wondering what the plans are for updating in connection with https://gdalbarn.com.

snowman2:
Current improvements:
```
>>> geog = Proj({'init': 'EPSG:4326'})
>>> geog.to_string()
'+proj=pipeline +step +proj=longlat +ellps=WGS84 +step +proj=unitconvert +xy_in=rad +xy_out=rad'
>>> geog.to_wkt()
'GEOGCRS["WGS 84",DATUM["World Geodetic System 1984",ELLIPSOID["WGS 84",6378137,298.257223563,LENGTHUNIT["metre",1]],ID["EPSG",6326]],PRIMEM["Greenwich",0,ANGLEUNIT["degree",0.0174532925199433],ID["EPSG",8901]],CS[ellipsoidal,2],AXIS["longitude",east,ORDER[1],ANGLEUNIT["radian",1,ID["EPSG",9101]]],AXIS["latitude",north,ORDER[2],ANGLEUNIT["radian",1,ID["EPSG",9101]]],USAGE[SCOPE["unknown"],AREA["World"],BBOX[-90,-180,90,180]]]'
```
```
>>> from pyproj import Proj
>>> projection_string = (
...     'PROJCS["USA_Contiguous_Albers_Equal_Area_Conic_USGS_version",'
...     'GEOGCS["GCS_North_American_1983",DATUM["D_North_American_1983",'
...     'SPHEROID["GRS_1980",6378137.0,298.257222101]],'
...     'PRIMEM["Greenwich",0.0],'
...     'UNIT["Degree",0.0174532925199433]],'
...     'PROJECTION["Albers"],'
...     'PARAMETER["false_easting",0.0],'
...     'PARAMETER["false_northing",0.0],'
...     'PARAMETER["central_meridian",-96.0],'
...     'PARAMETER["standard_parallel_1",29.5],'
...     'PARAMETER["standard_parallel_2",45.5],'
...     'PARAMETER["latitude_of_origin",23.0],'
...     'UNIT["Meter",1.0],'
...     'VERTCS["NAVD_1988",'
...     'VDATUM["North_American_Vertical_Datum_1988"],'
...     'PARAMETER["Vertical_Shift",0.0],'
...     'PARAMETER["Direction",1.0],UNIT["Centimeter",0.01]]]')
>>>
>>> proj = Proj(projection_string)
>>> proj.to_string()
[...]

PR #155, opened 2018-12-29
update to proj.4 6.0.0

snowman2:
Connected to issue #152 

This depends on: https://github.com/OSGeo/proj.4/pull/1208
Also, one of the tests fails, but should be fixed with: https://github.com/OSGeo/proj.4/pull/1205

There are a ton of changes - mostly to work with the new API. Feel free to make comments, ask questions, and be thorough in the review.


jswhit:
Does this really add dependencies on sqlite and cmake?

snowman2:
> Does this really add dependencies on sqlite and cmake?

They have moved from csv files to sqlite database:
"The use of a SQLite-based database for EPSG and other definitions will allow the projects to add more capability (area-aware validation), transition the custom peculiar data structures of the projects to something more universally consumable, and promote definition interoperability between many coordinate system handling software tools."

The cmake dependency makes building proj.4 from source very easy. However, there are other methods for compiling proj.4 without cmake. As aside note, if proj.4 is already installed, cmake is not required.

jswhit:
Maybe this would be a good time to remove the bundled proj4 source, and just require the library as a dependency.  Do your additions to the python API require proj4 6.0.0?

snowman2:
I think the changes I made make it required to have the proj.4 source installed to work. It has a step to compile the source for you in the setup.py internally, but that only works when installing locally from a clone of the repo. Otherwise it results in strange behavior with the path to files that should probably be handled better (could look into relative paths?). Otherwise it requires PROJ_DIR to be set beforehand.

The process definitely could be improved, simplified, or the behavior changed. cartopy is potentially a model that could be followed.

To answer your question, yes it will require version 6.0.0 of proj.4 to work due to the API changes.

snowman2:
@jswhit it looks like the integration with Travis CI has been updated and needs to be changed for the builds to work. To do this, I believe all you need to do is login to Travis CI and activate the `GitHub Apps Integration`.

jswhit:
Hmm. I see the travis tests are passing - so maybe whatever I did to travis actually worked.

README.md
```diff
@@ -3,20 +3,43 @@ pyproj
...
 * clone github repo or download source release at http://python.org/pypi/pyproj.
-  * If you clone the github repo, [Cython](http://cython.org/) is a dependency.
-* python setup.py build
-* python setup.py install (with sudo if necessary).
+  * [Cython](http://cython.org/) is required for the installation.
+
+Setup PROJ.4
+~~~~~~~~~~~~
+
+To install PROJ.4 insternally in pyproj, these are the dependencies:
```
jswhit: type here ('insternally')

README.md
```diff
@@ -3,20 +3,43 @@ pyproj
...
+    * C99
+    * C++11
+    * SQLite 3.7
+    * CMake 3.5
 
-To use proj4 lib (and data files) that are already installed on the system, 
+* git submodule init && git submodule update
+* python install_internal_proj4.py
+
+To use PROJ.4 lib (and data files) that are already installed on the system,
```
jswhit: Does this mean if you have an old proj4 installation, you can still use it (presumably sacrificing the new API you added that uses features from proj4 v6.0.0?)

jswhit:
Build fails with internal proj4 unless I set DYLD_LIBRARY_PATH=$PWD/pyproj/proj_dir/lib.  Even with that I'm getting a failure on import:

ImportError: dlopen(/Users/jsw/Library/Python/2.7/lib/python/site-packages/pyproj-2.0.0-py2.7-macosx-10.11-x86_64.egg/pyproj/_proj.so, 2): Symbol not found: _proj_angular_input
  Referenced from: /Users/jsw/Library/Python/2.7/lib/python/site-packages/pyproj-2.0.0-py2.7-macosx-10.11-x86_64.egg/pyproj/_proj.so

snowman2:
Hmmm, I haven't tested it out on a mac. Maybe adding something on the Travis builds for `osx` would be a good idea.

>This seems to be because the build linked an older version of proj4 on my system, not version 6.0.0.

I added functionality [here](https://github.com/snowman2/pyproj/blob/6463a6fa6066cade6e4a2cf3aba6bb744df83673/setup.py#L25) to let the user know if their version is not compatible. One way to check if that functionality is working would be to change `PROJ_MIN_VERSION` in the setup.py and see if it throws an error based on the version.

Also in the setup.py, the order for searching for PROJ.4 is:
1. PROJ_DIR env variable
2. Check if the internal proj dir exists (pyproj/proj_dir)
3. Check for the `proj` executable on the PATH.

So, you could try tinkering with those options.

README.md
snowman2: Unfortunately, no. The proj_api.h is deprecated and the behavior changed within the file. I moved all imports to use the new proj.h. But, wheels and  conda are good solutions for this problem.

snowman2: Well, I take the hard "no" back as I haven't tried it and there might be a way.

jswhit: OK, understood.

jswhit:
I was able to build and install proj.4 separately on macos x, then set PROJ_DIR to point to the installation, installed your gdalbarn pyproj branch and the tests passed.  This is with the specific revision of proj.4 used in the included source - if I use proj.4 master the tests fail.

I'm beginning to think we should just drop the included proj.4 source - there are too many issues building correctly on all platforms.

When is the anticipated release of proj.4 6.0.0?

snowman2:
>if I use proj.4 master the tests fail.

They updated the functionality of their code. I just made some changes so that all should work properly now.

jswhit:
travis is  building proj.4 correctly now, but a couple of tests fail.

snowman2:
I took a look at the tests and it looks like they pushed another change. One of the macos tests failing is unique - `to_epsg`. Is this what you see on your machine?

jswhit:
Yes, that's what I'm seeing on my machine also.

snowman2:
Good to know. I wonder if it has to do with how we are compiling the code or if it is an upstream issue.

snowman2:
I found the bug in the to_epsg function and all is good now.

snowman2:
Looks like I made a blunder when squashing and accidentally squashed one of your commits - sorry about that.

snowman2:
Sounds like a good plan to me. Two thumbs up for using Sphinx 👍

snowman2:
@jswhit if you get the repo connected to Appveyor, I can tinker with getting the tests/builds running there.

jswhit:
Appveyor is enabled now, it should run the next time you make a commit.

snowman2:
The Windows build works! The tests fail, but at least the build works.

snowman2:
@jswhit, I am noticing in the issues that the `preserve_units=False` flag default is throwing some people off. Any issue with me switching the default behavior to be `preserve_units=True` to minimize confusion since this will be a major version change?

snowman2:
Most recent wheel for Python 3.6 (worked alongside an older version of PROJ.4 installed for me) generated on Ubuntu 18.04.1: [pyproj-2.0.0-cp36-cp36m-linux_x86_64.whl.zip](https://github.com/jswhit/pyproj/files/2884900/pyproj-2.0.0-cp36-cp36m-linux_x86_64.whl.zip)

jswhit:
What needs to be done to get the tests passing?

snowman2:
> What needs to be done to get the tests passing?

I think it has to do with the version of pip: https://pip.pypa.io/en/stable/reference/pip/#pep-517-and-518-support. I am pinning to `pip>=10.0.1` and testing.

snowman2:
> worked alongside an older version of PROJ.4 installed for me

For clarification, I have an older version of PROJ.4 on the system and the new version of PROJ.4 (6.0.0) in the wheel does not conflict with it.

snowman2:
The Python 2.7 & 3.6 manylinux wheels build okay, but the 3.4, 3.5, and 3.7 versions are not including the `proj_data` directory for some reason.

snowman2:
All manylinux wheels build properly now.

jswhit:
This is fantastic!

jswhit:
Looks like we will be able to merge this and make a new release when proj4 6.0.0 is released (supposedly March 1?)

jswhit:
The README should probably mention that proj4 version 6.0.0 is required.

snowman2:
@jswhit I definitely hope that pyproj can be released along with the next version of PROJ.4.

What are your thoughts about:
- [ ] Using [coveralls](https://coveralls.io/)
- [ ] Using pytest for testing instead of nose2
- [ ] Switching to Sphinx for docs (you mentioned you may be interested above)

An example .travis.yml using pytest with coveralls is [here](https://github.com/snowman2/gazar/blob/7e0317466afa96ccdd5dd214b924b9c93d8a7788/.travis.yml#L56).

jswhit:
I don't have an opinion on coveralls and pytest - I'll defer to those who know more about testing frameworks.  Switching to sphinx for docs really needs to be done before the release.

snowman2:
Sounds like a good plan to me. The test framework currently in use works fine, so a decision on that can be postponed to after the release. Would it be helpful for me to to add Sphinx docs to this PR?

jswhit:
Great! I think this is ready to merge now - what do you think @snowman2 and @micahcochran?   After merging, I'll create a ticket for sphinx doc updates which should be done before a release.

snowman2:
@jswhit, sounds good to me. No issues here. Do you have access to the windows wheels uploaded as artifacts in Appveyor?

snowman2:
Found them:

https://ci.appveyor.com/api/buildjobs/235jw58mg0qsi9xr/artifacts/dist%2Fpyproj-2.0.0-cp35-cp35m-win_amd64.whl
https://ci.appveyor.com/api/buildjobs/964dt9t1n9d27709/artifacts/dist%2Fpyproj-2.0.0-cp36-cp36m-win_amd64.whl
https://ci.appveyor.com/api/buildjobs/so47mvw7a53a7lsm/artifacts/dist%2Fpyproj-2.0.0-cp37-cp37m-win_amd64.whl

I am curious how they work. Need to test on a windows machine.
Should this release drop support for Python 2.7 since it it dies this year? Or do you want to add the Python 2.7 build back in appveyor?

snowman2:
The windows wheels are faulty. I have some ideas to try if you don't mind holding off merging for a bit.

snowman2:
The wheels with the proj.lib & sqlite3.dll aren't working either. Not sure why. Well, feel free to merge anyway and maybe an issue can be made to look into it later.

PR #179, opened 2019-03-10
Datadir Problems Patch

snowman2:


snowman2:
Addresses: https://github.com/jswhit/pyproj/issues/176

jswhit:
Should we get a 2.0.1 release out soon with this fix?

snowman2:
@jswhit, yes I think getting this patch out sooner than later would be good. I just removed the change causing the build issues and targeted the exact fix. Should be good to merge when the build passes.

snowman2:
It passes and is ready to merge.

PR #182, opened 2019-03-11
Include .pxd files in sdist

cgohlke:


snowman2:
Nice catch, thanks!

jswhit:
Is this preventing builds from the source tarball from working? If so, we need a 2.0.2 release.

Merging now...

snowman2:
I would think so. I agree with the 2.0.2 release. 👍

Issue #183, opened 2019-03-11
+no_rot=True worked in 1.9.6, but not in 2.0.1


TAlonglong:
Currently we are passing a dict with parameters to `Proj`

We then passes like:
```
p_d = {'proj': 'omerc',
       'lat_2': 80.27942,
       'lat_0': 62.87671,
       'lat_1': 42.751232,
       'ellps': 'WGS84',
       'no_rot': True,
       'lon_1': 33.793186,
       'lon_2': -18.374414}

pyproj.Proj(p_d)
```

This fails with:
```
proj_create: invalid value for no_rot
Traceback (most recent call last):
  File "test-proj.py", line 14, in <module>
    pyproj.Proj(p_d)
  File "/home/trygveas/.local/lib/python3.5/site-packages/pyproj/__init__.py", line 322, in __init__
    self.crs = CRS.from_user_input(projparams if projparams is not None else kwargs)
  File "/home/trygveas/.local/lib/python3.5/site-packages/pyproj/crs.py", line 218, in from_user_input
    return cls(**value)
  File "/home/trygveas/.local/lib/python3.5/site-packages/pyproj/crs.py", line 140, in __init__
    super(CRS, self).__init__(projstring)
  File "pyproj/_crs.pyx", line 322, in pyproj._crs._CRS.__init__
pyproj.exceptions.CRSError: Invalid projection: +proj=omerc +no_rot=True +lon_1=33.793186 +lat_2=80.27942 +lat_0=62.87671 +ellps=WGS84 +lat_1=42.751232 +lon_2=-18.374414 +type=crs
```

I think the problem is `+no_rot=True` which should be `+no_rot`

How can this be achieved?

TAlonglong:
The conversion from dict 2 string happens here:

https://github.com/jswhit/pyproj/blob/master/pyproj/crs.py#L32

Was that different before?

jswhit:
I don't see anything different happening in v2.0.1 compared to v1.9.6.  I think the difference must be in how the C lib handles this in v6.0.0.

snowman2:
That sounds about right. We may need to add a filter for boolean values for the new version of proj.

jswhit:
Pull request #184 fixes this.  Still needs a test though.

jswhit:
Pull request #184 merged.  This will be in the 2.0.2 release, which we will try to get out the door this week.

PR #184, opened 2019-03-11
add filter for boolean values in dict2string so "no_rot=True" works  (issue #183)

jswhit:


snowman2:
Also, what about if they use None for the value?

jswhit:
handles False and None now

snowman2:
I was thinking `None` could be considered analagous to True as the `None` would be considered the same as not having a value for the argument. Thoughts on this logic?

```python
...
if value is None or (type(value) is bool and value):
    pjargs.append("+" + key + " ")
elif type(value) is bool and not value:
    pass
...
```

snowman2:
Just updated the code ^^

jswhit:
Or None might be interpreted as don't have a no_rot flag added to the string....

snowman2:
It seems cartopy interprets None as not having a value.
See: https://github.com/SciTools/cartopy/blob/45d0479abed27f73c5bb8bac35036772a3faa773/lib/cartopy/_crs.pyx#L177

jswhit:
OK, changed it to your suggestion

snowman2:
Sounds good, thanks for fixing this 👍

TAlonglong:
Thank you for this quick respond to my issue.

Issue #192, opened 2019-03-14
Upgrade from 1.9.6 to 2.0.2 causes `ProjError: proj_trans_generic error: latitude or longitude exceeded limits`


kdebrab:
The following causes `ProjError proj_trans_generic error: latitude or longitude exceeded limits` with pyproj 2.0.2:

```python
import pyproj
custom_proj = pyproj.Proj("+proj=geos +lon_0=0.000000 +lat_0=0 +h=35807.414063 +a=6378.169000 +b=6356.583984")
wgs84 = pyproj.Proj('+proj=longlat +ellps=WGS84 +datum=WGS84 +no_defs')
lat, lon = 51.04715, 3.23406
pyproj.transform(wgs84, custom_proj, lon, lat)
```
Traceback:
```
File "/usr/local/lib/python2.7/site-packages/pyproj/__init__.py", line 515, in transform
    _proj._transform(p1, p2, inx, iny, inz, radians)
File "pyproj/_proj.pyx", line 353, in pyproj._proj._transform
    ProjError: proj_trans_generic error: latitude or longitude exceeded limits
```

(maybe relevant: the error occured in a Python 2.7 docker container)

The same code yields no errors with pyproj 1.9.6.

snowman2:
pyproj==2.1.0 is released with the fix. Feel free to reopen if the issue persists. Thanks for the report!

PR #193, opened 2019-03-14
fixed logic for converting between radians/degrees for transformations

snowman2:
Attempt for #192 & #195 - needs work.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -80,7 +80,8 @@ install:
 
 script:
   - python -c "import pyproj; pyproj.Proj(init='epsg:4269')"
-  - nose2 -v --coverage pyproj --coverage-report xml
+  - nose2 -v
+  - nose2 --coverage pyproj --coverage-report xml # coverage report on separate line as does not show failures
 
 after_success:
 - coveralls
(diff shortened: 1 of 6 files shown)


jswhit:
I still don't understand what `proj_angular_output` does.  The docs says "Check if an operation returns angular output", but that's pretty vague.

snowman2:
I am wondering if it is not working properly or I am missing something. Example usage: https://github.com/OSGeo/proj.4/blob/5dc01e985e86b441ba2a671b712d3b211c8a2b33/src/apps/proj.cpp#L144

snowman2:
Looks like my logic is off there. I will tweak it this evening.

snowman2:
Also, looks like CI/CD are not enabled for the new repo location,

jswhit:
I think travis and appveyor should be activated now.

snowman2:
@jswhit, if the tests pass, it should be good to merge in.

jswhit:
OK, thanks @snowman2 !

snowman2:
Not a problem. Fingers crossed it is stable 🤞

jswhit:
Do you think we should release 2.1.0 soon?

snowman2:
Yes, I do. I think there are some much needed fixes in the next version.

jswhit:
OK, I'll make a release tag tomorrow.

snowman2:
Sounds good. Also, the docs need to be rebuilt.

Issue #202, opened 2019-03-16
1.9.6 and 2.1.0 behave differently in case of impossible transformation


drnextgis:
I'm not sure that this is a right place for this issue but anyway the following code works fine with pyproj 1.9.6:

```
>>> import pyproj
>>> from functools import partial
>>> from shapely.geometry import Polygon
>>> from shapely.ops import transform
>>> 
>>> project = partial(
...     pyproj.transform,
...     pyproj.Proj(init='epsg:4326'),
...     pyproj.Proj(init='epsg:3857'))
>>> 
>>> g1 = Polygon([(-180, -90), (-180, 90), (180, 90), (180, -90), (-180, -90)])
>>> g2 = transform(project, g1)
>>> print(g2)
POLYGON ((inf inf, inf inf, inf inf, inf inf, inf inf))
>>> pyproj.__version__
'1.9.6'
```

but fails with the recent 2.1.0:

```
>>> g2 = transform(project, g1)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/denis/satellogic/env/lib/python3.6/site-packages/shapely/ops.py", line 234, in transform
    zip(*func(*izip(*geom.exterior.coords))))
  File "/home/denis/satellogic/env/lib/python3.6/site-packages/pyproj/transformer.py", line 307, in transform
    return Transformer.from_proj(p1, p2).transform(x, y, z, radians)
  File "/home/denis/satellogic/env/lib/python3.6/site-packages/pyproj/transformer.py", line 151, in transform
    self._transformer._transform(inx, iny, inz, radians)
  File "pyproj/_transformer.pyx", line 149, in pyproj._transformer._Transformer._transform
pyproj.exceptions.ProjError: proj_trans_generic error: tolerance condition error
```

Maybe @sgillies has some thoughts how to deal with it.

snowman2:
That is a PROJ error being raised. It is likely due to changes in version 6.0.0. @kbevers @rouault. I think the solution is to use 89.99999 instead of 90 etc. for the points.

rouault:
For webmercator, when latitudes tend to 90 deg, northing tend to infinity, so using 89.9999 is not the solution. The recommended area of use of EPSG:3857 is for latitudes between -85 and 85 degrees

snowman2:
Thanks for the clarification!

jswhit:
[...]

PR #203, opened 2019-03-16
add errcheck kwarg, default False so behavior is as in 1.9.6 (issue #202)

jswhit:
needs some docstring updates and a test.

kwarg `errcheck` argument added to transform.  Default `False` means invalid projection transform returns `inf` as in 1.9.6, instead of raising an exception.

`pyproj.Proj` already behaves this way, except 1.e30 is returned.


snowman2:
Sounds like a good solution to me 👍

snowman2:
Thoughts about raising warning if errcheck is false? Same for Proj?

jswhit:
If tests pass, this should be ready to merge.

jswhit:
I'm +0 on raising a warning - I personally find them annoying.

snowman2:
Can you add numpy to requirements-dev.txt?

snowman2:
I agree with you on the warning vote.

snowman2:
Should a 2.1.1 release be made with this update?

jswhit:
Yes, I think so.  Will try to do that tomorrow.

snowman2:
Sounds good. Thanks for fixing this and taking care of the release!

Issue #128, opened 2017-11-23
performance large number of transforms same datum


artttt:
Ive noticed on my machine that  when running thousands of transforms if the 2 projections have the same datum then transform takes longer then doing calls on the proj instance/s.

If the check of having many points and datum match can done cheaply then there is a 10%+ (greater with more points it seems) speed up on the common case of not needing to change datum.

Im sure this is known and that is why calling a proj instance is in the api as is transform. However, Ive tended to use transform as a single call that can deal with moving between any 2 projections. I dont know how it works under the hood but maybe it can be adjusted to pick a faster route in these instances.

below is my timing testing
```
import pyproj
import numpy as np
import os, re

n = 100000

def datum_check():
    textfile = open(os.path.join(pyproj.pyproj_datadir,'epsg'), 'r')
    filetext = textfile.read()
    textfile.close()
    return re.findall(r"\n<(\d{4})>.*datum=WGS84", filetext)

epsg_with_wgs84_datum = datum_check()

for epsg in epsg_with_wgs84_datum:
    p1 = pyproj.Proj(init='epsg:3857')
    p2 = pyproj.Proj(init='EPSG:'+epsg)
    coords = (np.linspace(16000000, 17000000, num=n),np.linspace(-3000000, -4000000, num=n))
    same_projection = 'epsg:3857' in p2.srs
    epsg_srs = re.search(r'epsg:(\d{4})',p2.srs)
    same_datum = epsg_srs and epsg_srs.group(1) in epsg_with_wgs84_datum
    
    print epsg + ' transform time'
    %timeit f1 = pyproj.transform(p1, p2, *coords)
    f1 = pyproj.transform(p1, p2, *coords)
    if same_projection:
        print 'same projection'
        %timeit f2 = coords
        f2 = coords
    elif 'epsg:4326' in p2.srs:
        print 'same datum - geographic'
        #wgs84 plate carree projection ie geographic projection with same datum as epsg:3857
        %timeit f2 = p1(*coords,inverse=True)
        f2 = p1(*coords,inverse=True)
    elif same_datum:
        print 'same datum'
        %timeit f2 = p2(*p1(*coords,inverse=True))
        f2 = p2(*p1(*coords,inverse=True))
    else:
        print 'different datum full transform'
        %timeit f2 = pyproj.transform(p1, p2, *coords)
        f2 = pyproj.transform(p1, p2, *coords)
    print np.allclose(f1,f2)
```

micahcochran:
[...]

PR #206, opened 2019-03-18
skip tranformation operation if input and output projections are equivalent

snowman2:
issue #128

### DIFF
--- a/pyproj/_crs.pyx
+++ b/pyproj/_crs.pyx
@@ -584,6 +584,11 @@ cdef class _CRS:
         """
         return self.proj_type == PJ_TYPE_GEOCENTRIC_CRS
 
+    def is_exact_same(self, _CRS other):
+        """Compares projections to see if they are exactly the same."""
+        return proj_is_equivalent_to(
+            self.projobj, other.projobj, PJ_COMP_STRICT) == 1
+
     def __eq__(self, _CRS other):
         """Compares projections to see if they are equivalent."""
         return proj_is_equivalent_to(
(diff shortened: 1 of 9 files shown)


snowman2:
If tests pass, should be good to go if you like the implementation.

jswhit2:
Do we really need a `skip_equivalent` keyword, or can we just always skip the transform if the projections are equivalent?

snowman2:
I have had the same thought. My main concern is the possibility that there may be similar projections but going through with the transform would have a minor shift. So, adding the flag was my way of addressing this concern. 

If I knew that it was something I could depend on 100%, I wouldn't have any issues removing the flag.

snowman2:
@kbevers @rouault, how safe would it be to skip transformations based on `

snowman2:
I am hitting all kinds of wrong buttons with my phone 😛

snowman2:
Looks like there's a strict option. https://github.com/OSGeo/proj.4/blob/4b99c3b47e76afdeda2425e39ff53fed0f12b679/src/proj.h#L543

Maybe always skip on strict with equivalent as optional?

snowman2:
Thoughts on the updated version?

snowman2:
Different thoughts I have:
1. If <100% sure the equivalent projections will produce the same results, keep in current form
2. If ~100% sure equivalent projections will produce the same as exact, and feeling okay about it, make `skip_equivalent` default to true.
3. 100% the equivalent projections will produce same results, remove `skip_equivalent` option and exact comparison, and skip all equivalent transformations.

I went with (1) due to my lack of knowledge/experience with all the different combinations of projections. But, feel free to fill in any missing gaps.

PR #209, opened 2019-03-18
Remove cruft from unsupported software versions

micahcochran:
Remove cruft in unittest/test.py from supporting old Python and old PROJ versions.

### DIFF
--- a/unittest/test.py
+++ b/unittest/test.py
@@ -1,22 +1,11 @@
 # -*- coding: utf-8 -*-
 """Rewrite part of test.py in pyproj in the form of unittests."""
-from __future__ import with_statement
 
 import math
-from distutils.version import LooseVersion
-from sys import version_info as sys_version_info
+import unittest
 
-import pyproj
-from pyproj import Geod, Proj, pj_ellps, pj_list, proj_version_str, transform
+from pyproj import Geod, Proj, pj_ellps, pj_list, transform
 from pyproj.crs import CRSError
-
-if sys_version_info[:2] < (2, 7):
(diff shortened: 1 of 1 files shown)


snowman2:
Looks good. Thoughts on adding a test to check the proj version string? Maybe a regex match?

Mostly thinking that as it was kinda tested in the pieces you are removing.

micahcochran:
@snowman2 I'd try distutils.version either StrictVersion or LooseVersion to see how far you get with those.  Those work pretty well.  I'm not sure how fast they are, but that didn't matter so much because I was using them in unittest code.

Personally, regex would be my last resort.  Baring distutils, I might use some kind of proj_version_str.split('.')[0] comparison.  Or I might actually somehow cleaverly expose PROJ_VERSION_MAJOR and  PROJ_VERSION_MINOR from C, if it needed to be fast.

This PR has two goals, (1) get rid of some code needed for Python version below 2.6. If PROJ versions less than 5 are no longer supported, (2) there is no longer a reason to test for an RuntimeError exception created by the inverse of the Hammer projection, which is what was happening in the standard version of PROJ versions 4.9.2 and below.

snowman2:
I definitely agree with your goals of the PR. I think it is nice to do some cleanup. Any methods for checking would be fine. Even just asserting it is there works for me.

jswhit:
Agree that the place to check the Proj version is in setup.py.  However, build will fail with Proj < 6 anyway, so this is not critical.  Going to go ahead and merge.

micahcochran:
Thanks.

PR #211, opened 2019-03-20
Update setup.py PROJ version checking 

micahcochran:
In check_proj_version() add installed PROJ version to error message, return proj_version, use pkg_resources.parse_version() to provide a more featureful Version object.

After @snowman2 asked about the matter, so I just dug into it.  This is the [better Version object](https://stackoverflow.com/questions/11887762/how-do-i-compare-version-numbers-in-python) because distutils LooseVersion and StrictVersion are poorly documented and support the superceeded PEP 386.  pkg_resources.parse_version follows PEP 440, and pkg_resources is provided by setuptools.

The returned proj_version value could be useful if there needs to be some other steps that have to be taken for future version of PROJ, while supporting older PROJ versions with operators such as 
`version >= parse_version('6.1.0')`.

So on my system it provides this error message
```
micah@micah-laptop:~/prj3/pyproj$ pip3 install .
Processing /home/micah/prj3/pyproj
    Complete output from command python setup.py egg_info:
    ERROR: Minimum supported proj version is 6.0.0, installed version is 4.9.3.
    
    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-o3a25tgo-build/
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,11 +1,10 @@
-from __future__ import with_statement
-
 import os
 import subprocess
 import sys
 from collections import defaultdict
 from distutils.spawn import find_executable
 from glob import glob
+from pkg_resources import parse_version
 
 from setuptools import Extension, setup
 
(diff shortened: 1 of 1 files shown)


snowman2:
This is a nice addition! Side note: I wasn't referring to the setup.py in my previous comment, but this seems to be a beneficial miscommunication. I probably should have communicated that better. Originally thinking about test coverage for checking the pyproj.proj_version_str was set properly.

Issue #176, opened 2019-03-08
Transform can not find proj.db


stuart23:
I can no longer perform transformations. The minimum (platform independent) example code is:

    docker run -it python bash -c "pip install pyproj && python"
    >>> import pyproj
    >>> pyproj.transform(pyproj.Proj(init='epsg:4326'), pyproj.Proj(init='epsg:3857'), 0, 0)

This returns:
    proj_create_operation_factory_context: Cannot find proj.db
    pj_obj_create: Cannot find proj.db

running with `pip install pyproj==1.9.6` works fine though.

jswhit:
Looks like you need to set the PROJ_LIB environment variable to point to the PROJ.4 data directory (where proj.db lives), or use pyproj.datadir.set_data_dir

stuart23:
Should that be automatically done as a default? What's the benefit of making the user specify it manually?

jswhit:
the proj.4 source and data is no longer bundled with the python module, so it doesn't know where it is installed until you tell it.

stuart23:
Imho that goes against what python is. It either comes bundled or is listed
as a dependancy

stuart23:
ESPECIALLY when used in a containerized environment. Unless its pip
installable, no one will use it.

snowman2:
This will be fixed in the 2.0.1 release coming soon.

snowman2:
Did pyproj==2.0.1 fix this issue?

patrickziegler:
Not at all and I am probably missing how it was supposed to do so anyway.

The default guess on where to find `proj.db` is seen in the following excerpt from `datadir.py`. This results in `/usr/lib64/python3.7/site-packages/pyproj/proj_dir/share/proj` on my system (openSUSE Tumbleweed, pyproj 2.0.1).

```python
internal_datadir = os.path.join(
    os.path.dirname(os.path.abspath(__file__)), "proj_dir", "share", "proj"
    )
```

Why should the path to some files of PROJ.4 be relative to pyproj when you say yourself that PROJ.4 is not shipped with it?

If your thinking was to let maintainers create a symlink in `/usr/lib64/python3.7/site-packages/pyproj` pointing to the appropriate folder when he refrains to use the PROJ_LIB variable approach (which actually works btw.), why even bother with demanding those subfolders `"proj_dir"`, `"share"` and `"proj"`?

[...]

PR #213, opened 2019-03-21
Update data directory functionality

snowman2:
- look in PATH for data directory #176
- support multiple data directories #199
- use functions in setup.py to logically divide functionality (from old windows wheel attempts that were lost)

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -123,10 +123,10 @@ after_test:
   # If tests are successful, create binary packages for the project.
   - mkdir pyproj\proj_dir\share\proj
   - copy %PROJ_LIB%\* pyproj\proj_dir\share\proj
-  - mkdir pyproj\proj_dir\lib
-  - copy %PROJ_DIR%\lib\* pyproj\proj_dir\lib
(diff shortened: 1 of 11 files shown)


jswhit:
What's the use case for this?  A user with multiple data directories that wants to switch back and forth?

snowman2:
I was thinking along the lines of a user who does not have permissions to the main data directory who wanted to add datumgrids.

jswhit2:
> I was thinking along the lines of a user who does not have permissions to the main data directory who wanted to add datumgrids.

OK, that makes sense.

micahcochran:
> use functions in setup.py to logically divide functionality (from old windows wheel attempts that were lost)
@snowman2 That is a big improvement to setup.py.

snowman2:
@jswhit it should be ready for a final review.

jswhit:
Do you want this to be in the 2.1.2 release?  If so, go ahead and merge.  If not, I'll release 2.1.2 this weekend and we can merge after.

snowman2:
I merged it in, thanks for the reviews!

PR #218, opened 2019-03-24
added support for 4d coordinate transformations

### LINKED ISSUE #208: Plans to support 4D coordinates?
Now that pyproj uses PROJ 6 behind the curtains it would be nice to be able to do more than just plain projections and simpel datum shifts. So, are there any plans to support 4D coordinates? Without having studied the pyproj in details it seems to be fairly straight forward to me to extend `pyproj.transform()` with an optional `t` input.

This would allow users of pyproj to do transformations that uses the 14-parameter version of the [Helmert transform](https://proj4.org/operations/transformations/helmert.html). Personally I would like to use it so that I can support the [NKG transformations](https://github.com/NordicGeodesy/NordicTransformations) code based on pyproj.

snowman2:
Resolves #208


snowman2:
I am not sure if this will work outside of the `from_pipeline`, so it may be restricted to the `Transformer` class. But, I could be wrong.

snowman2:
@jswhit, I was thinking the location of the additional time keyword arguments would make this a 2.2.0 release as it could break for those not using kwargs in the functions. Thoughts?
Also, I made more changes. Mind reviewing again when you get a moment?

jswhit:
I don't think adding another kwarg qualifies as an API change.  Yes, it will break code that treats the kwargs as positional args, but I'm not too concerned about that.

snowman2:
Sounds good. I was back and forth on that, so this solidified that decision. :+1:

unittest/test_transformer.py
```diff
@@ -108,3 +109,120 @@ def test_equivalent_pipeline():
...
+    )
+
+
+def test_4d_transform_crs_obs1():
+    transformer = Transformer.from_proj(7789, 8401)
+    assert_almost_equal(
+        transformer.transform(
+            xx=3496737.2679, yy=743254.4507, zz=5264462.9620, tt=2019.0
+        ),
+        (3496737.2679, 743254.4507, 5264462.962, 2019.0),
```
kbevers: This doesn't seem right. From PROJ master:

```
$ echo 3496737.2679 743254.4507 5264462.9620 2019.0 |./cs2cs EPSG:7789 EPSG:8401
3496737.76	743253.99 5264462.70 2019.0
```

I don't know what tolerance is used in `assert_almost_equal` but a difference of ~70 cm in coordinates is significant and should, in my opinion, not be deemed equal.

Just for a bit of perspective: Often time these 4D transforms are used to account for the effect of continents moving over time. The Eurasian plate moves at a rate of 2-2.5 cm/year.

snowman2:
Assert almost equal goes to 7 decimals by default. So, definitely something problematic going on here.

Issue #207, opened 2019-03-18
pyproj 1.9.5.1 and 2.0.1 give different result for Lambert conical


fgro:
I have been using it for large topographic data transformations.

Now, I have switched my python IDE and updated pyproj (pip install pyproj) along the way.

For the transformation from 
Morocco, epsg 26191 (E, N, H) -> wgs84, I get three results:
	lat	long
```
http://epsg.io/transform#s_srs=26191&t_srs=4326&x=567623.9310000&y=256422.7870000	

epsq website    32.9021985°	-4.6753453°
spyder, pyproj-1.9.5.1	32.902199035649964	-4.6753455960055135
PyCharm,  pyproj-2.0.2	32.904806501443474	-4.676927631902659
PyCharm, pyproj-2.1.1    32.904806501443474    -4.676927631902659
reverting to 
PyCharm, pyproj 1.9.5.1	32.902199035649964	-4.6753455960055135
```

I run the same script and init the projection as:

    Midelt = pyproj.Proj(init='epsg:26191')
    WGS84 = pyproj.Proj(init='epsg:4326')

    E = 567623.931
    N = 256422.787
    h = 1341.467

    Long1, Lat1, H1 = pyproj.transform(Midelt, WGS84, E, N, h, radians=False)

Do you have an idea where the discrepancy comes from?
Which version is the more trustworthy?

snowman2:
Would you mind upgrading to pyproj>=2.1.0 and checking if this issue still exists in the most recent version?

jswhit2:
Your example is incomplete - how is `WGS84` defined?  epsg:4326?

fgro:
WGS84 = pyproj.Proj(init='epsg:4326')

fgro:
With pyproj 2.1.1:
Long = -4.676927631902659°
Lat = 32.904806501443474°

So, the difference persists.

snowman2:
@kbevers @rouault what are your thoughts on this?

kbevers:
> what are your thoughts on this?

I don't know, the data is incomplete. What is the z-value of the output?

fgro:
pyproj 1.9.5.1
lon -4.675346° E, lat 32.902199° N, H 1392.252 m

pyproj 2.1.1
lon -4.676928° E, lat 32.904807° N, H 1341.467 m

kbevers:
[...]

PR #219, opened 2019-03-24
convert PROJ error messages from bytes to strings

cjmayo:
As seen here: https://github.com/pyproj4/pyproj/issues/207#issuecomment-474667389.

Already used here:
https://github.com/pyproj4/pyproj/blob/b1eac98bef754edbb28fe8b080e9f8421ed61cee/pyproj/_transformer.pyx#L157


snowman2:
Thanks for the update! Approved from my end.

PR #220, opened 2019-03-24
make tests in /test Python 3 compatible

cjmayo:
Replace:
  cPickle with pickle
  commands.getoutput() with subprocess.check_output()

All also run on Python 2.7.


snowman2:
Nice! Approved on my end assuming tests pass.

PR #221, opened 2019-03-24
make testing fail on transformer docstring test failures

cjmayo:


snowman2:
Thanks for the fix. The failures are expected (see #202). Mind checking if the proj version == 6.0.0, the error count in transforms docs is 2, otherwise 0?

cjmayo:
May well be the same problem I was fixing with #222.

snowman2:
Maybe cherry-pick that commit into this PR?

snowman2:
Thanks @cjmayo!

PR #222, opened 2019-03-24
avoid transformer docstring test failure due to CRS

cjmayo:
Replace Beijing 1954 with New Beijing that superseded it.

---

This must be passing for others but my installation of PROJ 6.0.0 does not like this:

```
echo 116.366 39.867 | cs2cs epsg:4214 epsg:4326 
*	* inf
```

Seems easier to fix like this, hope that is OK.


snowman2:
Fine with me 👍. Thanks!

PR #212, opened 2019-03-20
 Updated to use the CRS definition for Proj instances in transforms

snowman2:
#207 

The main question I have is: Why is it failing for epsg:4214 to epsg:4326 using the CRS method but works for the PROJ method?

```
pyproj.exceptions.ProjError: proj_trans_generic error: b'no system list, errno: 22\n'

```

PR #223, opened 2019-03-24
replace use of nose2.tools in test.py

cjmayo:
Use unittest.TestCase.subTest() instead of nose2.tools.params().

test.py will now run and all tests execute without nose2, but
ForwardInverseTest cannot be used with Python 2.

---

A trade-off here, depends what is most important.

### DIFF
--- a/unittest/test.py
+++ b/unittest/test.py
@@ -2,6 +2,7 @@
 """Rewrite part of test.py in pyproj in the form of unittests."""
 
 import math
+import sys
 import unittest
 
 from pyproj import Geod, Proj, pj_ellps, pj_list, transform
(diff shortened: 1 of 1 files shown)


snowman2:
Another option would be to use pytest.mark.parametrize & switch to use pytest since it's already a test dependency.

snowman2:
Or wrap it in a try/except Import Error and default to the nose2 version.

snowman2:
Actually, looks good as is to me. Python 2 support is already dropped for Windows.

micahcochran:
There are 3 tests that are failing that aren't bubbling up to Travis CI.  Is that something to be concerned about?

snowman2:
Those are known failures related to proj. A workaround has been added in #222.

snowman2:
See #212 as well for reference.

snowman2:
Also see #221

PR #228, opened 2019-03-27
added check for error count in transformations

snowman2:
Sometimes errors may not be caught in the default way, this is a fallback method for catching them.


snowman2:
Refer to #208

snowman2:
Interesting, it is raising some new errors:
```
pyproj.exceptions.ProjError: 1 proj_trans_generic error(s): (Internal Proj Error: proj_crs_get_coordinate_system: Object is not a SingleCRS)
```

PR #230, opened 2019-03-27
use pytest for running the tests; merge test & unittest folders

snowman2:
Since we already have pytest as a dependency and it is the standard testing tool used by rasterio, fiona, geopandas, etc.. Feel free to share your thoughts on this.

This also addresses some tests not being run on windows.


snowman2:
Well, this snowballed :). If tests pass, should be good for review.

PR #231, opened 2019-03-27
fixed projection equivalence testing & projecting with time transformations

snowman2:
issue #208


snowman2:
@jswhit, this one should be good to go pending your approval.

snowman2:
Thanks!

PR #233, opened 2019-03-30
update doc links; add deprecation for from_crs; update changelog

snowman2:


snowman2:
Thoughts about deprecating the `from_crs` method? With the changes it seems redundant to me.

jswhit:
> Thoughts about deprecating the `from_crs` method? With the changes it seems redundant to me.

Don't think it should be removed right away. Is it redundant because `from_proj` does the same thing?  If so, maybe it can just forward to `from_proj`?

snowman2:
Currently it is forwarding to `from_proj` and this PR adds a deprecation warning that it will be removed in the 2.2.0 release. Does that sound reasonable?

snowman2:
Also, what are you thinking about the timeline for the 2.1.3 release?

jswhit:
> Also, what are you thinking about the timeline for the 2.1.3 release?

Whenever you're ready

snowman2:
I think it is good to go for the 2.1.3 release.

jorisvandenbossche:
Can I ask why `from_crs` was deprecated? 
I understand that it is duplicating the functionality of `from_proj`. 
But, from the point of view of a new user who would deal with CRS objects, having a `from_crs` feels more natural than having to use the `from_proj`.

snowman2:
> Can I ask why from_crs was deprecated?

I think that having two methods doing essentially the same thing is not necessary and can confuse users to think the CRS object cannot be used with the `from_proj` method. Though, I see your point about the `from_crs` method. Maybe adding docs/examples to clarify this would help.

jorisvandenbossche:
Yep, I agree that two methods to do the same thing are not really needed.

But, given that in the "new" proj 6 / pyproj 2 world, the first entry point for users is probably the CRS object (although this might depend on your use case of course), I find it a bit unfortunate it is `from_crs` that is being deprecated, instead of `from_proj` (now we still have the chance to change that, IMO). 
It is also more or less wrapping the proj `proj_create_crs_to_crs`, which also uses `crs`.

Anyway, I said my opinion, so will leave it at that :-)

snowman2:
I have been thinking about it and agree with keeping from_crs. This can go into a separate issue to make the change later. Thanks for bringing this up :+1:

PR #245, opened 2019-04-06
replace deprecated time.clock() in tests

cjmayo:
Fix warnings and prepare for Python 3.8.

### DIFF
--- a/test/test_awips221.py
+++ b/test/test_awips221.py
@@ -1,5 +1,9 @@
 import array
-import time
+
+try:
+    from time import perf_counter
+except ImportError:
+    from time import clock as perf_counter
 
 import numpy
 from numpy.testing import assert_allclose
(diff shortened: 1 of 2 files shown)


snowman2:
I think a python 2/3 compatible solution would be good to add here.

micahcochran:
The release of Python "3.8.0 final: Monday, 2019-10-21."  

This could merged once this project has dropped Python 2.7 support.  For the moment, I agree with @snowman2 for a solution compatible with both Python 2/3.

cjmayo:
My mistake - I forgot I had been starting the process of removing Python 2 from my system.

Unfortunately I can't see a good solution that works for Python 2 and Python 3. I think it means going for either a potentially lower resolution `time.time()` or removing the timings completely. At least there is no immediate need as pointed out.

snowman2:
You could instead use `datetime.datetime.utcnow()`

snowman2:
Maybe a try/except ImportError using `as` with a preference towards the new version would provide the solution.

micahcochran:
Another way to do this might be to convert this code to some kind of benchmark package, which would involve more effort.

Recently, there have been some issues regarding performance concerns.

I do find it interesting the `osx` test should be running Python 3.6, but instead is running Python 2.7.

snowman2:
> I do find it interesting the osx test should be running Python 3.6, but instead is running Python 2.7.

I noticed it as well, but didn't fix it yet as it is not a huge problem until python2 support is dropped.

cjmayo:
> Maybe a try/except ImportError using `as` with a preference towards the new version would provide the solution.

OK. That works. I went with `as perf_counter` thinking that although `clock` may not be equivalent (depending on platform)  it would cause the least work when eventually dropping Python 2.

snowman2:
Approved from my end. :+1:

PR #248, opened 2019-04-06
don't define unused Sphinx html_static_path

cjmayo:
No custom static files are used.
Avoid warning about html_static_path entry not existing.


snowman2:
Approved from my end. Thanks!

Issue #249, opened 2019-04-06
pyproj 2.1.3 transform() fails with errcheck=True


cjmayo:
https://github.com/pyproj4/pyproj/pull/228 introduced a check for the return value of proj_trans_generic which is saved as `err_count` although it is described as "Number of transformations completed" in PROJ:
https://github.com/OSGeo/proj.4/blob/e87b3744350baca2b7fe35070e621a6b45bcbf67/src/4D_api.cpp#L353

(proj_trans_array returns 0 if all coordinates are transformed without error, otherwise returns an error number:
https://github.com/OSGeo/proj.4/blob/e87b3744350baca2b7fe35070e621a6b45bcbf67/src/4D_api.cpp#L282)

This code:
```py
from pyproj import Proj, transform

p1 = Proj(init="epsg:4326")
p2 = Proj(init="epsg:27700")

print(transform(p1, p2, 1.716073972, 52.658007833, errcheck=True))
```

now fails with:
```
  File "/usr/lib/python3.7/site-packages/pyproj/transformer.py", line 367, in transform
    xx=x, yy=y, zz=z, tt=tt, radians=radians, errcheck=errcheck
  File "/usr/lib/python3.7/site-packages/pyproj/transformer.py", line 180, in transform
    inx, iny, inz=inz, intime=intime, radians=radians, errcheck=errcheck
  File "pyproj/_transformer.pyx", line 144, in pyproj._transformer._Transformer._transform
pyproj.exceptions.ProjError: 1 proj_trans_generic error(s)
```

As per this comment https://github.com/pyproj4/pyproj/issues/208#issuecomment-477136179, a return value of 0 may be an error (or just there was nothing to transform).
But you would expect a positive integer result on a successful transformation as above?

snowman2:
@kbevers, can you clarify this?

snowman2:
Nevermind, this is correct. I misinterpreted this and is a bug.

Issue #243, opened 2019-04-05
from_crs() in test_transformer.py


micahcochran:
Since `from_crs()` is deprecated since #233 .

Here's what I did to find the problem from the python console:
```
In [10]: pyproj.test()
/home/micah/.local/lib/python3.6/site-packages/pyproj/transformer.py:92: UserWarning: `from_crs` is deprecated and will be removed in 2.2.0. Please use `from_proj` instead.
  "`from_crs` is deprecated and will be removed in 2.2.0. "
```

There are 3 instances of `from_crs()` in test_transformer.py.  There is also [one here](https://github.com/pyproj4/pyproj/blob/master/pyproj/transformer.py#L144) that is being ran by the doctests. I thought that it might be a good idea to fix those in order to get rid of the warning messages.

snowman2:
I am fine getting rid of them in the doctests. But, until it is removed, the other tests need to remain. You can wrap it with `pytest.warns` to both test the warning exists and hide it.

PR #250, opened 2019-04-07
fix error count for transformations; fix from_crs deprecation warning

snowman2:
Addresses issues #243 & #249


snowman2:
Thanks for the trust here. Hopefully it is stable 🤞

Issue #240, opened 2019-04-04
Updating pj_ellps


micahcochran:
`pj_ellps` is derived from PROJ.

It seems to be missing a few ellipses.  ([History of pj_ellps.c as of PROJ 5.2.0](https://github.com/OSGeo/proj.4/blame/98ba73123dafe1fb56e4a9c6aef4555023d020e3/src/pj_ellps.c)), namely danish, GSK2011, and PZ90.

As of 6.0.0 this is in [ellps.cpp](https://github.com/OSGeo/proj.4/blob/master/src/ellps.cpp)

Is it better to just update the static dictionary?  Or should the approach be to call PROJ's `proj_list_ellps()`, parsed the output, and populate `pj_ellps`?

snowman2:
I like the idea of populating the list with `proj_list_ellps()`.

PR #251, opened 2019-04-08
add interface to get the projection/ellps/prime_meridian/units lists

snowman2:
Addresses #240


test/test_list.py
```diff
@@ -0,0 +1,41 @@
...
+    assert isinstance(ang_map["deg"], Unit)
+    assert ang_map["deg"].id == "deg"
+    assert ang_map["deg"].name == "Degree"
+
+
+def test_get_ellps_map():
+    ellps_map = get_ellps_map()
+    assert ellps_map["WGS84"]["description"] == "WGS 84"
+    assert_almost_equal(ellps_map["WGS84"]["a"], 6378137.0, decimal=1)
+    assert_almost_equal(ellps_map["WGS84"]["rf"], 298.257223563, decimal=1)
```
micahcochran: The "rf" factor should be tested to a few more decimal places.  Several of them have "rf" factors around 298.25.

snowman2: This is mostly just checking that it is in the correct format. It does not attempt to check values as that should be handled by PROJ.

snowman2:
I updated `Geod.__init__()` as well.

micahcochran:
Thanks @snowman2

PR #252, opened 2019-04-08
Travis-CI: peg osx to a specific PY version

micahcochran:
OSX was reverting to using Python 2.7.  Apparently, it needs all three numbers specified, so "3.6.8".

sudo: true is no longer needed on xenial.


snowman2:
Nice 👍

Issue #258, opened 2019-04-10
Provide option to "pretty print" WKT (MULTILINE=YES) ?


jorisvandenbossche:
When inspecting the WKT of a `CRS` (with `CRS.to_wkt())`, it can sometimes be useful to be able to "pretty print" the WKT instead of having it as one long string (eg in tutorials, explaining the different representations).

[`proj_as_wkt`](https://proj4.org/development/reference/functions.html#_CPPv311proj_as_wktP10PJ_CONTEXTPK2PJ11PJ_WKT_TYPEPPCKc) has a MULTILINE option to get this, but currently this is hardcoded to `NO`:

https://github.com/pyproj4/pyproj/blob/236538976e3e7193339737ed013c218a5a6d3982/pyproj/_crs.pyx#L276

As a test, I set this to `YES`, and then I get something like:

```
In [2]: crs = pyproj.CRS.from_epsg(4326) 

In [3]: crs.to_wkt()
Out[3]: 'GEOGCRS["WGS 84",\n    DATUM["World Geodetic System 1984",\n        ELLIPSOID["WGS 84",6378137,298.257223563,\n            LENGTHUNIT["metre",1]]],\n    PRIMEM["Greenwich",0,\n        ANGLEUNIT["degree",0.0174532925199433]],\n    CS[ellipsoidal,2],\n        AXIS["geodetic latitude (Lat)",north,\n            ORDER[1],\n            ANGLEUNIT["degree",0.0174532925199433]],\n        AXIS["geodetic longitude (Lon)",east,\n            ORDER[2],\n            ANGLEUNIT["degree",0.0174532925199433]],\n    USAGE[\n        SCOPE["unknown"],\n        AREA["World"],\n        BBOX[-90,-180,90,180]],\n    ID["EPSG",4326]]'

In [4]: print(crs.to_wkt())
GEOGCRS["WGS 84",
    DATUM["World Geodetic System 1984",
        ELLIPSOID["WGS 84",6378137,298.257223563,
            LENGTHUNIT["metre",1]]],
    PRIMEM["Greenwich",0,
        ANGLEUNIT["degree",0.0174532925199433]],
    CS[ellipsoidal,2],
        AXIS["geodetic latitude (Lat)",north,
            ORDER[1],
            ANGLEUNIT["degree",0.0174532925199433]],
        AXIS["geodetic longitude (Lon)",east,
            ORDER[2],
            ANGLEUNIT["degree",0.0174532925199433]],
    USAGE[
[...]

PR #260, opened 2019-04-10
added option to pretty print CRS.to_wkt

snowman2:
Addresses #258


jorisvandenbossche:
So the main question here I think is whether we want to use `pretty=True/False`, or keep closer to the proj api and have `multiline=True/False` as raised in https://github.com/pyproj4/pyproj/issues/258. 
(I don't have a strong preference)

snowman2:
No strong preference from me either. Either option is good.

Issue #253, opened 2019-04-08
CRS.srs attribute is not documented


jorisvandenbossche:
The `srs` attribute is not documented on the API page: https://pyproj4.github.io/pyproj/html/api/crs.html

From looking at the code, it seems this attribute is keeping the original "source" string how the CRS was created (which is then used in eg the repr) ?

snowman2:
You are correct that it is the original "source" string for how the CRS was created. It might be good to add it to the documentation.

jorisvandenbossche:
To take a step back: does this need to be public? (internally it is also only used for the repr and pickling, as far as I can see) 
And in that case, is `srs` a good name?

snowman2:
`srs` is also used in the Transformer class for transformations. It does not have to be public, but it is useful for users to compare what they input to what was used to create their projection. The name was chosen to be consistent with the existing `Proj.srs`.

jorisvandenbossche:
>  The name was chosen to be consistent with the existing Proj.srs.

Ah, that's a good reason (if starting fresh, I would rather go with something more descriptive such as `user_input`, as I find `srs` a bit confusing)

jorisvandenbossche:
But with a good documentation entry, it maybe shouldn't be that confusing. Because as I see, the transformer is actually using this as the actual CRS definition to create the proj pipeline?

snowman2:
I agree with you on the name. But, since it is there, for backwards compatibility, I think a documentation entry would be the way to go.

> the transformer is actually using this as the actual CRS definition to create the proj pipeline?

Yes, it has been the most stable method for creating the proj pipeline.

PR #261, opened 2019-04-10
document attributes on CRS,Geod,Proj classes

snowman2:
Addresses #253


pyproj/crs.py
```diff
@@ -52,6 +52,13 @@ class CRS(_CRS):
     * https://github.com/opendatacube/datacube-core/blob/83bae20d2a2469a6417097168fd4ede37fd2abe5/datacube/utils/geometry/_base.py
     * https://github.com/mapbox/rasterio/blob/c13f0943b95c0eaa36ff3f620bd91107aa67b381/rasterio/_crs.pyx
 
+    Attributes
+    ----------
+    srs: str
+        The string form of the user input used to create the CRS.
+    name: str
+        The name of the projection from `proj_get_name`.
```
jorisvandenbossche: projection -> CRS ? 
(and I also not sure if mentioning `proj_get_name` is that useful for the typical user)

snowman2: Good points. I will update that.

snowman2: Thoughts on the current iteration?

snowman2:
Thanks for the better descriptions for the Geod attributes @jswhit! I updated the docs with your suggestion.

Issue #262, opened 2019-04-10
API: CRS.datum being a CRS itself ?


jorisvandenbossche:
(this might be more a question on why this was implemented like this, but since I don't know of a better place to ask, opening an issue here. If there is a better venue to ask such questions, let me know!)

So if you define a CRS, eg Web Mercator here, and get the `datum` property, you get back a CRS object:

```
In [7]: crs = pyproj.CRS.from_epsg(3857)

In [8]: crs  
Out[8]: 
<CRS: epsg:3857>
Name: WGS 84 / Pseudo-Mercator
Ellipsoid:
- semi_major_metre: 6378137.00
- semi_minor_metre: 6356752.31
- inverse_flattening: 298.26
Area of Use:
- name: World - 85°S to 85°N
- bounds: (-180.0, -85.06, 180.0, 85.06)
Prime Meridian:
- longitude: 0.0000
- unit_name: degree
- unit_conversion_factor: 0.01745329
Axis Info:
- Easting[X] (east) EPSG:9001 (metre)
- Northing[Y] (north) EPSG:9001 (metre)

In [9]: crs.datum 
Out[9]: 
<CRS: DATUM["World Geodetic System 1984", ELLIPSOID["WGS ...>
Name: World Geodetic System 1984
Ellipsoid:
- semi_major_metre: 6378137.00
- semi_minor_metre: 6356752.31
- inverse_flattening: 298.26
Area of Use:
- UNDEFINED
Prime Meridian:
- longitude: 0.0000
- unit_name: degree
- unit_conversion_factor: 0.01745329
Axis Info:
- UNDEFINED
```

but, since this is a Datum, it is not a complete CRS object. For example, it has no AxisInfo (by definition I think, as only a CoordinateSystem has that), it also has no proj4 string representation (`crs.datum.to_proj4()` returns None), etc

I know that from the proj4 side, both CRS and Datum are `PJ` objects. So it can make sense to have them in a single class in the python wrapper. But, we already don't allow all types of `PJ` objects in the `pyproj.CRS` class (eg an Ellipsoid is also a PJ object and can have a EPSG number, but for Ellipsoid we have a separate class). 
And moreover, I think from a user perspective, it might be useful and clearer that a `CRS` class always is a CRS (and not Datum, Ellipsoid, ..)

snowman2:
[...]

PR #263, opened 2019-04-11
CRS Refactor (Add Datum, CoordinateOperation, CoordinateSystem classes; update CRS repr)

snowman2:
@jorisvandenbossche

Addresses:
#262
#264
#267 
#265
#274


pyproj/_crs.pyx
```diff
@@ -257,8 +257,117 @@ cdef class PrimeMeridian:
...
+        """destroy projection definition"""
+        if self.projobj != NULL:
+            proj_destroy(self.projobj)
+        # let the parent CRS class manage this
+        # if self.projctx != NULL:
+        #     proj_context_destroy(self.projctx)
+
+
+    @staticmethod
+    cdef create(PJ_CONTEXT* projcontext, PJ* projobj, int horizontal):
```
jorisvandenbossche: Is there a reason that you use here a static method `create` (and as you do in the other Ellipsoid et al classes), while in the `_CRS` class you use a more typical `__init__` ?

pyproj/_crs.pyx
```diff
@@ -387,15 +517,24 @@ cdef class _CRS:
...
-            return self.__class__(_to_wkt(self.projctx, projobj))
-        finally:
-            proj_destroy(projobj) # deallocate temp proj
+        if self._datum is not None:
+            return self._datum
+        self._datum = Datum.create(self.projctx, self.projobj, horizontal=False)
+        return self._datum
+
+    @property
+    def horizontal_datum(self):
```
jorisvandenbossche: I am not very familiar with "horizontal datum" (and also don't find much about that in the ISO specs or proj docs), but when is datum and horizontal datum different?

Eg for `pyproj.CRS.from_epsg(3857)` (or some others I tried) they give me the same.

(just wondering if it is needed, at least to start with, to have both attributes)

snowman2: Honestly, I am not familiar either. From the code, it seemed datum only works with a single CRS whereas horizontal datum works for a compound CRS. Maybe the proj tests have an example.

pyproj/_crs.pyx
snowman2: The reason is because the `__init__` does not support the PJ_CONTEXT* typing. But, the staticmethod does.

snowman2:
> Maybe the properties on CRS can point to the one on the Datum to avoid creating them twice?

There are cases where the datum is None (see the tests), but the ellipsoid and prime meridian still exist on the base CRS. So, I think leaving it as is may be better.

snowman2:
If you weren't referring to inheritance, there are cases where the datum is None (see the tests).

jorisvandenbossche:
Yes, I wasn't really referring to inheritance (although that would be a good idea for some common functionality), but to the actual properties. 

> If you weren't referring to inheritance, there are cases where the datum is None (see the tests).

That's indeed a problem for my suggestion. So I think this is the case for BoundCRS or CompoundCRS. The general question here is also to what extent we want to give access to specific properties of those (like getting access to the underlying CRSs)

pyproj/_crs.pyx
snowman2: Removed for now

snowman2:
I updated it again for your review - will probably add more later to address other issues.

pyproj/_crs.pxd
```diff
@@ -22,7 +22,14 @@ cdef class AreaOfUse:
     @staticmethod
     cdef create(PJ_CONTEXT* projcontext, PJ* projobj)
 
-cdef class Ellipsoid:
+
+cdef class BaseCRS:
```
jorisvandenbossche: Small nitpick: I would not call this `BaseCRS`, as it is not the base of only the CRS classes

pyproj/_crs.pyx
```diff
@@ -322,6 +414,23 @@ cdef class _CRS:
...
+                PJ_TYPE_GEOGRAPHIC_CRS,
+                PJ_TYPE_GEOGRAPHIC_2D_CRS,
+                PJ_TYPE_GEOGRAPHIC_3D_CRS,
+                PJ_TYPE_VERTICAL_CRS,
+                PJ_TYPE_PROJECTED_CRS,
+                PJ_TYPE_COMPOUND_CRS,
+                PJ_TYPE_TEMPORAL_CRS,
+                PJ_TYPE_ENGINEERING_CRS,
+                PJ_TYPE_BOUND_CRS,
+                PJ_TYPE_OTHER_CRS,
```
jorisvandenbossche: As mentioned in the issue, I suppose `proj_is_crs` can be used for this

pyproj/_crs.pyx
```diff
@@ -391,15 +500,56 @@ cdef class _CRS:
...
+        Returns
+        -------
+        CRS: The source CRS.
+        """
+        if self._source_crs is not None:
+            return self._source_crs
+        cdef PJ * projobj
+        projobj = proj_get_source_crs(self.projctx, self.projobj)
         if projobj == NULL:
             return None
```
jorisvandenbossche: Also assign the None case to `self._source_crs` so we don't have to retry every time you access this property? (since CRS is immutable, this should be fine?)

pyproj/_crs.pyx
```diff
@@ -391,15 +500,56 @@ cdef class _CRS:
...
         try:
-            return self.__class__(_to_wkt(self.projctx, projobj))
+            self._source_crs = self.__class__(_to_wkt(self.projctx, projobj))
```
jorisvandenbossche: In general, it might be preferable to construct the sub CRS from an EPSG number, if available (and fallback on WKT) ?

(we could have a helper function for this, as this pattern happens in a few places)

pyproj/_crs.pyx
```diff
@@ -391,15 +500,56 @@ cdef class _CRS:
...
+        cdef int iii = 0
+        cdef PJ * projobj = proj_crs_get_sub_crs(self.projctx, self.projobj, iii)
+        self._sub_crs_list = []
+        while projobj != NULL:
+            try:
+                self._sub_crs_list.append(self.__class__(_to_wkt(self.projctx, projobj)))
+            finally:
+                proj_destroy(projobj) # deallocate temp proj
+            iii += 1
+            projobj = proj_crs_get_sub_crs(self.projctx, self.projobj, iii)
```
jorisvandenbossche: It might be that this only can be 0 and 1 (for horizontal and vertical crs), so that this extra complexity of the while loop is not needed (at least the compound CRS defined in the database are like that)

pyproj/_crs.pxd
snowman2: I don't disagree, what were you thinking for a name?

jorisvandenbossche: In the C++ code this is called `IdentifiedObject`, if we would like to get ispiration from there, but for me something like `Base` or `Common` is also fine

pyproj/_crs.pyx
snowman2: Good catch

pyproj/_crs.pyx
snowman2: The wkt is usually as reliable as the epsg from my experience. Do you have a use case where it is not?

jorisvandenbossche: Yes, that is true. Should have clarified why I mentioned it: now the string with which the object is created (`srs` attribute) is heavily used in the repr, and for that I find it nicer to have the epsg instead of the first few characters of the wkt. But that's maybe more a repr issue that can also be solved there. 
(and more in general, if there is a epsg number available, this is also shorter than the wkt)

pyproj/_crs.pyx
snowman2: That would be nice if that can be verified. I would be surprised if it was not the case, but always good to check.

pyproj/_crs.pyx
snowman2: Okay, in that case I would prefer the WKT as it is more reliable. The `to_epsg` does a best guess and may not match the WKT perfectly (has a 70% confidence threshold). Though, I agree that having a cleaner srs in the repr would be nice.

snowman2: I guess you could call the `to_epsg(100)` for simplifying the `srs`.

jorisvandenbossche: Yes, just wanted to mention the same that we could use a confidence for 100 for the concern that the epsg might not be a perfect match. If doing that, then at least for the "pefectly defined" CRSses this will work, but for not-fully specified ones (eg from older WKT or proj4 string), it makes sure it uses the correct wkt.

pyproj/_crs.pyx
jorisvandenbossche: So as already mentioned, in the database there are only CompoundCRS pre-defined with 2 components. But checking the proj.4 source code: in the CompoundCRS class itself, it also uses loops to handle the components. So at least in theory you can create a CompoundCRS with more than 2 component CRSses, from C++ to be clear. In the C api however, the creation method for a CompoundCRS just takes the horizontal and vertical CRS. 
But given it is like that in the C++ code, let's leave it here as a loop as well.

snowman2: :+1: Thanks for checking!

pyproj/_crs.pyx
snowman2: I just made the change and removed the `EPSG` from the repr as it will be in the srs_repr and is not necessary anymore.

snowman2:
Another thing to note is that I use horizontal_datum as backup for datum which is useful for the compound CRS.

snowman2:
@jorisvandenbossche, after going through all of this, I am so glad you brought this up. This is much better than it used to be and has a much better design.

pyproj/_crs.pyx
snowman2: I believe this is resolved. If not, feel free to add more comments.

snowman2:
@jswhit @jorisvandenbossche I think it is ready for a final review when you get a moment. Thanks!

snowman2:
I forgot to test the grids - everything but the coordinate operation grids have been tested. So, I will need to test that ...

jorisvandenbossche:
> > Is there a reason that you use here a static method create (and as you do in the other Ellipsoid et al classes), while in the _CRS class you use a more typical __init__ ?
>
> The reason is because the __init__ does not support the PJ_CONTEXT* typing. But, the staticmethod does.

That's indeed that pattern they also explain in the cython docs: note in https://cython.readthedocs.io/en/latest/src/userguide/special_methods.html#initialisation-methods-cinit-and-init, which refers to https://cython.readthedocs.io/en/latest/src/userguide/extension_types.html#existing-pointers-instantiation

pyproj/_crs.pyx
```diff
@@ -117,15 +244,165 @@ cdef class AreaOfUse:
...
+    Ellipsoid for CRS
+
+    Attributes
+    ----------
+    name: str
+        The name of the ellipsoid.
+    is_semi_minor_computed: int
+        1 if True, 0 if False
+    ellipsoid_loaded: bool
+        True if it is loaded without errors.
```
jorisvandenbossche: Also add the `semi_major_metre` and `semi_minor_metre` attributes here?

pyproj/_crs.pyx
```diff
@@ -215,97 +473,439 @@ cdef class Ellipsoid:
...
+            projcontext,
+            coord_operation.projobj,
+            &out_method_name,
+            &out_method_auth_name,
+            &out_method_code
+        )
+        coord_operation.method_name = decode_or_undefined(out_method_name)
+        coord_operation.method_auth_name = decode_or_undefined(out_method_auth_name)
+        coord_operation.method_code = decode_or_undefined(out_method_code)
+        coord_operation.name = "{method_name} [{method_auth_name}:{method_code}]".format(
```
jorisvandenbossche: A CoordinateOperation can actually also have a name, eg:

```
In [36]: crs = pyproj.CRS.from_epsg(31370)                                                                                                                    

In [37]: crs.coordinate_operation.name                                                                                                                        
Out[37]: 'Lambert Conic Conformal (2SP) [EPSG:9802]'

In [38]: crs.coordinate_operation                                                                                                                             
Out[38]: 
CONVERSION["Belgian Lambert 72",
    METHOD["Lambert Conic Conformal (2SP)",
        ID["EPSG",9802]],
    PARAMETER["Latitude of false origin",90,
        ANGLEUNIT["degree",0.0174532925199433],
        ID["EPSG",8821]],
    PARAMETER["Longitude of false origin",4.36748666666667,
        ANGLEUNIT["degree",0.0174532925199433],
        ID["EPSG",8822]],
    PARAMETER["Latitude of 1st standard parallel",51.1666672333333,
        ANGLEUNIT["degree",0.0174532925199433],
        ID["EPSG",8823]],
    PARAMETER["Latitude of 2nd standard parallel",49.8333339,
        ANGLEUNIT["degree",0.0174532925199433],
        ID["EPSG",8824]],
    PARAMETER["Easting at false origin",150000.013,
        LENGTHUNIT["metre",1],
        ID["EPSG",8826]],
    PARAMETER["Northing at false origin",5400088.438,
        LENGTHUNIT["metre",1],
        ID["EPSG",8827]],
    ID["EPSG",19961]]
```
In the first line of the WKT, there is "Belgian Lambert 72", which is the "name" of this EPSG:19961 coordinate operation. 
So maybe we should call the above something else? Or only set it the "name" is not defined? (I suppose this name of the CoordinateOperation can be get with `proj_get_name`)

pyproj/_crs.pyx
```diff
@@ -215,97 +473,439 @@ cdef class Ellipsoid:
...
+        )
+        coord_operation.accuracy = proj_coordoperation_get_accuracy(
+            projcontext,
+            coord_operation.projobj
+        )
+        coord_operation.is_instantiable = proj_coordoperation_is_instantiable(
+            projcontext,
+            coord_operation.projobj
+        )
+        coord_operation.has_ballpark_transformation = \
```
jorisvandenbossche: Maybe we should cast this (and the above as well) to booleans? Because now it are 0/1 ints

pyproj/_crs.pyx
```diff
@@ -215,97 +473,439 @@ cdef class Ellipsoid:
...
+            proj_coordoperation_has_ballpark_transformation(
+                projcontext,
+                coord_operation.projobj
+            )
+
+        # TODO: How do you get the value count?
+        # The example in the PROJ test just says 7 ....
```
jorisvandenbossche: By definition it are always 3 or 7 parameters I think

pyproj/_crs.pyx
```diff
@@ -117,15 +244,165 @@ cdef class AreaOfUse:
...
     def bounds(self):
         return self.west, self.south, self.east, self.north
 
 
-cdef class Ellipsoid:
+cdef class Base:
+    def __cinit__(self):
+        self.projobj = NULL
+        self.projctx = NULL
+        self.name = "undefined"
```
jorisvandenbossche: Should we have a typed definition of those attributes? 

Like 

```
cdef class Base:
    cdef PJ *projobj
    cdef PJ_CONTEXT *projctx
...
```

(I don't know enough from cython to know what benefit this gives, as the current code is also working, but the example I give (and addition to initializing them in (c)init ) is how they show it in the docs)

pyproj/_crs.pyx
```diff
@@ -215,97 +473,439 @@ cdef class Ellipsoid:
...
         self._ellipsoid = None
         self._area_of_use = None
         self._prime_meridian = None
+        self._datum = None
+        self._sub_crs_list = None
+        self._source_crs = None
+        self._coordinate_system = None
+        self._coordinate_operation = None
+
+    def __init__(self, projstring):
```
jorisvandenbossche: Is this `__init__` actually needed? (I mean the user facing functionality part of it). 
Because now that means that you can do both `CRS("..some string..")` as `CRS.from_string("..some string..")` ? Which means we give users two ways to do exactly the same, I would be fine to only have the `from_..` methods as the user facing way to create a CRS object.

(note: this is not really related to the rest of the PR, so can also open an issue to separate that out)

pyproj/_crs.pyx
```diff
@@ -117,15 +244,165 @@ cdef class AreaOfUse:
...
+            projobj
+        )
+        if coord_system.projobj == NULL:
+            return None
+
+        cdef PJ_COORDINATE_SYSTEM_TYPE cs_type = proj_cs_get_type(
+            coord_system.projctx,
+            coord_system.projobj
+        )
+        coord_system.name = _COORD_SYSTEM_TYPE_MAP[cs_type]
```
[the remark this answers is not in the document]
snowman2: I found that the `type` matched the name in the `wkt` and the `proj_get_name` returned undefined. So, this made sense to me. Also, the `name` is used in the `__str__` method and I figured this would best represent the coordinate system name.

pyproj/_crs.pyx
snowman2: Those are documented as properties, so not needed here.

pyproj/_crs.pyx
snowman2: The name in the coordinate operation is stored in `method_name`. The `name` is carries over from `Base`, so I filled it with a name that seemed useful for `CRS.__repr__`.

pyproj/_crs.pyx
snowman2: Sounds like a better user-interfacing type.

pyproj/_crs.pyx
snowman2: My main question is how do we know? Maybe a question for the PROJ maintainers.

pyproj/_crs.pyx
snowman2: it is typed in the `_crs.pxd` file.

pyproj/_crs.pyx
snowman2: Definitely a thought for a separate issue. I have mixed feelings about it at the moment.

snowman2:
@jorisvandenbossche, I like your thinking about the general architecture idea. 👍

snowman2:
So, I gave your idea a try, but it seems the sub WKT strings to PJ objects is not supported:
```python
>>> cc = CoordinateSystem.from_string(crs.coordinate_system.to_wkt())
pyproj.exceptions.CRSError: Invalid projection: CS[Cartesian,2],AXIS["easting (X)",east,ORDER[1],LENGTHUNIT["metre",1,ID["EPSG",9001]]],AXIS["northing (Y)",north,ORDER[2],LENGTHUNIT["metre",1,ID["EPSG",9001]]]: (Internal Proj Error: proj_create: unhandled keyword: CS)
```
In light of that, I think the only way to create the internal Datum, CoordinateSystem, etc..  objects is from an initial CRS object. So with that in mind, I think that I will leave it as is in this PR.

snowman2:
The from_epsg works, so I refactored and added that method to the classes that support it.

pyproj/_crs.pyx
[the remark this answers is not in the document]
jorisvandenbossche: > I found that the `type` matched the name in the `wkt` and the `proj_get_name` returned undefined. So, this made sense to me. Also, the `name` is used in the `__str__` method and I figured this would best represent the coordinate system name.

OK, that makes sense (the reason that I mentioned this, is that in general the `name` is kind of a (somewhat?) unique description of the object, while here you have many entries in the database of the same type. But, I think in general you don't care about those different 2D cartesian coordinate systems, so yes, it sounds good to treat that as the name.

pyproj/_crs.pyx
jorisvandenbossche: > The name in the coordinate operation is stored in `method_name`. The `name` is carries over from `Base`, so I filled it with a name that seemed useful for `CRS.__repr__`.

`method_name` is not the name of the coordinate operation. It is the name of the projection method, as far as I understand the WKT. 
In the example above "Belgian Lambert 72" is the 'name' of the CoordinateOperation (conversion), and "Lambert Conic Conformal (2SP)" is the name of the method (the actual projection)

snowman2: Ah, I see what you mean. Definitely something to look into. Thanks for pointing that out!

pyproj/_crs.pyx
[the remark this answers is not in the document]
snowman2: I will double check on this since the DB has other names, just to be sure.

pyproj/_crs.pyx
jorisvandenbossche: > Definitely a thought for a separate issue. I have mixed feelings about it at the moment.

Opened https://github.com/pyproj4/pyproj/issues/271 for this

pyproj/_crs.pyx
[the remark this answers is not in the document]
jorisvandenbossche: > I will double check on this since the DB has other names, just to be sure.

I think CoordinateSystem is one of the only tables that has no name column:

https://github.com/OSGeo/proj.4/blob/fc395adffc379ec2281288ee4d87a09eca9653bc/data/sql/proj_db_table_defs.sql#L131-L135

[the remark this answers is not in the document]
snowman2: That's unfortunate.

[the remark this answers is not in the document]
jorisvandenbossche: I don't think it's really a problem. We can just either have the `name` as None or use the type of `name` in this case, as you did.

pyproj/_crs.pyx
snowman2: FIxed - I must have mixed this up with CoordinateSystem. Thanks!

pyproj/_crs.pyx
snowman2: Thanks!

pyproj/_crs.pyx
snowman2: Updated, thanks!

pyproj/_crs.pyx
snowman2: https://github.com/OSGeo/proj.4/issues/1430

pyproj/crs.py
```diff
@@ -19,11 +19,22 @@
 NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF OR IN
 CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
 """
-__all__ = ["CRS", "is_wkt"]
+__all__ = [
+    "CRS",
+    "CoordinateOperation",
+    "Datum",
+    "Ellipsoid",
+    "PrimeMeridian",
```
jorisvandenbossche: Is there a reason that CoordinateSystem is not in this list? 
(because there is no way to create it from the class itself?)

jorisvandenbossche: But I see below that you are including them in the sphinx docs (in the `_crs` namespace), so maybe we could add it to this namespace nevertheless? (so the namespace in sphinx is the same as for the other objects)

snowman2:
> For me, the main API question that I am wondering is the CRS class. Is a single class enough, or do we want more classes?

Great idea for another issue.

> But, regarding getting this merged: how do we go about that?

Currently just waiting on @jswhit for approval. Once he approves, it should be good to go.

pyproj/crs.py
snowman2: > because there is no way to create it from the class itself?

Yes, that is correct. No user facing way to create it. So, only access is through the CRS class. I prefer not to include it for clarity of usage. 

> I see below that you are including them in the sphinx docs

I am including it so the user knows what they are getting from the CRS class. Just a reference for them when looking at the return types from the CRS class.

snowman2:
> But, regarding getting this merged: how do we go about that?

Also, want to decide on what to do with the towgs84. I am leaning towards removing it.

jorisvandenbossche:
> Also, want to decide on what to do with the towgs84. I am leaning towards removing it.

It can also certainly be add later, so I would also remove it for now.

snowman2:
Set towgs84 from params instead of the PROJ method.

snowman2:
Sounds good. Thanks for the review @jswhit!

snowman2:
Sounds good, thanks @jswhit!

jorisvandenbossche:
Great, thanks a lot for this work, @snowman2 !

(side remark: given that this are quite some changes, and also additions to the API, should the next version rather be 2.2.0 instead of 2.1.4 ?)

snowman2:
> (side remark: given that this are quite some changes, and also additions to the API, should the next version rather be 2.2.0 instead of 2.1.4 ?)

If we wait to release until after PROJ 6.1.0 (May 1st) and add the option to keep the axis ordering, then bumping to 2.2.0 makes sense to me.

Issue #275, opened 2019-04-18
Undo deprecation of from_crs


snowman2:
It is wrapping `proj_create_crs_from_crs` and makes sense to keep the `from_crs`.

snowman2:
See: https://github.com/pyproj4/pyproj/pull/233

Issue #266, opened 2019-04-11
Provide direction / inverse option in Transformer.transform


jorisvandenbossche:
From discussion with @snowman2 

Currently the `Transformer.transform` method only provides the forward transform. But, we could also support the inverse transform. 
The `proj_trans` functions have a `direction` argument (https://proj4.org/development/reference/functions.html#c.proj_trans)

So it could be an `inverse=False|True` to be consistent with `Proj`. Or a `direction='forward'|'inverse'` to be closer to the proj API, but I think the `inverse` is nicer.

snowman2:
Thanks!

PR #277, opened 2019-04-20
undo deprecation of Transformer.from_crs &  added direction to Transformer

snowman2:
#275 
#266

### DIFF
--- a/pyproj/_transformer.pyx
+++ b/pyproj/_transformer.pyx
@@ -27,8 +27,8 @@ cdef class _Transformer:
         self.projctx = NULL
         self.input_geographic = False
         self.output_geographic = False
-        self._input_radians = False
-        self._output_radians = False
+        self._input_radians = {}
+        self._output_radians = {}
         self.is_pipeline = False
         self.skip_equivalent = False
         self.projections_equivalent = False
@@ -46,16 +46,16 @@ cdef class _Transformer:
             proj_context_destroy(self.projctx)
 
     def _set_radians_io(self):
-        self._input_radians = {
+        self._input_radians.update({
             PJ_FWD: proj_angular_input(self.projpj, PJ_FWD),
             PJ_INV: proj_angular_input(self.projpj, PJ_INV),
             PJ_IDENT: proj_angular_input(self.projpj, PJ_IDENT),
-        }
-        self._output_radians = {
+        })
+        self._output_radians.update({
             PJ_FWD: proj_angular_output(self.projpj, PJ_FWD),
             PJ_INV: proj_angular_output(self.projpj, PJ_INV),
             PJ_IDENT: proj_angular_output(self.projpj, PJ_IDENT),
-        }
+        })
 
     @staticmethod
     def from_crs(crs_from, crs_to, skip_equivalent=False):
--- a/pyproj/_transformer.pyx
+++ b/pyproj/_transformer.pyx
@@ -1,10 +1,16 @@
 include "base.pxi"
 
 from pyproj.crs import CRS
-from pyproj.proj import Proj
 from pyproj.compat import cstrencode, pystrdecode
 from pyproj._datadir cimport get_pyproj_context
 from pyproj.exceptions import ProjError
+
+
+_PJ_DIRECTION_MAP = {
+    "forward": PJ_FWD,
+    "inverse": PJ_INV,
+    "ident": PJ_IDENT,
+}
 
 cdef class _Transformer:
     def __cinit__(self):
@@ -18,6 +24,7 @@ cdef class _Transformer:
         self.skip_equivalent = False
         self.projections_equivalent = False
         self.projections_exact_same = False
+        self.pj_direction = PJ_IDENT
 
     def __init__(self):
         # set up the context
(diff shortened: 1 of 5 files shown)


pyproj/_transformer.pyx
```diff
@@ -30,47 +37,53 @@ cdef class _Transformer:
...
+                "Only {} are supported."
+                .format(direction, tuple(_PJ_DIRECTION_MAP)))
+
+    def _set_radians_io(self):
+        self.input_radians = proj_angular_input(self.projpj, self.pj_direction)
+        self.output_radians = proj_angular_output(self.projpj, self.pj_direction)
 
     @staticmethod
-    def _init_crs_to_crs(proj_from, proj_to, skip_equivalent=False):
+    def from_crs(crs_from, crs_to, skip_equivalent=False, direction="forward"):
```
jorisvandenbossche: The `direction` keyword could also be a keyword in the ``(i)transform`` methods instead of here?

snowman2: I thought about that, but for the interest if performance, I put it the way it is. I there weren't many use cases of a repeated transformation where a direction would need to be changed.

snowman2: It reduces the number of calls to `proj_angular_input` and `proj_angular_output`.

snowman2: But, I have not tested how performance impacting this is, only a theory at the moment.

jorisvandenbossche: Yeah, that's what I wanted ask when I saw your first answer: how does this impact performance? As the argument is not passed to `proj_create_crs_to_crs`, but only when doing the actual in transform in `proj_trans_generic` in the `transform` method.

The advantage of putting it in the transform method is to have it consistent with how PROJ does it (direction is specified when transforming, not when the transform object is created).

jorisvandenbossche: > It reduces the number of calls to proj_angular_input and proj_angular_output.

There are only two possible directions right (or three with ident)? So we could just store the result for both directions of those functions on the object, and then on transform get the correct one?

Or (thinking aloud here), is this is not always just swapped for inverse? 
As storing those attributes on the Transformer object is still useful I think, but should just always be for the forward version?

snowman2: I prefer to keep as much of the logic internally as simple as possible. Even lookups based on the direction have the potential to be costly with repeated transformations. Many users need the performance. My preference would be the simplest and most performant solution. So, probably need some performance testing for different solutions.

snowman2:
Results from the lookup method look pretty performant:
```python
n [1]: from pyproj import Transformer                                         

In [2]: import numpy as np                                                     

In [3]: transformer = Transformer.from_proj(2263, 4326) 
   ...: x_coords = np.random.randint(80000, 120000) 
   ...: y_coords = np.random.randint(200000, 250000)                           

In [4]: %timeit transformer.transform(x_coords, y_coords)                      
5.74 µs ± 15.1 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [5]: transformer = Transformer.from_proj(4326, 2263) 
   ...: x_coords = np.random.randint(80000, 120000) 
   ...: y_coords = np.random.randint(200000, 250000)                           

In [6]: %timeit transformer.transform(x_coords, y_coords, direction="inverse") 
5.93 µs ± 39 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

snowman2:
Changes here: https://github.com/snowman2/pyproj/tree/lookup_trans, will do more tinkering with the idea later.

snowman2:
@jorisvandenbossche, thoughts on the current version?

pyproj/_transformer.pyx
snowman2: Looks like my assumptions were incorrect in this case and the lookup method you provided is a great way to go. It definitely makes the Transformer more flexible and I am glad you brought this up!

pyproj/_transformer.pyx
```diff
@@ -30,38 +45,41 @@ cdef class _Transformer:
         if self.projctx is not NULL:
             proj_context_destroy(self.projctx)
 
-    def set_radians_io(self):
-        self.input_radians = proj_angular_input(self.projpj, PJ_FWD)
-        self.output_radians = proj_angular_output(self.projpj, PJ_FWD)
+    def _set_radians_io(self):
+        self.input_radians = {
```
jorisvandenbossche: those are not public right? (since this in the cython class) 
(just to be sure, as I don't think they should be in their current form)

pyproj/_transformer.pyx
```diff
@@ -30,38 +45,41 @@ cdef class _Transformer:
...
+            PJ_FWD: proj_angular_output(self.projpj, PJ_FWD),
+            PJ_INV: proj_angular_output(self.projpj, PJ_INV),
+            PJ_IDENT: proj_angular_output(self.projpj, PJ_IDENT),
+        }
 
     @staticmethod
-    def _init_crs_to_crs(proj_from, proj_to, skip_equivalent=False):
+    def from_crs(crs_from, crs_to, skip_equivalent=False):
+        crs_from = CRS.from_user_input(crs_from)
+        crs_to = CRS.from_user_input(crs_to)
```
jorisvandenbossche: Minor remark: I would move this input validation to the python class Transformer.from_crs, since it is python code itself (you know that this method is only called from Tranformer.from_crs/from_proj)

pyproj/_transformer.pyx
jorisvandenbossche: Yep, confirmed this myself :)

pyproj/_transformer.pyx
snowman2: Good point - it also allows enforcing the `from_crs` type to be a `_CRS` object. I made the change.

pyproj/_transformer.pyx
snowman2: I think making that private is a good idea. I made the change.

pyproj/_transformer.pyx
snowman2: I believe the current changes resolve this. Thanks again for bringing this up!

Issue #256, opened 2019-04-09
Don't include built documentation in the source files?


jorisvandenbossche:
I noticed that the built html documentation files are included in the repo. I suppose this was done to serve them directly with github pages on https://pyproj4.github.io/pyproj ? 

It's certainly not a big deal, but I thought to just mention (without wanting to seem arrogant or so) that from the experience I have with other python projects that use sphinx, they typically only include the source files (what is in the `/sphinx` repo). 
Including the built html files is not that useful I think (or is it to allow people to browse them locally from a git checkout without building them?), and only adds extra stuff to the git tree (although the docs here are small, so that's probably less of a problem than for bigger projects, where the docs can be several MBs and keeping them in the git tree might lead to big repo sizes)

Anyway, just something small I noticed.

snowman2:
I have seen other projects just have the sphinx pages in the repos. This is a good strategy for large projects or ones using readthedocs for document hosting and definitely cleans up the repo.

You are correct that:
1. The html files are in the repo to serve them up on github pages 
2. The number of pages for the project are relatively small
3. It allows for distributing pre-built release docs with the repo

Having it in the repo also simplifies the process for rebuilding release documentation as we can just rebuild the docs in a MR right before the release.

For a relatively small project, keeping the HTML in the repo is likely okay. However, if the number of docs pages ever gets too excessive (or if enough people prefer not to have the pre-built HTML docs in the repo :) ), github pages has an option to move it to another branch specifically for serving the docs.

jorisvandenbossche:
> Having it in the repo also simplifies the process for rebuilding release documentation as we can just rebuild the docs in a MR right before the release.

I see. You can also use a `gh-pages` branch for this (which I personally find cleaner, it is a separate branch in which you don't necessarily need to preserve history). 
But, it is actually not *that* easy to automate this to push the output of sphinx build to a separate branch (so without tools to do this, I agree the current approach is the simplest).

You have `doctr` which automatically does this using Travis (https://drdoctr.github.io/)

Another option is actually also to use readthedocs? (https://readthedocs.org/) Also that one automatically takes care of building your docs for every commit / every release. 

So with regard of simplifying the process, I think it is worth to consider them. I would happy to set it up (either readthedocs or doctr, have experience with both).

jorisvandenbossche:
[...]

PR #278, opened 2019-04-23
DOC/BLD: set up automated doc builds with travis / doctr

jorisvandenbossche:
Closes https://github.com/pyproj4/pyproj/issues/256

I added it in such a way that the doc build and upload only happens in a single build (using the DOC=true environment variable)


jorisvandenbossche:
Note from `doctr configure` to a repo admin:

```
1. Go to https://github.com/pyproj4/pyproj/settings/keys
   and add the following as a new key:

    ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAACAQC7OqGE1S72vUuOpvA/cv2MYImHu/NgrzBDZYzCzoTekb5nRhkPzBjxsg8TAzm6yM1x1EYiHqNDIqfR/jwEpYvvzkcgrt/I5RjcVNQnVktZe1q/1EkQFblSydnMRFCZMeIyL2ds7MeH4ZgGslA4aUnOc2v5VOMDoLsThVDOjqOBvpm1qktxxxizrSNJgg/MCKweQn0+gOr46/xiBs/bZPfI15oM7c7ug8sr8lSEtzJzGBLFQneAOBMPmxST+LNVP4TwRE29Xh7hhKluC9isanZhZchw7e1591XZegva0lMLszXPTgaT8jRkYXe0RZ//i71RwIaUd5PGswxPYZagc9XS4VoYoLriWXOcVDNc5x2BK7eDr6+7qI/Nk6LALAfrPfmZvZxS/LgZr0M5pq6BGc64/WN2oEwrMUwOJvtcdf8gphVqbg+4GIBrKZkDm54ijLdbGtiZpdzWoNwLTSzphGQMTMUzy94ihqQ9g7nCimK0Qwn2jiC9vFiAaQ8DMttajjQBTXWFhntOOTW3TLkAU3xhrUbMmsohSACxf1QfMQc2gZHUB9ZrkuR4xxiOyAxsZ5iXTDx8lBEzfS6j4qkn2hpS3sOxhBtgjPPqo/H3amF7fV4bRmCL/L4dBb5pmQ4FyDI2U/z8OtxTnJP8HSrXhfm6wwQaWjh4ZF1wSmg/PPWA6Q==
   Be sure to allow write access for the key.
```

See https://drdoctr.github.io/recipes.html#setting-up-doctr-for-a-repo-you-don-t-have-admin-access-to

jorisvandenbossche:
@snowman2 we will only be able to *actually* test it once it is merged in master, since doctr only uploads the docs for commits in master, not in PRs

jorisvandenbossche:
(also changed the sphinx theme, as discussed in the issue)

This should be ready now.

.travis.yml
```diff
@@ -80,6 +83,22 @@ install:
...
   - python -c "import pyproj; pyproj.Proj(init='epsg:4269')"
   - py.test --cov-report term-missing --cov=pyproj -v -s
+  # Building and uploading docs with doctr
+  - set -e
+  - |
+    if [ "$DOC" ]; then
+      pip install sphinx sphinx_rtd_theme
+      cd docs
+      make html
+      cd ..
```
snowman2: What are your thoughts on building the docs outside of the `if` and just be part of the tests?

Also, what are your thoughts on deploying docs in the `after_success` section?

jorisvandenbossche:
> What are your thoughts on building the docs outside of the if and just be part of the tests?

I put it deliberately here, to only build the docs in a single travis build (with the idea: it not needed to build it in each one) 
But for that particular build, it is still "part of the tests", meaning you can check the log output if everything is fine / it will fail if sphinx crashes

> Also, what are your thoughts on deploying docs in the after_success section?

I don't mind much. Looking at the docs, there is a warning about it here: https://drdoctr.github.io/index.html?highlight=after_success#edit-your-travis-file (scroll a bit down). Saying that if you put it in `after_success` it will not cause the build to fail.

snowman2:
Good points. Probably better as is then. Thanks for the clarification!

snowman2:
Thoughts on updating the create_docs.sh and use that for building docs?

jorisvandenbossche:
> Thoughts on updating the create_docs.sh and use that for building docs?

Yep, I actually wanted to propose to move the added code in travis.yml to a separate script. But OK to put that in ci/travis/ then instead of top-level ?

snowman2:
> But OK to put that in ci/travis/ then instead of top-level ?

Are you referring to the entire doctr process or only the sphinx part?

jorisvandenbossche:
Ah, I was referring to the entire process. But you might want to have a script that just builds the docs, as it is now?

jorisvandenbossche:
I am happy to do such a change (update the current `create_docs.sh` for just installing deps + building docs, no doctr) and use that in travis.yml. 
But personally not sure how useful it is to have that in the top-level repo (compared to having a script in travis/ci), as I would assume that somebody locally simply does a `cd docs; make html` ?

snowman2:
It is convenient to have. Much simpler to autocomplete and it is nice have a reproducible way to build docs for newcomers to use as well.

jorisvandenbossche:
I would personally say: then it should not install things (one should try to keep running a script and setting up environment separate IMO)

But, if that's actually the goal, it might be worthwile to add an actual Makefile, so you can eg do `make docs` from the top-level directory. And then we can also provide other development workflows, eg things like `make inplace` as a shortcut for building the C extensions inplace.

jorisvandenbossche:
Anyway, this discussion is getting unrelated to the actual PR :) Will use that script for now, we can always update it in a separate issue/PR.

snowman2:
It would be fantastic to have a Makefile - but that is for another issue as you stated.

jorisvandenbossche:
Updated

.travis.yml
```diff
@@ -80,6 +83,19 @@ install:
 script:
   - python -c "import pyproj; pyproj.Proj(init='epsg:4269')"
   - py.test --cov-report term-missing --cov=pyproj -v -s
+  # Building and uploading docs with doctr
+  - set -e
+  - |
+    if [ "$DOC" ]; then
+      create_docs.sh
```
snowman2: probably need `bash create_docs.sh` here.

jorisvandenbossche:
Note that before merging this, it is best to add the key mentioned above in https://github.com/pyproj4/pyproj/pull/278#issuecomment-485906648

jorisvandenbossche:
@jswhit is this approach fine for you?

snowman2:
Alright, testing now :crossed_fingers:

snowman2:
Looks successful to me. Nice work @jorisvandenbossche!
https://pyproj4.github.io/pyproj/dev/index.html

jorisvandenbossche:
> Looks successful to me.

Great!

Issue #264, opened 2019-04-11
Feedback on the CRS repr


jorisvandenbossche:
More feedback / questions .. :)

<sub>(as some context for this and the other issues I am opening: I am considering to use the pyproj `CRS` class as the actual public user end-point for working with CRSs in GeoPandas, so I am exploring their usage and usability)</sub>

---

When having a CRS object, as a user I am mainly interested in the following questions: what is the epsg number? What is its name? In what unit is it expressed? (degrees or meters) Is this a projected or geographic CRS? What is the area of use? What is the projection being used? What is the Datum of this CRS? ..

Of course, there are APIs to get all this information. And the repr should also try to be an unambiguous representation of the object and its content (from a programmer's perspective). But it is also one of the first encounters of the end user with the object, so it would be nice to make it as user-friendly / informative as possible (which is subjective of course!).

Let's first show an example of the current repr (for a projected CRS constructed from WKT):

```
>>> crs = pyproj.CRS.from_string(pyproj.CRS.from_epsg(3857).to_wkt())
>>> crs
<CRS: PROJCRS["WGS 84 / Pseudo-Mercator", BASEGEOGCRS["W ...>
Name: WGS 84 / Pseudo-Mercator
Ellipsoid:
- semi_major_metre: 6378137.00
- semi_minor_metre: 6356752.31
- inverse_flattening: 298.26
Area of Use:
- name: World - 85°S to 85°N
- bounds: (-180.0, -85.06, 180.0, 85.06)
Prime Meridian:
- longitude: 0.0000
- unit_name: degree
- unit_conversion_factor: 0.01745329
Axis Info:
- Easting[E] (east) EPSG:9001 (metre)
- Northing[N] (north) EPSG:9001 (metre)

>>> crs = pyproj.CRS.from_string(pyproj.CRS.from_epsg(31370).to_wkt())
>>> crs 
<CRS: PROJCRS["Belge 1972 / Belgian Lambert 72",BASEGEOG ...>
Name: Belge 1972 / Belgian Lambert 72
Ellipsoid:
- semi_major_metre: 6378388.00
- semi_minor_metre: 6356911.95
- inverse_flattening: 297.00
Area of Use:
- name: Belgium - onshore
- bounds: (2.5, 49.5, 6.4, 51.51)
Prime Meridian:
- longitude: 0.0000
- unit_name: degree
- unit_conversion_factor: 0.01745329
Axis Info:
- Easting[X] (east) : (metre)
[...]

PR #281, opened 2019-04-24
update CRS repr to support compound CRS, use CRS name, etc

snowman2:
Beginning to address: https://github.com/pyproj4/pyproj/issues/264#issuecomment-484686605

### DIFF
--- a/pyproj/_crs.pxd
+++ b/pyproj/_crs.pxd
@@ -111,6 +111,7 @@ cdef class _CRS(Base):
     cdef PJ_PROJ_INFO projpj_info
     cdef char *pjinitstring
     cdef public object srs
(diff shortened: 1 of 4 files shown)


snowman2:
@jorisvandenbossche I think this is in a good place for an initial review.

On the TODO list is to potentially add a Sub CRS section (for compound CRS) as well as a Source CRS section (for bound CRS) with each displaying the name of the CRS(s).

snowman2:
Added Sub CRS section (for compound CRS) as well as a Source CRS section (for bound CRS) with each displaying the name of the CRS(s).

snowman2:
Thanks for the review!

PR #282, opened 2019-04-24
CI: use --no-use-pep517 to keep using editable install with pip 19.1

jorisvandenbossche:
Change in the recently released pip requires adding this flag if you want to editable install mode in presence of a pyproject.toml file

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -73,7 +73,7 @@ before_install:
 install:
   # coverage report requires a local install
   - pip install "pip>=10.0.1"
-  - PYPROJ_FULL_COVERAGE=YES pip install -e .
+  - PYPROJ_FULL_COVERAGE=YES pip install -e . --no-use-pep517
   - pip install -r requirements-dev.txt
   - pip install coveralls
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -73,9 +73,9 @@ before_install:
 install:
   # coverage report requires a local install
   - pip install "pip>=10.0.1"
-  - pip install -r requirements-dev.txt
-  - PYPROJ_FULL_COVERAGE=YES pip install -e . --no-use-pep517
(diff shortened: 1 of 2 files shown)


jorisvandenbossche:
Hmm, apparently not all builds use the latest pip, so some of them now fail on the option not being available ..

snowman2:
I think in this case, the editable flag is not really necessary in the tests as nothing will be changed. So, my preference would be to remove the `-e`. Side note: the dev requirements are installed after `pip install .` to ensure that the `pyproject.toml` file is working properly. Probably should add a comment for that for clarity.

Thanks for keeping up on this!

jorisvandenbossche:
The comments in the yml files seem to indicate that this is for coverage to be working properly, but honestly not sure if that is true (so if we can remove that, happy to do that!)

snowman2:
Mind giving it a go? I don't think the editable part needs to be there and it would be a simple fix.

jorisvandenbossche:
Will give it a try

> Side note: the dev requirements are installed after pip install . to ensure that the pyproject.toml file is working properly

BTW, I don't think this is needed to test that. If there is a pyproject.toml, pip creates an isolated build environment (although in pyproj's case not fully isolated, as I suppose it will still find the proj.4 from outside), so it should not matter what is already installed in the actual python environment in which you are installing pyproj.

jorisvandenbossche:
So coverage actually needs that the package is installed in editable mode (or you need to specify the path to the site-packages one, but that does not seem like good practice)

So maybe short term let's go with https://github.com/pyproj4/pyproj/pull/283, and then depending on what pip does in a next version, we can update this PR?

snowman2:
I agree. That's unfortunate it didn't work. Thanks for looking into it!

snowman2:
It helps to not already have anything in the build environment to ensure that any user can pip install pyproj without already having cython installed via the pyproject.toml dependency list.

PR #285, opened 2019-04-26
Move built docs to gh-pages branch

jorisvandenbossche:
To accompany PR https://github.com/pyproj4/pyproj/pull/278, which will start pushing the dev docs to a `/dev` directory in the gh-pages branch, this is moving the current built docs in the master branch to the gh-pages branch, so that the https://pyproj4.github.io/pyproj url keeps working the same if the github pages configuration gets switched from master/docs to gh-pages branch.

After the initial switch, we might also want to add a redirect of "stable" to the latest version, and rename the current "html" to "v2.1.3" (as that is what doctr will do in the future for coming releases)


snowman2:
Thanks for doing this, definitely will make the transition easier.

Issue #238, opened 2019-04-03
Use proj_normalize_for_visualization in Transformer for optional axis ordering


snowman2:
See: https://github.com/OSGeo/proj.4/commit/6a7e24dce79f93b73f4919f267df2fdf3ee95713

This will require PROJ >= 6.1.0

snowman2:
Merged into master.

Issue #225, opened 2019-03-27
Update docstrings


mullenkamp:
Hi,

Could you please update the docstring documentation for the Transformer class?
One minor typo is proj_from in the from_proj method, but the most important ones are the itransform and transform methods.

They both state that the input should be in the order of x, y when they actually are in the order of y, x. And the output is in the order of y, x. That got me quite confused for a while.

Thanks

snowman2:
Huh, it is x,y,z. I think it is because it is swapped for some EPSG projections. @rouault @kbevers can you verify my assumption?

snowman2:
@mullenkamp, would you mind providing specific code examples?

snowman2:
Reference to my assumption reasoning:https://github.com/OSGeo/proj.4/issues/1301

mullenkamp:
No problem.

```
from pyproj import CRS, Proj, Transformer

epsg1 = 4326
epsg2 = 2193

x1 = 173.2996473
y1 = -40.60674803

x2 = 1625350
y2 = 5504853

crs1 = CRS.from_epsg(epsg1)
crs2 = CRS.from_epsg(epsg2)

trans1 = Transformer.from_crs(crs1, crs2)

set1a = trans1.transform(y1, x1)

trans2 = Transformer.from_crs(crs2, crs1)

set2a = trans2.transform(y2, x2)
```
These produce the desired output. Reverse them to see strange output (i.e. inf or incorrect coords).

Thanks!

kbevers:
This entirely depend on how the source and destination CRS's are defined. By chance, both your source and destination are defined as having the first coordinate component point in a northerly direction. A general rule of thumb is that a CRS in geographic coordinates has an axis ordering of (latitude, longitude) where as most projected CRS's has axis on ordering of (easting, northing).

`projinfo` can help you understand the CRS's you are working with:

```
$ ./bin/projinfo EPSG:4326
PROJ.4 string:
+proj=longlat +datum=WGS84 +no_defs +type=crs

WKT2_2018 string:
GEOGCRS["WGS 84",
    DATUM["World Geodetic System 1984",
        ELLIPSOID["WGS 84",6378137,298.257223563,
            LENGTHUNIT["metre",1]]],
    PRIMEM["Greenwich",0,
        ANGLEUNIT["degree",0.0174532925199433]],
    CS[ellipsoidal,2],
        AXIS["geodetic latitude (Lat)",north,
            ORDER[1],
[...]

Issue #204, opened 2019-03-17
Update to PROJ release 6.1.0


snowman2:
https://github.com/OSGeo/proj.4/milestone/15
May 1, 2019

snowman2:
Looks like PROJ 6.1.0 release is delayed to May 15th.

snowman2:
Pipelines have been updated and the next release will include 6.1.0.

Issue #269, opened 2019-04-15
Coordinates outside bounds of area of use now return inf


jorisvandenbossche:
Running the GeoPandas tests with proj 6 / pyproj master, turned up this regression:

```
In [1]: import pyproj 

In [2]: pyproj.__version__ 
Out[2]: '2.1.3'

In [3]: p1 = pyproj.Proj(init='epsg:26918')
   ...: p2 = pyproj.Proj(init='epsg:4326')

In [4]: pyproj.transform(p1, p2, 0, 0)
Out[4]: (inf, inf)
```

vs

```
In [1]: import pyproj

In [2]: pyproj.__version__    
Out[2]: '1.9.6'

In [3]: p1 = pyproj.Proj(init='epsg:26918') 
   ...: p2 = pyproj.Proj(init='epsg:4326')  

In [4]: pyproj.transform(p1, p2, 0, 0)
Out[4]: (-79.48874388438705, 0.0)
```

So the transform now returns `inf` instead of an actual value. The coordinates however are out of the area of use (see bounds here http://epsg.io/26918). But previously it did calculate the transformed values nevertheless. 

Here, it was a dummy test case in geopandas where we did not check that the tested dummy values were actually properly inside the bounds of the tested CRS. So that is easy to edit. But, it still is a breaking change (I don't know that what extent people run into this in practice). 

This seems a bit similar to https://github.com/pyproj4/pyproj/issues/202, but the difference is that here pyproj 1.9.6 actually returned values and not inf.

snowman2:
I think this might need to be tested with PROJ master as well. I believe that they added a fix for that. Did you test it with the latest PROJ?

snowman2:
I think that is what you said already, but just checking to be sure.

jorisvandenbossche:
No, tested only with pyproj master, but released proj.4 from conda-forge (didn't set up yet dev tooling for proj.4)

jorisvandenbossche:
Additional comment, there is now the `errcheck` keyword, which properly raises an error:

```
In [5]: pyproj.transform(p1, p2, 0, 0, errcheck=True)                                                                                                                           
...
ProjError: proj_trans_generic error: no system list, errno: 22
```

but, it's not the most informative error message ... Not sure if we can get more information than that from proj4

snowman2:
Not fixed in the latest version. Probably need to raise issue in PROJ.
```python
>>> import pyproj
>>> p1 = pyproj.Proj(init='epsg:26918')
[...]

PR #286, opened 2019-04-30
added always_xy option to transformer & minumum PROJ version to 6.1.0

snowman2:
#225
#238 
#204

### DIFF
--- a/pyproj/transformer.py
+++ b/pyproj/transformer.py
@@ -42,7 +42,7 @@ class Transformer(object):
     """
 
     @staticmethod
-    def from_proj(proj_from, proj_to, skip_equivalent=False):
+    def from_proj(proj_from, proj_to, skip_equivalent=False, always_xy=False):
         """Make a Transformer from a :obj:`~pyproj.proj.Proj` or input used to create one.
 
         Parameters
@@ -54,6 +54,11 @@ class Transformer(object):
         skip_equivalent: bool, optional
             If true, will skip the transformation operation if input and output 
             projections are equivalent. Default is false.
+        always_xy: bool, optional
+            If true, the transform method will accept as input and return as output
+            coordinates using the traditional GIS order, that is longitude, latitude
+            for geographic CRS and easting, northing for most projected CRS.
+            Default is false.
 
         Returns
         -------
@@ -67,12 +72,15 @@ class Transformer(object):
 
         transformer = Transformer()
         transformer._transformer = _Transformer.from_crs(
-            proj_from.crs, proj_to.crs, skip_equivalent=skip_equivalent
+            proj_from.crs,
+            proj_to.crs,
+            skip_equivalent=skip_equivalent,
+            always_xy=always_xy,
         )
         return transformer
 
     @staticmethod
-    def from_crs(crs_from, crs_to, skip_equivalent=False):
+    def from_crs(crs_from, crs_to, skip_equivalent=False, always_xy=False):
         """Make a Transformer from a :obj:`~pyproj.crs.CRS` or input used to create one.
 
         Parameters
@@ -84,6 +92,11 @@ class Transformer(object):
         skip_equivalent: bool, optional
             If true, will skip the transformation operation if input and output
             projections are equivalent. Default is false.
+        always_xy: bool, optional
+            If true, the transform method will accept as input and return as output
+            coordinates using the traditional GIS order, that is longitude, latitude
+            for geographic CRS and easting, northing for most projected CRS.
+            Default is false.
 
         Returns
         -------
@@ -95,6 +108,7 @@ class Transformer(object):
             CRS.from_user_input(crs_from),
             CRS.from_user_input(crs_to),
             skip_equivalent=skip_equivalent,
+            always_xy=always_xy,
         )
         return transformer
 
@@ -309,7 +323,16 @@ class Transformer(object):
 
 
 def transform(
-    p1, p2, x, y, z=None, tt=None, radians=False, errcheck=False, skip_equivalent=False
+    p1,
+    p2,
+    x,
+    y,
+    z=None,
+    tt=None,
+    radians=False,
+    errcheck=False,
+    skip_equivalent=False,
+    always_xy=False,
 ):
     """
     x2, y2, z2 = transform(p1, p2, x1, y1, z1)
@@ -331,7 +354,10 @@ def transform(
     invalid transformation (and no exception is raised).
     If the optional kwarg skip_equivalent is true (default is False), 
     it will skip the transformation operation if input and output 
-    projections are equivalent.
+    projections are equivalent. If `always_xy` is toggled, the 
+    transform method will accept as input and return as output
+    coordinates using the traditional GIS order, that is longitude, latitude
+    for geographic CRS and easting, northing for most projected CRS.
 
     In addition to converting between cartographic and geographic
     projection coordinates, this function can take care of datum
@@ -395,9 +421,9 @@ def transform(
     >>> "%.0f %.0f" % (xeq, yeq)
     '30 60'
     """
-    return Transformer.from_proj(p1, p2, skip_equivalent=skip_equivalent).transform(
-        xx=x, yy=y, zz=z, tt=tt, radians=radians, errcheck=errcheck
-    )
+    return Transformer.from_proj(
+        p1, p2, skip_equivalent=skip_equivalent, always_xy=always_xy
+    ).transform(xx=x, yy=y, zz=z, tt=tt, radians=radians, errcheck=errcheck)
 
 
 def itransform(
@@ -409,6 +435,7 @@ def itransform(
     radians=False,
     errcheck=False,
     skip_equivalent=False,
+    always_xy=False,
 ):
     """
     points2 = itransform(p1, p2, points1)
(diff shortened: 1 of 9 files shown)


snowman2:
TODO: Change build pins to 6.1.0 when PROJ 6.1.0 is released.

snowman2:
Looks like PROJ 6.1.0 release is delayed to May 15th.

pyproj/transformer.py
```diff
@@ -53,6 +53,9 @@ def from_proj(proj_from, proj_to, skip_equivalent=False):
         skip_equivalent: bool, optional
             If true, will skip the transformation operation if input and output 
             projections are equivalent. Default is false.
+        static_xy: bool, optional
+            If true, will force the x,y axis to always be in the same order.
+            Default is false.
```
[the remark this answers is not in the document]
jorisvandenbossche: While writing the above, it might have become more clear :-) 
With  "x,y axis always in the same order", you mean "the *input* and *output* x, y will always be in the same order"? Meaning if the input is in east/north, the output will also be east/north?

Maybe we can just be more explicit? As now the "always same order" is a bit general, while in practice it means "always east, north order" (eg it will never be north, east order for both input/output).

In the PROJ docs, they also use the words "uniform axis order (and thus do not care about axis order mandated by the authority defining the CRS)". 
Also taken from there (slightly modified): If this keyword is toggled, the `transform` method will accept as input and return as output coordinates using the traditional GIS order, that is longitude, latitude (followed by elevation, time) for geographic CRS and easting, northing for most projected CRS.

pyproj/__init__.py
```diff
@@ -47,7 +47,7 @@
 LOSS OF USE, DATA OR PROFITS, WHETHER IN AN ACTION OF CONTRACT,
 NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF OR IN
 CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE. """
-__version__ = "2.1.3"
+__version__ = "2.2.0"
```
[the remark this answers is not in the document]
snowman2: Yeah, I have been thinking about undoing this and making the change closer to the release.

pyproj/transformer.py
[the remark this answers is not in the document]
snowman2: Thanks for the idea - I pretty much took your explanation almost word for word as it is very clear and makes a lot of sense. I left the argument as `static_xy` as I was not sure of a better alternative. I thought of maybe `gis_xy_order` or something, but I currently like `static_xy` better.

pyproj/__init__.py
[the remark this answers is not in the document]
snowman2: This is now undone. Thanks for pointing this out.

pyproj/transformer.py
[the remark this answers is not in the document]
jorisvandenbossche: What about `always_xy` instead of `static_xy` ? 
(that's how you explain it in the transformer.rst: "If you prefer to keep your axis order as always x,y, you can use the ...")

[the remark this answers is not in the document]
snowman2: I like it 👍

snowman2:
So, another thought I had is that in the `pyproj.transform` and `pyproj.itransform` functions, the transformer should always use `always_xy` for compatibility with pyproj 1.9.6. However, the default for `Transformer` will be `always_xy=False` as this was introduced in pyproj 2.1.0 and will be consistent with future behavior.

Thoughts @jswhit @jorisvandenbossche?

snowman2:
Pending a decision on https://github.com/pyproj4/pyproj/pull/286#issuecomment-489674657, I think this is good to go.

We are currently pinned to 6.1.0RC1, which is fine for now as we can update to the release of 6.1.0 before releasing pyproj 2.2.0. Merging this now will allow additional PROJ 6.1.0 specific functionality such as PR #289.

The reduction in coverage is due to PROJ no longer raising the exception & the associated test being removed from our code: See https://github.com/OSGeo/proj.4/issues/1449

test/test_transformer.py
```diff
@@ -238,12 +238,6 @@ def test_itransform_no_error():
     list(itransform(pj, Proj(4326), [(pjx, pjy)], radians=True, errcheck=True))
 
 
-def test_transform_exception():
-    transformer = Transformer.from_proj("+init=epsg:4326", "+init=epsg:27700")
-    with pytest.raises(ProjError):
-        transformer.transform(100000, 100000, errcheck=True)
```
jorisvandenbossche: Maybe keep the test but update the expected result?

jorisvandenbossche:
> So, another thought I had is that in the pyproj.transform and pyproj.itransform functions, the transformer should always use always_xy for compatibility with pyproj 1.9.6. However, the default for Transformer will be always_xy=False as this was introduced in pyproj 2.1.0 and will be consistent with future behavior.

I am not sure this is the case? When specifying a Proj in the "old" way as a proj4 string, it still has the old behaviour of not following the CRS defined axis order):

```
In [1]: import pyproj                                                                                                                                         

In [2]: p1 = pyproj.Proj('+init=epsg:4326')                                                                                                                   

In [3]: p2 = pyproj.Proj('+init=epsg:31370')                                                                                                                  

In [4]: pyproj.transform(p1, p2, 4, 50)                                                                                                                       
Out[4]: (123563.56657479196, 76585.07854289468)

In [5]: p1 = pyproj.Proj('EPSG:4326')                                                                                                                         

In [6]: p2 = pyproj.Proj('EPSG:31370')                                                                                                                        

In [7]: pyproj.transform(p1, p2, 4, 50)                                                                                                                       
Out[7]: (inf, inf)

In [8]: pyproj.transform(p1, p2, 50, 4)                                                                                                                       
Out[8]: (123563.56657479196, 76585.07854289468)
```

Should't that make it backwards compatible (even with `always_xy=False` as the default, as it is now in the PR) ?

But, I think it might be good to also expose the `always_xy` keyword in `pyproj.transform(..)` as well and pass it through to the Transformer creation (and the same for `itransform`).

test/test_transformer.py
snowman2: The purpose of the test is to check the errors raised by proj are caught properly. Maybe another method of failure not already used in the tests could be used here. Ideas welcome.

snowman2:
@jorisvandenbossche, you make a good point there about keeping `pyproj.transformer` as is as your example clearly defines that it is already backwards compatible. I agree that should be the case and I think doing so will prevent future confusion in behavior between the `pyproj.Transformer` and the `pyproj.transformer`.

My original reasoning for the change is that I ran into issues with mixing versions of pyproj and the rasterio.CRS class from an older code base, which is why I brought this up. It uses the `CRS.to_string` method which used to output in the proj4 string format (+init=epsg:4326) and has since been changed to [this](https://github.com/mapbox/rasterio/blob/b904e19e3357ab8b01f3c15f7397a016a3ba734c/rasterio/crs.py#L267-L287) which now uses the epsg:4326 syntax. Somehow the installation of pyproj=1.9.6 and rasterio with conda works (probably the older versions are pinned together) and the installation of pyproj=2.1.3 and rasterio also work with the new `to_string` method and the axis swapped. I think the best solution for this case would be to just update the code I was mentioning and more strict pins :).

Also, I agree that it should be added to `pyproj.transformer`. I was originally thinking about forcing it to be `always_xy`, but your reasoning makes it clear that adding the option is the better path.

I appreciate you taking the time to provide clarity here.

test/test_transformer.py
snowman2: If this still raises an error, maybe this can be used in its place: https://github.com/pyproj4/pyproj/issues/269.

jorisvandenbossche: But what does the code in this test return now with proj 6.1 ?

snowman2: Currently nether raise an error:
```
>>> from pyproj import Transformer
>>> transformer = Transformer.from_proj("+init=epsg:4326", "+init=epsg:27700")
>>> transformer.transform(100000, 100000, errcheck=True)
(inf, inf)
>>> transformer = Transformer.from_proj("+init=epsg:26918", "+init=epsg:4326")
>>> transformer.transform(0, 0, errcheck=True)
(-79.48874388438705, 0.0)
```
So, I guess this means that another method for raising an error will need to be used to keep the test.

jorisvandenbossche: I think it would also be good to keep this test (update with the new result), as it is relevant to test that we get an answer for out of area coordinates.

snowman2: Okay, keeping the test and modifying its purpose to check for `inf` instead of an error works for me.

snowman2: I just made the update.

snowman2:
Okay, I think this one is good to go now.

jorisvandenbossche:
> My original reasoning for the change is that I ran into issues with mixing versions of pyproj and the rasterio.CRS class from an older code base, which is why I brought this up. It uses the CRS.to_string method which used to output in the proj4 string format (+init=epsg:4326) and has since been changed to this which now uses the epsg:4326 syntax.

Ah, yes. Yeah, in the beginning, when "old" and "new" are still both used, this will certainly lead to some confusion. But, in practice this will be a reason to just always specify `always_xy=True`, then you can be sure of the output, regardless of whether you specified the CRS as 'EPSG:4326' or '+init=epsg:4326'.

snowman2:
> in practice this will be a reason to just always specify always_xy=True

This will definitely be the case for geopandas in its current state.

jorisvandenbossche:
> This will definitely be the case for geopandas in its current state.

Yes, but since we typically still use proj.4 dicts or strings, we should normally always get the "old" behaviour (which is what we want), without specifying that keyword (that's the reason that geopandas currently "works" with pyproj 2.1 / proj 6.0). 
But in our current implementation in geopandas, nothing forbids somebody to start doing `gdf.crs = "EPSG:4326"`, which didn't work before, but now would wrongly work. So indeed good to start using this keyword if pyproj >= 2.2 is installed.

PR #287, opened 2019-05-02
CF Fixes: added sweep & check for o_proj=latlon for rotated_latitude_longitude

snowman2:


snowman2:
Thanks for the review!

Issue #280, opened 2019-04-24
Add enumeration objects


snowman2:
It is always nice to have enumeration objects to not use strings where possible.

Current locations where it would be nice:
- to_wkt version option
- to_proj4 version option
- Transformer direction option

micahcochran:
In Python 3.4+ [enum in the standard library](https://docs.python.org/3/library/enum.html).  You can use the backport of [enum34](https://pypi.org/project/enum34/) for Python versions less than 3.4.  However much longer pyproj supports Python 2.7.

PR #288, opened 2019-05-03
added enumeration objects for pyproj

snowman2:
#280

### DIFF
--- a/docs/api/index.rst
+++ b/docs/api/index.rst
@@ -10,5 +10,6 @@ API Documentation
    geod
    transformer
    list
+   enums
    datadir
    exceptions
--- a/pyproj/_crs.pyx
+++ b/pyproj/_crs.pyx
@@ -3,18 +3,22 @@ from collections import OrderedDict
 
 from pyproj.compat import cstrencode, pystrdecode
 from pyproj._datadir cimport get_pyproj_context
+from pyproj.enums import WktVersion, ProjVersion
 from pyproj.exceptions import CRSError
+
 
 cdef cstrdecode(const char *instring):
     if instring != NULL:
         return pystrdecode(instring)
     return None
 
+
 cdef decode_or_undefined(const char* instring):
     pystr = cstrdecode(instring)
     if pystr is None:
         return "undefined"
     return pystr
+
 
 def is_wkt(proj_string):
     """
@@ -33,7 +37,7 @@ def is_wkt(proj_string):
     return proj_context_guess_wkt_dialect(NULL, tmp_string) != PJ_GUESSED_NOT_WKT
 
 
-cdef _to_wkt(PJ_CONTEXT* projctx, PJ* projobj, version="WKT2_2018", pretty=False):
+cdef _to_wkt(PJ_CONTEXT* projctx, PJ* projobj, version=WktVersion.WKT2_2018, pretty=False):
     """
     Convert a PJ object to a wkt string.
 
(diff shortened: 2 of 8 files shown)


jorisvandenbossche:
Personally, as a user of pyproj, I would be -1 on this. This is subjective of course, but I often find string argument options for a keyword easier to user than enums (`crs.to_wkt(version='WKT1_GDAL')` vs `crs.to_wkt(version=pyproj.enums.WktVersion.WKT1_GDAL)`)

But since I will (as a user, not as developer) be mostly using pyproj through geopandas, I won't get into touch with them typically, so you don't have to count on my opinion ;)

jorisvandenbossche:
Quickly tried it out, and it seems that for the WKT version, the string still works, since you still convert the user input to the enum and that's how enums work (which clearly indicates I don't use them often, which might also be the reason that I prefer to not use them ;))

In that case, I would indicate that in the docs. 
And for the direction of the transformer, because there an IntEnum is used, the string version actually doesn't work.

snowman2:
@jorisvandenbossche, thanks for your input. So, one thing to note is that the string version is still supported on the `to_wkt` as well as `to_proj4`. The only one that is not currently supporting strings is `TransformDirection`.

It is nice to have the option for both as it gives a programmer the ability to browse through the options in an editor with autocompletion. It also reduces the possibilities for errors due to typos in strings.

snowman2:
Oh hah - seems we typed at the same time :smile:

snowman2:
Also, @jorisvandenbossche, I believe you can use the values of the enum in transform direction (-1 for inverse) in the direction function. I went with this method as it removed the mapping code to map the direction from the string.

snowman2:
Looks like it works:
```python
>>> from pyproj import Transformer
>>> transformer = Transformer.from_crs(3857, 4326)
>>> transformer.transform(33, 98, direction=-1)
(10909310.09774081, 3895303.9633938945)
```

jorisvandenbossche:
One "regression" in usability I noticed is that before, when passing a string with a typo, you got the options in the message:

```
In [4]: crs.to_wkt(version='WKT_2015')                                                                                                                        
...
ValueError: Invalid version supplied 'WKT_2015'. Only ('WKT2_2015', 'WKT2_2015_SIMPLIFIED', 'WKT2_2018', 'WKT2_2018_SIMPLIFIED', 'WKT1_GDAL', 'WKT1_ESRI') are supported.
```

which is now no longer the case. 

Since it is now the enum `__init__`, there is no easy way to have that again?

For the transform direction, I think the previous string was nicer than an integer (which is not saying much), in the case one is not using the enums of course. 
Could the direction enum also be based on strings? (but that means the mapping between enum and int for the cython code needs to be added back?)

jorisvandenbossche:
> Since it is now the enum __init__, there is no easy way to have that again?

Apparently there are some methods you can override, and one of them is the message for non-existing value: https://docs.python.org/3/library/enum.html#supported-sunder-names (`_missing_` in this case)

Would you welcome a PR adding that?

snowman2:
> Could the direction enum also be based on strings?

If you have a strong preference for it, I am not opposed.

snowman2:
> Apparently there are some methods you can override, and one of them is the message for non-existing value: https://docs.python.org/3/library/enum.html#supported-sunder-names (_missing_ in this case)
Would you welcome a PR adding that?

yes, that would be great

PR #289, opened 2019-05-04
added support for ITRF, compound EPSG, and urn projection strings in CRS

snowman2:


snowman2:
TODO: Add support for additional non-wkt/non-proj initializations: https://github.com/OSGeo/proj.4/commit/292807eee9e1194175b64cb5c0a9f0b432352abc

snowman2:
The failures are due to the fixes in 6.1.0. This will need to wait until after upgrading to 6.1.0.

snowman2:
Moved to #294 as it depends on this change.

Issue #290, opened 2019-05-04
Add support for creating Ellipsoid, Datum, etc from string


snowman2:
See: https://github.com/OSGeo/proj.4/commit/292807eee9e1194175b64cb5c0a9f0b432352abc

snowman2:
This is in master now.

PR #292, opened 2019-05-08
Added from_string to Ellipsoid, Datum, PrimeMeridian, and CoordinateOperation

snowman2:
#290


pyproj/_crs.pyx
```diff
@@ -489,6 +490,34 @@ cdef class Ellipsoid(Base):
         """
         return Ellipsoid.from_authority("EPSG", code)
 
+    @staticmethod
+    def from_string(proj_string):
+        """
+        Create an Ellipsoid from a projection string.
```
jorisvandenbossche: I find it a bit unclear what a "projection string" is, because I would intuitively think about a "proj string", so something like "+ellps=WGS84", which I suppose is not actually supported?

snowman2: I guess I should say a valid ellipsoid projection string. As far as I could tell, it takes a wkt and the urn, but not sure about what else it accepts.

snowman2:
I have been looking at the duplication and attempted a couple of strategies, but ended up back where I am at. I think your idea makes sense and will give it a go later.

pyproj/_crs.pyx
snowman2: I agree examples here are warranted.

jorisvandenbossche: I would also not use the word *projection* in "projection string", because a) an ellipsoid is not a projection and b) it does let me think of a *proj* string, which is not accepted.

snowman2: How about an `ellipsoid string`?

snowman2:
It seems that the class base strategy has the same issues as the function based strategy I tried earlier.

The problem with attempting to reduce the duplication comes due to the fact that to do so would require a `cdef` python function/method . This allows and requires typing the input and output. However, when using the `cdef` python functions/methods, raising exceptions does not have the desired effect.

From: https://notes-on-cython.readthedocs.io/en/latest/misc.html#cdef-exceptions-and-the-return-type

> The Cython documentation says ”...a function declared with cdef that does not return a Python object has no way of reporting Python exceptions to its caller. If an exception is detected in such a function, a warning message is printed and the exception is ignored.”

To be able to raise exceptions requires a mixed python/cython function defined by `def` as in this PR. I like having the exceptions in two spots as it clarifies that the problem is due it either 1) and invalid CRS string versus 2) an invalid CRS string for that type. But, to go the `cdef` route would just return a NULL pointer and the error could be due to either one.

I chose code duplication (even though I always prefer to not do so when possible) as I believe the added clarity to the user was beneficial.

snowman2:
Here is an example of what the alternative would look like:
https://github.com/snowman2/pyproj/commit/a71f1efd3a1efb090b2d24b0f52b29241c56028b

snowman2:
> I chose code duplication (even though I always prefer to not do so when possible) as I believe the added clarity to the user was beneficial.

However, one could argue that there really isn't that much added value to having both exceptions.

pyproj/_crs.pyx
snowman2: I think I resolved it. Let me know if I didn't. Thanks for the comment!

snowman2:
Alright, my brain is finally thinking that the separate exceptions don't bring much value. So, I added a separate commit.

Current state of things:
```
>>> dd = crs.Datum.from_string("epsg:4326")
  File "pyproj/_crs.pyx", line 782, in pyproj._crs.Datum.from_string
    raise CRSError(
pyproj.exceptions.CRSError: Invalid datum string: epsg:4326: (Internal Proj Error: proj_create: missing [)
>>> dd = crs.Datum.from_string("aaaaaa")
  File "pyproj/_crs.pyx", line 782, in pyproj._crs.Datum.from_string
    raise CRSError(
pyproj.exceptions.CRSError: Invalid datum string: aaaaaa: (Internal Proj Error: proj_create: unrecognized format / unknown name)

```

snowman2:
The internal proj error is throwing a random error that does not really make sense though.

pyproj/_crs.pyx
```diff
@@ -489,6 +504,38 @@ cdef class Ellipsoid(Base):
...
+    def from_string(ellipsoid_string):
+        """
+        Create an Ellipsoid from a string.
+
+        Examples:
+          - urn:ogc:def:ellipsoid:EPSG::7001
+          - ELLIPSOID["Airy 1830",6377563.396,299.3249646,
+                LENGTHUNIT["metre",1],
+                ID["EPSG",7001]]
+        Parameters
```
jorisvandenbossche: Blank line needed between the example and the Parameters (rst / sphinx is quit picky about white space)

pyproj/_crs.pyx
jorisvandenbossche: > How about an ellipsoid string?

Sounds good! And certainly the examples help a lot!

jorisvandenbossche:
Last version looks good!

snowman2:
Updated version with more precise error handling where it clears out any other old errors beforehand:
```
>>> from pyproj import crs
>>> dd = crs.Datum.from_string("epsg:4326")
  File "pyproj/_crs.pyx", line 784, in pyproj._crs.Datum.from_string
    raise CRSError(
pyproj.exceptions.CRSError: Invalid datum string: epsg:4326
>>> dd = crs.Datum.from_string("aaaaaa")
  File "pyproj/_crs.pyx", line 784, in pyproj._crs.Datum.from_string
    raise CRSError(
pyproj.exceptions.CRSError: Invalid datum string: aaaaaa: (Internal Proj Error: proj_create: unrecognized format / unknown name)
```

pyproj/_crs.pyx
snowman2: Good catch.

snowman2: I tested with sphinx and needed to make some other fixes, but all should be good now. Thanks!

PR #295, opened 2019-05-10
Prepare for 2.2.0 release

snowman2:


snowman2:
@jswhit, if you would prefer to keep any of the parts I deprecated, just let me know and I would be happy to update the PR.

docs/history.rst
```diff
@@ -1,6 +1,36 @@
 Change Log
 ==========
 
+2.2.0
+~~~~~
+* Minimum PROJ version is 6.1.0
+* `pyproj.crs` updates:
+    * Updated CRS repr (issue #264)
+    * Add Datum, CoordinateSystem, CoordinateOperation clases (issue #262)
```
jorisvandenbossche:
```suggestion
    * Add Datum, CoordinateSystem, CoordinateOperation classes (issue #262)
```

docs/history.rst
```diff
@@ -1,6 +1,36 @@
...
+    * Added CRS mapping to/from CF 1.8 (pull #244)
```
jorisvandenbossche: maybe spell out CF in full? (for those less familiar with the abbreviation)

docs/history.rst
```diff
@@ -1,6 +1,36 @@
...
+    * Added to_proj4_dict (issue #226)
+    * Provide option to "pretty print" WKT (issue #258)
+    * Better handle CompoundCRS (issue #265)
+    * Add consistency for `is_geographic` and `is_projected` (issue #274)
+    * Added check in CRS.to_epsg for when proj_list is null (issue #257)
+    * Disallow creation of non*CRS (eg pipeline) in CRS class (issue #267)
```
jorisvandenbossche:
```suggestion
    * Disallow creation of non-CRS object (eg pipeline) in CRS class (issue #267)
```

docs/history.rst
```diff
@@ -1,6 +1,36 @@
...
+    * Added support for ITRF, compound EPSG, and urn projection strings in CRS (pull #289)
+    * Remove unused _CRS.proj_version from pxd file (pull #255)
```
jorisvandenbossche: Personally, I don't think this is worth mentioning (although I did the PR :-)) 
It more clutters this list. If you keep it, I would say that you actually removed the incorrect CRS.proj_version, as that is the consequence for the user (the user should not need to know about the private _CRS)

docs/history.rst
```diff
@@ -1,6 +1,36 @@
...
+* `pyroj.transformer` updates:
+    * Added `always_xy` option to Transformer (issue #225)
```
jorisvandenbossche: Maybe add a bit of explanation here? (eg copied from the rst docs / docstring) Because I think this is actually a very important change for many users.

docs/history.rst
```diff
@@ -1,6 +1,36 @@
...
+    * Provide direction option in Transformer.transform (issue #266)
+    * Undo deprecation of `from_crs` (issue #275)
+    * Fix false positive errors raised in transformer (issue #249)
+* Fix `pyproj.Proj` initialization from DerivedGeographicCRS (issue #270)
+* Add interface to get the projection/ellps/prime_meridian/units lists (issue #251)
+* Docs/Build/Test fixes:
+    * Set up automated doc builds with travis / doctr (pull #278)
+    * Replace deprecated time.clock() in tests (pull #245)
+    * don't define unused Sphinx html_static_path (pull #248)
```
jorisvandenbossche: Also here, I think the above two are not worth mentioning 

(it depends of course on what the purpose of putting it here is. But I see this mainly as a communication to users (about relevant changes they want to know about and not just a listing of all changes), and then the sphinx html_static_path is irrelevant to them)

snowman2: You are correct.. But, for changes I didn't make, I prefer to add it as it gives an extra pointer to the person who made the change as a minor way of appreciation.

docs/history.rst
snowman2: Yeah, that's a good idea 👍

docs/history.rst
snowman2: Since this is your PR, I am fine with removing it if you prefer.

docs/history.rst
snowman2: Maybe I should update it to be more like xarray: http://xarray.pydata.org/en/stable/whats-new.html

docs/history.rst
snowman2: Good idea. Maybe a link as well.

docs/history.rst
jorisvandenbossche: If appreciation is the main concern (and which is an important concern!), I think adding a list of contributors to this release mentioning their names would actually do more (or at least in addition)

snowman2: That is a good idea. Where is that normally put? Should I add it to the what's new or is there a better spot?

snowman2: Found it for geopandas: https://github.com/geopandas/geopandas/releases

I think I like a mix of what is going on in xarray as well as the release credits of geopandas.

jorisvandenbossche:
>  Where is that normally put? Should I add it to the what's new or is there a better spot?

I don't think there is an universal way to do it. The geopandas is way is a bit less work to do in the end (I can look-up the git command if you want), and the xarray way gives more explicit credit. But not each contribution maps to a single "enhancement" listed there, so having a list of names is also good IMO.

snowman2:
> But not each contribution maps to a single "enhancement" listed there

You are definitely correct there. I think a list of names is definitely important.

snowman2:
> I can look-up the git command if you want

That would be useful if you have it handy.

jorisvandenbossche:
You can do something like `git log v0.4.0.. --format='- %an' | sort | uniq`, but lately I actually use the script in https://github.com/pandas-dev/pandas/blob/master/doc/sphinxext/announce.py. If you copy that in the root of the pyproj repo, you can do `python announce.py v2.1.0..HEAD` to get something similar (it also includes a star for first timers, for the rest should be the same)

snowman2:
I think this is in a good place and future MRs can update the history as they come in before the release.

snowman2:
This is a modified version of the output from `python announce.py v2.1.3rel..HEAD`
```
Contributors
============

A total of 4 people contributed patches to this release.  People with a
"+" by their names contributed a patch for the first time.

* Alan D. Snow
* Chris Mayo
* Joris Van den Bossche +
* Micah Cochran
```

PR #296, opened 2019-05-10
DOC: update installation notes

jorisvandenbossche:
- Linking to Gohlke's wheels is no longer necessary, as those are now uploaded to PyPI as well?
- added some links

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -6,13 +6,13 @@ Installation
 
 The easiest methods for installing pyproj are:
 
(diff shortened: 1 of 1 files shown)


snowman2:
Maybe move the link to a side note that says that the Windows wheels on pypi are provided by Cohlke.

jorisvandenbossche:
I added it back between brackets. Does that look OK? 
I thought an actual note box would draw a bit too much attention, as for the end user it should actually not matter, they should just do `pip install pyproj` (although `conda install pyproj` is probably even better ;))

Issue #297, opened 2019-05-10
API: rename to_geodetic to as_geodetic ?


jorisvandenbossche:
Currently, the CRS has a `to_geodetic` method to get the underlying geographic CRS. For example:

```
In [9]: crs = pyproj.CRS.from_epsg(31370)

In [10]: crs.type_name
Out[10]: 'Projected CRS'

In [12]: crs.to_geodetic()
Out[12]: 
<Geographic 2D CRS: epsg:4313>
Name: Belge 1972
Axis Info [ellipsoidal]:
- Lat[north]: Geodetic latitude (degree)
- Lon[east]: Geodetic longitude (degree)
Area of Use:
- name: Belgium - onshore
- bounds: (2.5, 49.5, 6.4, 51.51)
Datum: Reseau National Belge 1972
- Ellipsoid: International 1924
- Prime Meridian: Greenwich

In [13]: crs.to_geodetic().type_name
Out[13]: 'Geographic 2D CRS'
```

But, the `to_geodetic` is a bit the "weird duck" in the middle of all other `to_..` methods, which all return *some* representation of the current CRS, while `to_geodetic` returns a *different* CRS. 

Therefore, I was thinking, we could also name it `as_geodetic`. Or `get_geodetic`.

snowman2:
I agree 👍

snowman2:
I think making it a property like `geodetic_crs` would be the way to go. It would match `source_crs` and the `sub_crs_list`.

jorisvandenbossche:
Ah, that's also a good option

snowman2:
Addressed in #299

jorisvandenbossche:
[ @snowman2 github sidenote : if you put "closes" before the issue numbers in the PRs, the issues get closed automatically, and you don't need to go close them yourself ]

snowman2:
Nice, I will try it out.

PR #299, opened 2019-05-11
add CRS.geodetic_crs as replacement for CRS.to_geodetic()

snowman2:
#297

### DIFF
--- a/pyproj/_crs.pyx
+++ b/pyproj/_crs.pyx
@@ -1,4 +1,5 @@
 import re
+import warnings
 from collections import OrderedDict
 
 from pyproj.compat import cstrencode, pystrdecode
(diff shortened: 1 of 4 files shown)


snowman2:
Also, made the private attributes private.

pyproj/_crs.pyx
```diff
@@ -1316,20 +1339,14 @@ cdef class _CRS(Base):
...
         """
-        cdef PJ * projobj
-        projobj = proj_crs_get_geodetic_crs(self.projctx, self.projobj)
-        if projobj == NULL:
-            return None
-        try:
-            return self.__class__(_to_wkt(self.projctx, projobj))
-        finally:
-            proj_destroy(projobj) # deallocate temp proj
+        return self.geodetic_crs
```
jorisvandenbossche: Add deprecation warning?

snowman2: Probably better to deprecated it. I had the thought as well.

snowman2: Just added deprecation warning.

Issue #298, opened 2019-05-10
Make transform direction enum string based


snowman2:
https://github.com/pyproj4/pyproj/pull/288#issuecomment-490131704

snowman2:
This will keep the enumeration behavior consistent.

PR #300, opened 2019-05-13
made TransformDirection enum string based; added _missing_ method to raise error when value missing

snowman2:
Addresses #298 and https://github.com/pyproj4/pyproj/pull/288#issuecomment-490122764

Uses idea from https://github.com/pyproj4/pyproj/pull/288#issuecomment-490123157 to better handle invalid input.

@jorisvandenbossche

### DIFF
--- a/pyproj/enums.py
+++ b/pyproj/enums.py
@@ -23,15 +23,6 @@ class WktVersion(Enum):
     #: WKT Version 1 ESRI Style
     WKT1_ESRI = "WKT1_ESRI"
 
-    @classmethod
-    def _missing_(cls, value):
-        raise ValueError(
-            "Invalid version supplied '{}'. "
-            "Only {} are supported.".format(
-                value, tuple(version.value for version in WktVersion)
-            )
-        )
-
 
 class ProjVersion(IntEnum):
     """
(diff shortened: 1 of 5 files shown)


snowman2:
Looks like `_missing_` only works post Python 3.5. Shame, it is really nice.

snowman2:
@jswhit, that sounds good. I will go ahead and merge. Would you be okay with me making a release as well (minus uploading to pypi)?

snowman2:
I was thinking about making a release along with PROJ 6.1.0 release on May 15. Is that okay? Or would you prefer to be able to review everything before then (which is fine - no rush on my end)?

snowman2:
> I was thinking about making a release along with PROJ 6.1.0 release on May 15. Is that okay?

I think that does not make much sense. I think waiting until you are ready would be the better decision in this case. Better to have a good review before the release since there isn't a rush. I was just getting ahead of myself there.

pyproj/enums.py
```diff
@@ -35,14 +35,14 @@ class ProjVersion(IntEnum):
...
 
-class TransformDirection(IntEnum):
+class TransformDirection(Enum):
     """
     Supported transform directions
     """
 
     #: Forward direction
-    FORWARD = 1
+    FORWARD = "FORWARD"
```
jorisvandenbossche: Small remark: we could also opt for lower case? (that might be a little bit nicer in user code (`transform(..., direction='inverse')` vs `transform(..., direction='INVERSE')`)

Just a remark, no strong preference either way!

snowman2: I thought about doing that, and decided that the upper case version is consistent with the upper case WKT enum strings.

jorisvandenbossche: Yep, on the other hand, WKT is an existing abbreviation that is written in upper case, while forward/inverse are normal words.

snowman2: That's true. What would be nice is a case insensitive solution.

snowman2: That would preserve past behavior as that's how I had it before.

PR #302, opened 2019-05-16
made enumeration lookup case insensitive

### LINKED ISSUE #301: Make enums case insensitive
Preserves pre-existing behavior.
https://github.com/pyproj4/pyproj/pull/300#discussion_r284097494

snowman2:
Closes #301

@jorisvandenbossche

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -134,7 +134,7 @@ after_test:
   # - "%CMD_IN_ENV% python setup.py bdist_wininst"
   # - "%CMD_IN_ENV% python setup.py bdist_msi"
   # test wheel
-  - pip install pyproj --ignore-installed --no-index -f dist
+  - pip install pyproj --ignore-installed -f dist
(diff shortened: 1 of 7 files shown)


snowman2:
The strange thing is that the Windows build is not finding aenum. It's a shame as it is nice to have.

jorisvandenbossche:
> The strange thing is that the Windows build is not finding aenum. It's a shame as it is nice to have.

It might be related to the pip install command (as there are for sure universal wheels for all platforms/python versions). The `--no-index` maybe?

snowman2:
> The --no-index maybe?

Maybe, definitely worth looking into.

snowman2:
I didn't look at the full logs earlier - the failure was after the tests. Thanks for the tip @jorisvandenbossche!

hobu:
Can aenum be made optional? It causes some complications for packagers https://lists.debian.org/debian-gis/2019/06/msg00000.html

Issue #271, opened 2019-04-17
API: different (duplicated) ways to create a CRS 


jorisvandenbossche:
From discussion in https://github.com/pyproj4/pyproj/pull/263#discussion_r275495702, moving it here.

The ``CRS`` object has multiple ways to be created:

- `CRS(..)`: the actual class constructor, accepts WKT string, a proj.4 string, a proj.4 dictionary, or with proj.4 keyword arguments (and actually also an `'authority:code'` string, which is not documented)
- `CRS.from_epsg(..)`: accepts an integer EPSG number
- `CRS.from_string(..)`: accepts an EPSG, PROJ, or WKT string
- `CRS.from_user_input(..)`: accepts various input, "dispatches to from_epsg, from_proj, or from_string" according to its documentation. And also accepts objects that have a `to_wkt` method.

So in the other PR, I was wondering if this `__init__` is actually needed? (I mean the user facing functionality part of it). Because now that means that you can do both `CRS("..some string..")` as `CRS.from_string("..some string..")` ? Which means we give users two ways to do exactly the same, I would be fine to only have the `from_..` methods as the user facing way to create a CRS object.

Another observation now I am writing this: there was the idea at one point to have a `from_proj` method? (if we keep the different `from_..` methods, that seems nice to have as well)

I think it would be nice to make a choice ourselves. What is the point of having a `CRS.from_user_input(..)` for the user if `CRS(..)` gives you exactly the same? (or the other way around)

snowman2:
I will need time to think on this one.

snowman2:
Okay, so I haven't made any decisions on this yet, but I figured a brain dump here might help me think though things and also give you more insight into what I am thinking at the same time.

The design for the `pyproj.CRS` initialization was a combination of the `rasterio.CRS` and `Proj` style initializations.

The `__init__` method on the CRS class is nice to have. It goes along with the `Geod` and `Proj` classes initalization methods. My current preference would be to allow the `from_user_input` style in the `__init__` method and just let the user throw whatever they want at it. But, `from_user_input` it does not support all initialization methods such as `from_auth` or `from_cf` (probably could add support for that if we wanted to).

The reasoning for having the current initialization options is for a couple of reasons:

1. To allow the CRS initialization methods be in alignment with the `rasterio.CRS` and likely the future `fiona.CRS` classes. This way, the user can easily switch between using `rasterio.CRS`, `fiona.CRS`, or `pyproj.CRS` in their code bases without much change necessary.
[...]

PR #307, opened 2019-05-17
allow for creation of CRS from user input in init

snowman2:
A potential solution for #271

### DIFF
--- a/pyproj/crs.py
+++ b/pyproj/crs.py
@@ -29,6 +29,7 @@ __all__ = [
 ]
 
 import json
+import re
 import warnings
 
 from pyproj._crs import CoordinateOperation  # noqa
@@ -103,7 +104,9 @@ def _from_string(in_crs_string):
         # look for EPSG, replace with epsg (EPSG only works
         # on case-insensitive filesystems).
         in_crs_string = in_crs_string.replace("+init=EPSG", "+init=epsg").strip()
-
+        # remove no_defs as it does nothing as of PROJ 6.0.0 and breaks
+        # initialization with +init=epsg:...
+        in_crs_string = re.sub(r"\s\+?no_defs([\w=]+)?", "", in_crs_string)
     return in_crs_string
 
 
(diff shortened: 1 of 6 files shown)


snowman2:
This will need tests for the CRS__init__ if this is the direction to proceed in.

snowman2:
Thoughts:
- I chose not to add the CF methods to the user input as I think it is non-standard input.  Also the potential for parameter loss.
- I personally like making the init method accept more combinations as it means I don't have to type `from_user_input`.
- `from_user_input` is good to keep for backwards compatibility as well as compatibility with rasterio.CRS.
- `from_string` is mostly useless except for another point of compatibility with rasterio.CRS. I am thinking about deprecating it.

pyproj/crs.py
```diff
@@ -249,59 +283,100 @@ def from_epsg(cls, code):
...
-            raise CRSError("CRS is empty or invalid: {!r}".format(proj_string))
+        return cls(_from_string(in_crs_string))
 
-        elif proj_string.strip().upper().startswith("EPSG:"):
-            auth, val = proj_string.strip().split(":")
-            if not val:
-                raise CRSError("Invalid CRS: {!r}".format(proj_string))
-            return cls.from_epsg(val)
+    def to_string(self):
+        """Convert the CRS to a string.
```
jorisvandenbossche: My first reaction was: what's the purpose of this? But actually, this could be quite nice. It kind of gives  the "best possible string representation": if possible a "EPSG:.." string, otherwise the string it is based on.

pyproj/crs.py
```diff
@@ -249,59 +283,100 @@ def from_epsg(cls, code):
...
+        Returns
+        -------
+        CRS
+        """
+        if not is_wkt(in_wkt_string):
+            raise CRSError("Invalid WKT string: {}".format(in_wkt_string))
+        return cls(_from_string(in_wkt_string))
+
+    @classmethod
+    def from_string(cls, in_crs_string):
```
jorisvandenbossche: Do we still need this? It seems a bit redundant with `__init__` and `from_user_input`? 
(although it gives some kind of consistency with `to_string` ..)

pyproj/crs.py
```diff
@@ -340,9 +406,24 @@ def get_geod(self):
             in_kwargs["b"] = self.ellipsoid.semi_minor_metre
         return Geod(**in_kwargs)
 
-    def to_proj4_dict(self):
+    @classmethod
+    def from_dict(cls, proj_dict):
```
jorisvandenbossche: I personally think `to/from_proj4_dict` is more explicit / better.

pyproj/crs.py
```diff
@@ -49,10 +50,9 @@
 from pyproj.geod import Geod
 
 
-def _dict2string(projparams):
```
jorisvandenbossche: I found dict2string a clearer name, as it that what it does (it does not return a CRS from a dict). I was actually confused by this, while reading the `_from_string` method below, thinking that it returned a CRS in case of the json string.

pyproj/crs.py
```diff
@@ -62,13 +62,60 @@ def _dict2string(projparams):
...
-        elif not proj_inserted and key in ("init", "proj"):
-            pjargs.insert(0, "+{key}={value}".format(key=key, value=value))
-            proj_inserted = True
         else:
             pjargs.append("+{key}={value}".format(key=key, value=value))
-    return " ".join(pjargs)
+    return _from_string(" ".join(pjargs))
+
+
+def _from_string(in_crs_string):
```
jorisvandenbossche: Something like `_proprocess_string` might be a better descriptor as function name

pyproj/crs.py
snowman2: Yeah, the consistency is what I was going after here.

pyproj/crs.py
snowman2: I had a similar thought process :-)

pyproj/crs.py
snowman2: I agree. That's why I originally went with it. But, I changed it to be consistent with `rasterio.CRS`. Better or worse reasoning :man_shrugging:

PR #309, opened 2019-05-20
_CRS to CCRS to make accessible to user as it is potentially useful for a user

snowman2:
This may be useful for someone who just wants a direct interface to `proj_create` with a slim version of the CRS class:
```
>>> from pyproj.crs import CCRS
>>> cc = CCRS("epsg:4326")
>>> cc
GEOGCRS["WGS 84",
    DATUM["World Geodetic System 1984",
        ELLIPSOID["WGS 84",6378137,298.257223563,
            LENGTHUNIT["metre",1]]],
    PRIMEM["Greenwich",0,
        ANGLEUNIT["degree",0.0174532925199433]],
    CS[ellipsoidal,2],
        AXIS["geodetic latitude (Lat)",north,
            ORDER[1],
            ANGLEUNIT["degree",0.0174532925199433]],
        AXIS["geodetic longitude (Lon)",east,
            ORDER[2],
            ANGLEUNIT["degree",0.0174532925199433]],
    USAGE[
        SCOPE["unknown"],
        AREA["World"],
        BBOX[-90,-180,90,180]],
    ID["EPSG",4326]]
>>> cc.to_proj4()
'+proj=longlat +datum=WGS84 +no_defs +type=crs'
```

### DIFF
--- a/docs/api/crs.rst
+++ b/docs/api/crs.rst
@@ -68,3 +68,12 @@ Prime Meridian
 .. autoclass:: pyproj.crs.PrimeMeridian
     :members:
     :inherited-members:
+
+
+Cython CRS Base Class
+---------------------
+This is for those who want direct access to passing a string to `proj_create`
+without any helpful filtering.
+
+.. autoclass:: pyproj.crs.CCRS
+    :members:
--- a/pyproj/_crs.pxd
+++ b/pyproj/_crs.pxd
@@ -106,7 +106,7 @@ cdef class CoordinateOperation(Base):
     cdef create(PJ* coordinate_operation_pj)
 
 
-cdef class _CRS(Base):
+cdef class CCRS(Base):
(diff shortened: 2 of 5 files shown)


snowman2:
Maybe another part of #271. This would provide a simpler CRS class without any duplicated methods.

jorisvandenbossche:
Do you think of actual use cases for this? 

I am not sure this warrants exposing it to the user, which also means that this interface needs to be kept stable (otherwise the cython class is just an implementation detail we can change however we want)

snowman2:
No particular use case in need at the moment (which probably means it isn't needed). Just an idea that popped into my head when I noticed that the `_CRS` class was becoming pretty useful on its own. The added flexibility of keeping it hidden will likely be very beneficial, so I agree with you there. Thanks for the input! I think tabling this for later is a good idea.

PR #315, opened 2019-05-22
fix is_geocentric for bound crs & add is_vertical, is_engineering to CRS

### LINKED ISSUE #314: fix is_geocentric for bound crs
https://github.com/pyproj4/pyproj/blob/19040c2d1578a5547f8990076c1431b53bd41f5b/pyproj/_crs.pyx#L1664

### LINKED ISSUE #316: Add is_vertical, is_local
Vertical: https://github.com/OSGeo/gdal/blob/60ab2831abf9446c629b3a417a16cb35e7f41e63/gdal/ogr/ogrspatialreference.cpp#L7773-L7802

Local/Engineering: https://github.com/OSGeo/gdal/blob/60ab2831abf9446c629b3a417a16cb35e7f41e63/gdal/ogr/ogrspatialreference.cpp#L7734-L7739

### LINKED ISSUE #317: Check if is_bound for to_authority?
https://github.com/OSGeo/gdal/blob/60ab2831abf9446c629b3a417a16cb35e7f41e63/gdal/ogr/ogrspatialreference.cpp#L7376-L7391

snowman2:
closes #314 
closes #316
closes #317

### DIFF
--- a/pyproj/_crs.pyx
+++ b/pyproj/_crs.pyx
@@ -1440,7 +1440,6 @@ cdef class _CRS(Base):
         Returns
         -------
         list[CRS]
-
         """
         if self._sub_crs_list is not None:
             return self._sub_crs_list
@@ -1463,7 +1462,6 @@ cdef class _CRS(Base):
         Returns
         -------
         pyproj.CRS: The geodetic CRS based on this CRS.
-
         """
         if self._geodetic_crs is not None:
             return self._geodetic_crs if self. _geodetic_crs is not False else None
@@ -1506,7 +1504,6 @@ cdef class _CRS(Base):
         Returns
         -------
         pyproj.CRS: The geodetic CRS from this CRS.
-
         """
         warnings.warn(
             "This method is deprecated an has been replaced with `CRS.geodetic_crs`."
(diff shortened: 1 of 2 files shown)


pyproj/_crs.pyx
```diff
@@ -1599,28 +1599,35 @@ cdef class _CRS(Base):
 
         return None
 
+    def _is_crs_property(self, property_name, property_types, sub_crs_index=0):
+        if self.sub_crs_list:
```
jorisvandenbossche: Can you add a docstring here?

pyproj/_crs.pyx
```diff
@@ -1660,11 +1673,22 @@ cdef class _CRS(Base):
         warnings.warn("CRS.is_valid is deprecated.")
         return self._type != PJ_TYPE_UNKNOWN
 
+    @property
+    def is_local(self):
```
jorisvandenbossche: Do we want to use the "local" instead of "engineering"? (since the "engineering" is the proj.4 / ISO terminology?)

"local" is an easier word, but it could also be confused with "local" in the sense of "use a local projected CRS suited for the study area" advice.

pyproj/_crs.pyx
```diff
@@ -1599,28 +1599,35 @@ cdef class _CRS(Base):
...
+            is_property = self._type in property_types
+        return is_property
+
+
     @property
     def is_geographic(self):
         """
         Returns
         -------
         bool: True if projection in geographic (lon/lat) coordinates.
```
jorisvandenbossche: the "projection" here is not really correct? (as geographic CRS is not projected?)

snowman2:
Those are some good thoughts, thanks for reviewing!

pyproj/_crs.pyx
snowman2: Sounds good to me.

pyproj/_crs.pyx
snowman2: Good point - I will attempt to clarify.

pyproj/_crs.pyx
snowman2: Sure, I will do that.

PR #319, opened 2019-05-24
Only require aenum for python < 3.6 versions

jorisvandenbossche:
Thought to make this dependency conditional, like before for enum34 was also done.


snowman2:
Sounds reasonable to me. Thanks!

PR #324, opened 2019-05-29
 added getting started guide to documentation 

### LINKED ISSUE #320: Use case of the pyproj.Proj class
Copying part of the discussion in https://github.com/pyproj4/pyproj/issues/271 here:

> > me: it is still not fully clear to me what the usecase of pyproj.Proj is
>
> @snowman2: It is an essential class used for transformations in pyproj 1.9.6 and is there at the very least for backwards compatibility. I think most use cases could simply switch to the pyproj.CRS class instead.

I fully understand its use case and importance in pyproj < 2. But I think it is important to make it clear to users (potentially new users that never used the Proj class before) in the documentation what is recommended now. 

Is it only kept for backwards compatibility? If so, then let's state that in the docs, point to use CRS instead and provide users an explanation how to switch to CRS if they were already using Proj before.

Or are there still use cases where you would use Proj instead of CRS?

snowman2:
Closes #320

### DIFF
--- a/docs/api/index.rst
+++ b/docs/api/index.rst
@@ -5,11 +5,11 @@ API Documentation
    :maxdepth: 2
    :caption: Contents:
 
-   proj
    crs
-   geod
    transformer
+   geod
    list
-   enums
    datadir
+   enums
    exceptions
+   proj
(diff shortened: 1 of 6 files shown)


snowman2:
I think this is a good start. Can definitely add to/edit later.

PR #333, opened 2019-06-02
fix testProj4Version for Proj.proj_version being an integer

cjmayo:
```
ERROR: testProj4Version (__main__.BasicTest)
----------------------------------------------------------------------
  File "test/test.py", line 16, in testProj4Version
    assert re.match(r"\d+\.\d+", awips221.proj_version)
  File "/usr/lib/python3.7/re.py", line 173, in match
    return _compile(pattern, flags).match(string)
TypeError: expected string or bytes-like object
```


snowman2:
Oh man. Thanks for catching this! Wouldn't be a proper release without an issue with it :).

PR #348, opened 2019-06-12
Add "stable" doc site pointing to latest release

### LINKED ISSUE #347: Add non-versioned endpoint for sphinx documentation
Right now the generated documentation includes the package version tag in the URL:

https://pyproj4.github.io/pyproj/v2.2.0rel/

This means it is impossible to link to things in the documentation over a long period of time. Most importantly blog posts or github issues talking about pyproj. The one that is effecting me now is that I would like to reference pyproj's CRS object in my sphinx documentation for satpy (https://satpy.readthedocs.io/en/latest/) where we use the builtin `intersphinx` extension to link to an `objects.inv` database that intersphinx creates for other packages.

So in Satpy I configure my package (`conf.py`) with:

```
intersphinx_mapping = {
    'python': ('https://docs.python.org/3', None),
    'numpy': ('https://docs.scipy.org/doc/numpy', None),
    'scipy': ('https://docs.scipy.org/doc/scipy/reference', None),
    'xarray': ('https://xarray.pydata.org/en/stable', None),
    'dask': ('https://docs.dask.org/en/latest', None),
    'pyproj': ('https://pyproj4.github.io/pyproj/v2.2.0rel', None)
}
```

And any links to pyproj in my docs (ex. ``:class:`pyproj.crs.CRS```) will link to the pyproj docs. However, this will break for the next tag or build of the docs where the URL gets updated.

Would it be possible to have a stable or latest URL similar to the ones readthedocs provides to avoid this issue? Sorry if this already exists.

I tested this with:

```
python -m sphinx.ext.intersphinx https://pyproj4.github.io/pyproj/v2.2.0rel/objects.inv
```

jorisvandenbossche:
Closes #347

I am not fully sure this will work, but we do something similar for the pandas docs, so I assume it will (but will only be sure once merged ..)

When making a new release, we need to update where the symlink is pointing.


snowman2:
Thanks :crossed_fingers:

PR #349, opened 2019-06-13
Prepare for 2.2.1 release

snowman2:


docs/gotchas.rst
```diff
@@ -0,0 +1,140 @@
...
+
+    >>> CRS("+init=ESRI:54009") 
+    ...
+    pyproj.exceptions.CRSError: Invalid projection: +init=ESRI:54009 +type=crs: (Internal Proj Error: proj_create: cannot expand +init=ESRI:54009 +type=crs)
+
+
+Proj (Not a generic latitude/longitude to projection converter)
+---------------------------------------------------------------
+
+:class:`~pyproj.proj.Proj` is limited is limited to converting between geographic and
```
jorisvandenbossche: twice "is limited"

docs/gotchas.rst
```diff
@@ -0,0 +1,140 @@
...
+which may not be what you want.
+
+.. code-block:: python
+
+    >>> from pyproj import Proj
+    >>> Proj('epsg:28992')(5.068913, 52.067567)
+    (133148.22970574044, 453192.24450392975)
+    >>> transg = Transformer.from_crs(crs_proj.geodetic_crs, crs_proj)
+    >>> transg.transform(52.067567, 5.068913)
+    (133148.22970574044, 453192.24450392975)
```
jorisvandenbossche: The example here shows twice the same result, so it is not really illustrating the point you are making?

docs/gotchas.rst
snowman2: Thanks! I updated it later. So, the change is in master now.

docs/gotchas.rst
snowman2: The previous results is slightly different than the second set with Proj and the geodetic CRS based transformer (where both should be the same).

jorisvandenbossche: Ah, I need to compare with the numbers of the code block above. Sorry, missed that. Those are indeed different!

PR #358, opened 2019-07-01
add deprecation warning when using +init= syntax

### LINKED ISSUE #355: Add deprecation warning for +init= syntax
From: https://gitter.im/pyproj4-pyproj/community?at=5d13c274b3f22a4b2a4018be

More info about deprecation: https://github.com/pyproj4/pyproj/issues/224#issuecomment-476647824

It probably would be a good idea to add a deprecation warning when users use the `+init=` syntax.

snowman2:
- [x] Closes #355 
 - [x] Tests updated


snowman2:
@jorisvandenbossche, I think this is going to confuse geopandas users unless geopandas uses the `always_xy` option when reprojecting.

pyproj/warn.py
```diff
@@ -0,0 +1,16 @@
+class PyProjWarning(UserWarning):
+    """
+    This is a warning for pyproj
+    """
+
+
+class PyProjDeprecationWarning(PyProjWarning):
```
jorisvandenbossche: Is there a reason you went with a custom warning instead of using the default DeprecationWarning? 

Also, I think it is better to at least subclass from DeprecationWarning, so people can catch just DeprecationWarning.

snowman2: In principle, I agree with you that it could subclass from DeprecationWarning.

However, by default the DeprecationWarning is filtered out:
```python
>>> import warnings
>>> def deprecated():
...     warnings.warn("deprecated", DeprecationWarning)
... 
>>> deprecated()
>>> def user_deprecated():
...     warnings.warn("user_deprecated", UserWarning)
... 
>>> user_deprecated()
__main__:2: UserWarning: user_deprecated

```
And I find that not very useful if you want to give someone heads up that they need to change something.

Also, with a custom instance of a warning, the user can filter based on that specific warning instance.

```python
with warnings.catch_warnings():
    warnings.simplefilter("ignore", ProjDeprecationWarning)
    ...
```

What are your thoughts?

jorisvandenbossche:
> However, by default the DeprecationWarning is filtered out:

They are shown when calling a deprecated functionality directly (already for a long time in IPython, and since Python 3.7 also in the plain python repl, see https://docs.python.org/3/whatsnew/3.7.html#whatsnew37-pep565):

```
Python 3.7.3 | packaged by conda-forge | (default, Mar 27 2019, 23:01:00) 
[GCC 7.3.0] :: Anaconda, Inc. on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import warnings
>>> def deprecated():
...     warnings.warn("deprecated", DeprecationWarning)
... 
>>> deprecated()
__main__:2: DeprecationWarning: deprecated
```

Therefore, in general, DeprecationWarnings are good to signal to developers of other libraries that they need to update their code, before users of those libraries see them (which would be warnings they can't do anything about, as they are from the code in the library using the deprecated feature, which would eg be the case with geopandas).

In pandas, for that reason, we use FutureWarning if it is something we want end-users to see, and sometimes first DeprecationWarnings before turning them into FutureWarnings to give library authors time to fix them.

> Also, with a custom instance of a warning, the user can filter based on that specific warning instance.

jorisvandenbossche:
About the custom class, it is true that it makes it a bit easier to filter them explicitly. Although you can still filter them as well with some of the other arguments of filterwarnings (message or module match). 
But in general, I find using the standard exceptions and warnings cleaner. They are provided by the standard library for a reason. 

Personally, I think one should only go to custom exceptions/warnings if there is a good reason that they don't fit into one of the existing ones or that you might want to specifically catch them.

snowman2:
That makes sense. I will think on it and make a PR later.

PR #363, opened 2019-07-06
added to_proj4 to the pyproj.crs.Base class

snowman2:
I didn't do this before because it didn't always work. But, I don't think that is a great reason to not have it.
Also connected to: https://github.com/OSGeo/PROJ/issues/1548

 - [ ] Tests added (for Transformer, Ellipsoid, etc,)


snowman2:
PROJ strings wont be supported for these classes. Seems like a good idea to not add this for clarity.

PR #366, opened 2019-07-08
Geodesic: Added polygon area computations & support for shapely geometries

### LINKED ISSUE #210: line length, polygon perimeter
Im wondering if a useful addition to pyproj would be geodetic line length for a sequence of points.
An example of the calculation that i wrote recently.
```
from shapely.geometry import LineString
import pyproj
geom = LineString([(2, 0), (2, 4), (3, 4)])
arr = np.asarray(geom)
_geod = pyproj.Geod(ellps="WGS84")
_geod.inv(arr[:-1, 0], arr[:-1, 1], arr[1:, 0], arr[1:, 1])[2].sum()
```
A function built around that last line to support geod inv on a sequence of coordinates rather then pairs of coords is probably whats required.

would also work for polygon perimeters (i wonder how area works on a geod?)
```
from shapely.geometry import Polygon
polygon = Polygon(geom)
#ignoring holes
arr = np.asarray(polygon.exterior)
......
```
I regularly see issues relating to lengths calculated on projections so a nice way to avoid this problem might be well used.

snowman2:
- [x] Closes #210 
 - [x] Tests added

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -63,7 +63,7 @@ lint: ## check style with flake8
 
 check: lint ## flake8 black isort check
 	black --check setup.py pyproj/ test/ docs/
-	isort --recursive -m 3 -w 88 -tc -p test setup.py pyproj/ test/ docs/
+	isort --check --recursive -m 3 -w 88 -tc -p test setup.py pyproj/ test/ docs/
 
 isort: ## order imports
 	isort --recursive -m 3 -w 88 -tc -p test setup.py pyproj/ test/ docs/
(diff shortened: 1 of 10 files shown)


snowman2:
Potentially useful for: https://github.com/geopandas/geopandas/issues/942

snowman2:
Looks like shapely and windows needs assistance for testing.

snowman2:
Hmmm, need to have special case for windows.

Issue #356, opened 2019-06-27
Remove support in `Proj` for operations with angular output?


snowman2:
There have been a couple of issues where users get confused with this:
https://gis.stackexchange.com/questions/326500/pyproj-epsg4326-converts-degrees-to-radians

It is currently not supported in the `proj` program:
```
$ proj +proj=longlat +datum=WGS84 +no_defs
Rel. 6.1.1, July 1st, 2019
<proj>: 
can't initialize operations that produce angular output coordinates
program abnormally terminated
```
```
$ proj +init=epsg:4326
Rel. 6.1.1, July 1st, 2019
<proj>: 
can't initialize operations that produce angular output coordinates
program abnormally terminated
```

snowman2:
Another option would be to use the `Transformer` class under the hood and have it use the CRS.geodetic_crs to CRS as shown in this example: http://pyproj4.github.io/pyproj/stable/gotchas.html#proj-not-a-generic-latitude-longitude-to-projection-converter

snowman2:
The `always_xy` would need to be added to `Proj` as shown in: https://github.com/snowman2/pyproj/commit/3b99e6f1acf8772588956f477256a4cfd7a224d7

snowman2:
Disallowing this would probably be bad for users who use `Proj` in transformations. So, if it was disallowed, it would have to be blocked on invoking `__call__`.

snowman2:
Another option would be to do nothing if it is angular output. Just return what was input.

PR #380, opened 2019-07-29
enable openmp parallel with cython; always return latlon with Proj

### LINKED ISSUE #364: Add parallel processing where it makes sense
https://cython.readthedocs.io/en/latest/src/userguide/parallelism.html

snowman2:
- [x] Closes #364 #356
 - [x] Tests added


snowman2:
Does not appear to be stable. Tests pass on my machine. Depending on OS, it seems different tests fail ...

snowman2:
First, only one failed on windows .. then 3 fail.

snowman2:
Making it optional to enable with default being off. Added warning to docs about instability & being experimental.

jorisvandenbossche:
Cool! Thanks for working on this

Do you have any idea why it is even failing on linux? (that seems like the minimal OS to have it working?)

Do you plan to test it with conda-forge?

---

Some feedback on the Transformer: in contrast with the geodesic and Proj functions, here it is not the actual PROJ call that is parallelized, but only the possible conversion to/from radians (since the actual PROJ call already works on an array level, so the loop does not happen in the pyproj cython code). Did you do any benchmark to see if this part is actually significant in the overall performance? 
Because if it is not, I am not sure adding the prange in that case is necessarily beneficial. Actually, when using eg dask to parallelize this, having it spawn multiple threads when the bulk is operating in a single thread anyway might even slow down things (not an expert here, only guessing). 

To be clear, making those `nogil` is still essential for those. But so in practice, it might be that ensuring that the bulk of the operation (the actual PROJ call to `proj_trans_generic`, assuming this is the bulk) is behind a nogil call is more important (I am not sure this is currently the case?) when trying to speed up this operation (eg in a dask / geopandas setting).

snowman2:
> Do you have any idea why it is even failing on linux?

I am not entirely sure. It seems like I am completely missing something. The behavior is inconsistent. Sometimes it works and sometimes it doesn't.


> Do you plan to test it with conda-forge?

I think I will wait until it is stable. It is off by default because it is unstable. Maybe in a future iteration.

> Transformer

I was planning on digging deeper into this later. Probably open up some issues. One idea was to re-create proj_trans_generic in cython with a parallel call. It would be very similar to PROJ.


Also, good point about the nogil. More functions could probably use them.

This is the beginning stage to hopefully bring it some thought/ideas after some people try it out in various scenarios. The orange can take in the num_threads argument. It might be nice to expose that. But, I was thinking about waiting to expose those until it was more stable.

snowman2:
Reason for instability is that only one thread can use a PJ* at a time. https://github.com/OSGeo/PROJ/blob/f567ee4b8739505716cbb000093d78678a97b170/src/iso19111/c_api.cpp#L338

jorisvandenbossche:
Ah, that's a good reason ... 
Do you know if there is an issue or discussion on the PROJ side about thread safety? Otherwise it might be worth opening one? (unless they already decided it is out of scope)

I think keeping some of the nogil parts of the code is still useful as I mentioned above, will open an issue for that.

PR #383, opened 2019-08-05
Added pyproj.transformer.TransformerGroup to make all transformations available

### LINKED ISSUE #381: Allow user to select transform from available options
Based on the information provided here:
https://lists.osgeo.org/pipermail/proj/2019-July/008740.html

It would be useful to provide an interface to allow the more experienced user select the best transform for their use case instead of the auto-select feature. In addition, allowing the user to refine their search by an area of use would be a nice to have feature.

This could also be expanded to potentially provide an interface to download missing grids into the data directory.

Direct quote:

"""
Applications should notify users when more accurate transformations would be possible, but are not available due to missing grid transformation files

By default, the PROJ 6 API automatically selects the best possible available coordinate operation to use when transforming between a source and destination CRS. In some cases, the default operation created by PROJ may not represent the preferred operation, but rather a secondary choice only. The most common cause of this is due to missing (or incorrectly installed) grid transformation files.

Instead of silently proceeding to utilise a known inferior coordinate operation, it is our recommendation that instead applications should check whether the preferred operation is unavailable, and if so, present a warning to users and advise them of how they can remedy the situation by installing the missing grid transformation file. The QGIS desktop application does this by showing a warning bar in the main application window, alerting users that the preferred transformation was unavailable. Clicking the warning displays a dialog explaining the situation in greater detail, along with links to download the corresponding proj-datumgrid package (where available).

In order to detect this situation, it is necessary for applications to:

1. Create a new PROJ PJ_OPERATION_FACTORY_CONTEXT via a call to proj_create_operation_factory_context. E.g.
 PJ_OPERATION_FACTORY_CONTEXT *operation_factory_context = proj_create_operation_factory_context( 0, 0 ); 
2. Configure the factory to ignore grid file availability when calculating available coordinate operations between a pair of CRSes:  proj_operation_factory_context_set_grid_availability_use( 0, operation_factory_context, PROJ_GRID_AVAILABILITY_IGNORED );
3. Create a list of the available operations between the source and destination CRS objects, using the configured PJ_OPERATION_FACTORY_CONTEXT:  PJ_OBJ_LIST *ops = proj_create_operations( 0, source_crs, dest_crs, operation_factory_context )
4. Iterate through the returned list of available operations using proj_list_get_count and proj_list_get. This list will be ordered by preference - so the most preferred coordinate operation will be the first in the list.
5. Test whether the operation is available for use, via a call to proj_coordoperation_is_instantiable. If the first returned operation is available, it indicates that the preferred coordinate operation to use between the source and destination CRS is available for use on the system.
6. If, however, the first returned operation is not instantiable, it is likely due to the use of a missing grid transformation file. In this case, the application should iterate through the grids used by the operation via calls to proj_coordoperation_get_grid_used_count and proj_coordoperation_get_grid_used. The proj_coordoperation_get_grid_used API helpfully provides a means for retrieving the name, download URL, software license, and other metadata for missing grid files. These details can be used to provide helpful warning messages to users, which also give the user the steps required to download and install the missing grid files.

"""

snowman2:
- [x] Closes #381
 - [x] Tests added

### DIFF
--- a/docs/api/transformer.rst
+++ b/docs/api/transformer.rst
@@ -25,6 +25,14 @@ pyproj.Transformer
 .. autoclass:: pyproj.transformer.Transformer
     :members:
 
+
+pyproj.transformer.TransformerGroup
+-----------------------------------
+
+.. autoclass:: pyproj.transformer.TransformerGroup
+    :members:
+
+
 pyproj.transform
 ----------------
 
--- a/docs/examples.rst
+++ b/docs/examples.rst
@@ -353,6 +353,48 @@ Step 2: Create Transformer to convert from geodetic CRS to CRS
     (3496737.757717311, 743253.9940103051, 5264462.701132784, 2019.0)
 
 
+Transformation Group
+--------------------
+
+The :class:`~pyproj.transformer.TransformerGroup` provides both available
+transformations as well as missing transformations.
+
+1. Helpful if you want to use an alternate transformation and have a good reason for it.
+
+.. code-block:: python
+
+    >>> from pyproj.transformer import TransformerGroup
+    >>> trans_group = TransformerGroup("epsg:4326","epsg:2964")
+    >>> trans_group.best_available
+    True
+    >>> trans_group.transformers[0].transform(66, -153)
+    (149661.2825058747, 5849322.174897663)
+    >>> trans_group.transformers[1].transform(66, -153)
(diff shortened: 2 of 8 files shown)


snowman2:
Windows build cannot find sqlite3 ... not sure why. If persists may need another issue.

Issue #365, opened 2019-07-06
Add future warning for export to PROJ string/dict


snowman2:
Not sure when would be the best time to do this is, but it seems PROJ strings are on their way out. Ref: https://github.com/OSGeo/PROJ/issues/1548#issuecomment-508944612

PR #394, opened 2019-08-18
made init= syntax use DeprecationWarning instead of FutureWarning

snowman2:
Decided against promoting to FutureWarning at this stage.  Probably make it a FutureWarning around the time of #365 or when a PROJ release with its removal is more imminent.


jorisvandenbossche:
And in recent Python versions, DeprecationWarnings do show up when calling functionality that raises them directly (but not when eg calling a geopandas method that calls the deprecated functionality)

PR #395, opened 2019-08-18
search on sys.prefix for the PROJ data directory

### LINKED ISSUE #387: Automatically try to find the data directory?
Fiona has some code (or maybe coming from rasterio originally) to find the data directory of `PROJ`:

https://github.com/Toblerity/Fiona/blob/2cafe743292fb25485a6d320785a5690abfa8eb3/fiona/_env.pyx#L312-L363

and then uses that to automatically set the data directory if not found:

https://github.com/Toblerity/Fiona/blob/2cafe743292fb25485a6d320785a5690abfa8eb3/fiona/_env.pyx#L403-L414

I was wondering if that is functionality we might want to have in pyproj as well.

One caveat, `fiona` only does it if you explicitly use there `Env` context manager (I think), which is eg what other libraries such as geopandas can use, and so fiona does not do it generally by just importing fiona (although I am not fully sure how / if they clean up the set env variables). IN pyproj, we don't really have such concept of context.

One occasion where I regularly bump into the data directory issues is using conda installed pyproj in an environment using a jupyter kernel (where your environment is registered as a kernel, and not running jupyter yourself). In conda, activating an environment sets the env variable, but choosing a kernel in the notebook does not do this, leading to the typical errors if you don't do a `pyproj.set_datapath(..)` manually.

snowman2:
- [x] Closes #387
 - [x] Tests added

### DIFF
--- a/docs/history.rst
+++ b/docs/history.rst
@@ -21,6 +21,7 @@ Change Log
 * Always return latlon with Proj (issue #356)
 * Remove aenum dependency (issue #339)
 * Removed deprecated functions `Proj.proj_version`, `CRS.is_valid`, and `CRS.to_geodetic()` (pull #371)
+* Search on `sys.prefix` for the PROJ data directory (issue #387)
 
 2.2.2
 ~~~~~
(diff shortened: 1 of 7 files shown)


snowman2:
Couldn't reproduce the problem, but I assume this will be a helpful addition nonetheless and will hopefully solve the problem mentioned in the issue.

jorisvandenbossche:
Thanks!

PR #427, opened 2019-09-05
use python3 in stead of python

jdkloe:
explicitely use python3 in the makefile

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -3,7 +3,7 @@
 
 # You can set these variables from the command line.
 SPHINXOPTS    =
-SPHINXBUILD   = python -msphinx
+SPHINXBUILD   = python3 -msphinx
 SPHINXPROJ    = pyproj
 SOURCEDIR     = .
 BUILDDIR      = _build
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 # -*- coding: utf-8 -*-
 #
 # pyproj documentation build configuration file, created by


snowman2:
Thanks @jdkloe! Mind squashing the commits into one commit?

jorisvandenbossche:
> Mind squashing the commits into one commit?

Sidenote: you can do this when merging the PR, which is IMO easier than asking the contributor to do it (and actually, I personally think it is nice practice to squash/merge by default the PRs, which gives a cleaner git history to navigate).

snowman2:
I like the feature and regularly use it for my own PRs. However, if I do it for someone else's PR it also merges my name with the commit. I prefer in the case of someone else's PR to have it clear in the commit history that they made the changes in their commit with a separate commmit in the history that states all I did was merge it in.

jorisvandenbossche:
> However, if I do it for someone else's PR it also merges my name with the commit

Only as the committer, not as the actual author of the commit (git distinguishes both, eg see https://github.com/pandas-dev/pandas/commits/master where it says who did what ). To say it in "github terms", that is not counted as contribution (eg in number of commits in https://github.com/pyproj4/pyproj/graphs/contributors, only the commits you "authored" are counted) 

> I prefer in the case of someone else's PR to have it clear in the commit history that they made the changes in their commit with a separate commmit in the history that states all I did was merge it in.

That is exactly what the "committed" means, it was you who merged (commited) the PR into master. The separate "merge commit" indeed also signifies that, but in a (IMO) more noisy way (and with the disadvantage of needing contributors to squash if you want to have the commits in a PR squashed)

snowman2:
Thanks for the further clarification! The pandas method works as well.

All a matter of preference. I prefer it separate as it is very clear. But, I am not saying that the other method is bad, or that I won't use it, I just don't prefer it.

PR #428, opened 2019-09-05
use python3 in stead of python in the docs Makefile and use python3 o…

jdkloe:
new try, now both changes in one commit.
@snowman2: thanks for your suggestion to squash the commits. Hope this is what you intended.


snowman2:
Thanks!

PR #429, opened 2019-09-07
move doctests out of __init__.py && put show_versions in main

### LINKED ISSUE #417: Move doctests code out of pyproj.__init__
Just thinking it would be good to move all of the test code to the test file.

snowman2:
- [x] Closes #417
 - [x] Tests added

Enables this:
```
$ python -m pyproj

System:
    python: 3.7.3 | packaged by conda-forge | (default, Mar 27 2019, 23:01:00)  [GCC 7.3.0]
executable: /home/snowal/miniconda3/envs/pyproj/bin/python
   machine: Linux-4.15.0-58-generic-x86_64-with-debian-buster-sid

PROJ:
      PROJ: 6.2.0
  data dir: /home/snowal/scripts/pyproj/pyproj/proj_dir/share/proj

Python deps:
    pyproj: 2.3.1
       pip: 19.2.1
setuptools: 41.0.1
    Cython: 0.29.10

```


.travis.yml
```diff
@@ -107,6 +107,7 @@ install:
   - pip install coveralls
 
 script:
+  - python -m pyproj  # test __main__ module
```
jorisvandenbossche: Is this needed? (I think the somewhat longer alternative is `python -c "import pyproj; pyproj.show_versions()"`?) 

Just asking, as the alternative is not really cumbersome to me (and actually more explicit that you are printing the versions), and adding it is yet another corner feature that needs to be supported (as external users might start to use it as well)

snowman2:
@jorisvandenbossche, thanks for taking the time to look into this and share your thoughts! I will let this idea sit for a bit before any final decisions as you have raised some valid concerns. In the meantime, here are some of my current thoughts that may change:

> Is it really needed? ...

I added it because I find myself doing the show versions operation pretty often and chopping off ~35 characters is kinda nice IMO. But, you are correct that it is not explicit. Nonetheless, it provides a concise access point to version information about `pyproj` & PROJ useful to the user similar to the `PROJ` cli:

```
$ ./pyproj/proj_dir/bin/proj
Rel. 6.2.0, September 1st, 2019
usage: proj [-bdeEfiIlmorsStTvVwW [args]] [+opt[=arg] ...] [file ...]
```

> adding it is yet another corner feature that needs to be supported (as external users might start to use it as well)

Well, I certainly hope people use it :). I find debugging pyproj is a common occurrence and making the interface for users to provide information to us a little simpler could be handy IMO.

jorisvandenbossche:
> you are correct that it is not explicit. 

So we could also make it explicit? Eg with a `--versions` parameter (and with an abbreviated form, it should only be a few more characters to type). I think that is a typical pattern with command lines.

But you certainly give good arguments to have it ;)

snowman2:
Interesting, I like those ideas. I will try out an iteration of on those ideas later.

snowman2:
Current iteration:
```
$ python -m pyproj
pyproj version: 2.3.1 [PROJ version: 6.2.0]
usage: __main__.py [-h] [-v]

optional arguments:
  -h, --help     show this help message and exit
  -v, --verbose  Show verbose debugging version information.
$ python -m pyproj -v
pyproj info:
    pyproj: 2.3.1
      PROJ: 6.2.0
  data dir:~/scripts/pyproj/pyproj/proj_dir/share/proj

System:
    python: 3.7.3 | packaged by conda-forge | (default, Mar 27 2019, 23:01:00)  [GCC 7.3.0]
executable: ~/miniconda3/envs/pyproj/bin/python
   machine: Linux-4.15.0-60-generic-x86_64-with-debian-buster-sid

Python deps:
       pip: 19.2.1
setuptools: 41.0.1
    Cython: 0.29.10

```

snowman2:
@jorisvandenbossche thoughts?

PR #437, opened 2019-09-12
release gil in core cython/PROJ code

### LINKED ISSUE #386: Release the GIL in core transformation code
See discussion in https://github.com/pyproj4/pyproj/pull/380#issuecomment-516365850.

By using `nogil` in cython we can ensure to release the GIL in the hot spots of the transformation code, which can ensure that other applications can use pyproj code in parallel (eg using dask)

snowman2:
- [x] Closes #386

@jorisvandenbossche mind trying this out?

### DIFF
--- a/docs/history.rst
+++ b/docs/history.rst
@@ -11,6 +11,7 @@ Change Log
 * Added `scope` & `remarks` to `pyproj.crs` objects and `pyproj.Transformer` (issue #441)
 * Added `operations` to `pyproj.crs.CoordinateOperation` objects and `pyproj.Transformer` (issue #441)
 * Added :func:`~pyproj.get_authorities` and :func:`~pyproj.get_codes` (issue #440)
+* Release gil in core cython/PROJ code (issue #386)
 * BUG: Add checks for unititialized `pyproj.crs` objects to prevent core dumping (issue #433)
 * BUG: added fix for get_transform_crs when checking type (pull #439)
 
--- a/pyproj/_geod.pxd
+++ b/pyproj/_geod.pxd
@@ -6,18 +6,18 @@ cdef extern from "geodesic.h":
   void geod_init(geod_geodesic* g, double a, double f)
   void geod_direct(geod_geodesic* g,
               double lat1, double lon1, double azi1, double s12,
-              double* plat2, double* plon2, double* pazi2)
+              double* plat2, double* plon2, double* pazi2) nogil
   void geod_inverse(geod_geodesic* g,
                double lat1, double lon1, double lat2, double lon2,
-               double* ps12, double* pazi1, double* pazi2)
+               double* ps12, double* pazi1, double* pazi2) nogil
   void geod_lineinit(geod_geodesicline* l,
                geod_geodesic* g,
-               double lat1, double lon1, double azi1, unsigned caps)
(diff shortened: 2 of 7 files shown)


jorisvandenbossche:
@snowman2 Thanks a lot for looking into this!

I think we need to do some timings to check the overhead of releasing the GIL in single threaded cases (releasing/re-acquiring the GIL many times for tiny computations can give quite some overhead). 
(I can try to help with that, but that might take a few days)

snowman2:
Looks fine to me:
```
In [1]: import numpy as np 
   ...: from pyproj import Transformer, transform                               

In [2]: transformer = Transformer.from_proj(2263, 4326) 
   ...: x_coords = np.random.randint(80000, 120000) 
   ...: y_coords = np.random.randint(200000, 250000)                            

In [3]: %timeit transform(2263, 4326, x_coords, y_coords)                       
63.9 ms ± 4.85 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [4]: %timeit transformer.transform(x_coords, y_coords)                       
6.18 µs ± 67.7 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

jorisvandenbossche:
@snowman2 I don't think those timings can tell us much (or at least we should compare them with the timing before this PR). 

I did some timings on a bigger array, and also looked at how it performs when parallelizing: https://nbviewer.jupyter.org/gist/jorisvandenbossche/a97d780c04959812c40b22d6c866f4fa

So I think the conclusion from that can be that the nogil added in `_Proj._fwd` does more harm than good. 
That said, I was wondering: is there something that those `_fwd` and `_inv` methods do differently compared to Transformer? Could they use `Transformer` under the hood? Because in those methods, we iterate over the coordinates and feed them one by one to `proj_trans`, while Transformer also handles arrays of coordinates but passes them in one go as array to `proj_trans_generic`, making this easier to release the GIL.

snowman2:
Interesting! Thanks for taking a look. I just looked at the transformer so it sounds like timing will be needed for the others for verification.

jorisvandenbossche:
For the transformer it is fine, but for `_Proj._fwd`, I would remove it (can do a PR if you want). The geodesic one I didn't time.

snowman2:
I will think on it. I think there is probably a way to wrap the whole loop in a nogil with some better reorganization.

jorisvandenbossche:
See my note above about that. At least conceptually, Proj._inv/fwd seems the same as transform. Does it behave differently is some corner cases?

snowman2:
> Could they use Transformer under the hood?

The Transformer can do essentially the same thing.  Re-directing people to use it instead of Proj is my current plan as I don't know everyone's use case to fully test the transition and mostly have Proj for backwards compatibility and don't plan on adding the same functionality. Currently, only has the original functionality.

jorisvandenbossche:
In that case, I would personally also not do the effort to refactor that for loop to be able to wrap the fool loop in `nogil` (as such a refactor can then also change corner cases), and simply don't bother with `nogil` in this function (`transform` is faster anyway).

snowman2:
Ha, I had the same thought after writing that. If you would like to remove the nogil stuff from Proj, that would be helpful. Thanks!

snowman2:
It was a simple enough fix for Proj, so I added it in #446

PR #443, opened 2019-09-14
start adding versionadded

### LINKED ISSUE #409: Add `.. versionadded::` to docstrings

snowman2:
- [x] Closes #409


snowman2:
Probably missed something, but it is a start. Main convention to add version added to the main class and then add version added to methods/properties if they differ.

PR #446, opened 2019-09-18
fix releasing gil in pyproj.Proj code

snowman2:
Addresses https://github.com/pyproj4/pyproj/pull/437#issuecomment-532093283

Related to: #386


snowman2:
image

snowman2:
image

jorisvandenbossche:
Ah, that was in the end easy to put it around the full for loop, indeed :)

Issue #451, opened 2019-09-19
2.4.0 Release


snowman2:
Release Candidate 0 Status:

https://github.com/pyproj4/pyproj/releases/tag/2.4.rc0

- :-1:  Linux/OSX wheels [PR](https://github.com/pyproj4/pyproj-wheels/pull/16) [python -m pyproj TESTS FAILED] 
- [ ] Windows wheels (ping @cgohlke)
- [x] :+1: Conda Forge [PR](https://github.com/conda-forge/pyproj-feedstock/pull/51)
- :-1: Debian (ping @sebastic)

snowman2:
Wheel errors:
```
Traceback (most recent call last):
  File "/usr/lib/python3.5/runpy.py", line 183, in _run_module_as_main
    mod_...oj/__init__.py", line 70, in <module>
    from pyproj import _datadir
ImportError: cannot import name '_datadir'
```

sebastic:
The Debian package also fails to build due to tests using the `python` interpreter instead of `python3`:
```
=================================== FAILURES ===================================
__________________________________ test_main ___________________________________

    def test_main():
>       output = subprocess.check_output(["python", "-m", "pyproj"]).decode("utf-8")

test/test__main__.py:7: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/usr/lib/python3.7/subprocess.py:395: in check_output
    **kwargs).stdout
/usr/lib/python3.7/subprocess.py:472: in run
    with Popen(*popenargs, **kwargs) as process:
/usr/lib/python3.7/subprocess.py:775: in __init__
    restore_signals, start_new_session)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
```

snowman2:
@sebastic, thanks for running the build.

I am going to iterate a bit and see if I can fix the Linux/OSX builds later. Once those are ready, I will post a new release candidate.

snowman2:
Release Candidate 1 Status:

https://github.com/pyproj4/pyproj/releases/tag/2.4.rc1

- [ ]  Linux/OSX wheels [PR](https://github.com/pyproj4/pyproj-wheels/pull/16)
[...]

PR #452, opened 2019-09-19
Use python3 interpreter.

sebastic:
The Debian package build for 2.4~rc0 failed due to test failures as reported in #451 

Fixed by using the `python3` interpreter, per PEP 394.


snowman2:
I tried this before the release and it failed on Windows as shown in the PR. I am planning to use sys.executable and some other fixes to address this. Thanks!

PR #454, opened 2019-09-19
update to use sys.executable for test in main

snowman2:
#451


snowman2:
Windows wheels behavior is strange. But, tests pass.

PR #458, opened 2019-09-24
Only require gil for pyproj_log_function when needed

snowman2:

### DIFF
--- a/docs/history.rst
+++ b/docs/history.rst
@@ -1,6 +1,10 @@
 Change Log
 ==========
 
+2.4.1
+~~~~~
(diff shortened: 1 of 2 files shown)


snowman2:
Minor improvement:
image

jorisvandenbossche:
> Hmmm, nevermind. Does not seem to be consistent. This change may or may not be beneficial.

As long as the function is not called from a function releasing the GIL, this change can in principle have no effect I think. Declaring a function with 'nogil' is the prerequisite to use that function in a `with nogil` block, but you still need to release the GIL somewhere with such a block

PR #482, opened 2019-11-12
Docs

glostis:
This PR fixes a few typos/broken links in the docs.

I have changed all the `~pyproj.CRS` to `~pyproj.crs.CRS` in the docs because for some reason sphinx autodoc does not seem to understand the former reference, even though there is a
```python
from pyproj.crs import CRS
```
in `pyproj/__init__.py` (see broken link at the last bullet point [here](https://pyproj4.github.io/pyproj/dev/api/crs.html#pyproj.crs.CRS.__init__) for example).

### DIFF
--- a/docs/api/transformer.rst
+++ b/docs/api/transformer.rst
@@ -6,8 +6,8 @@ Transformer
 
 The `pyproj.Transformer` has the capabilities of performing 2D, 3D, and 4D (time)
 transformations. It can do anything that the PROJ command line programs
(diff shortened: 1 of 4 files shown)


snowman2:
Thanks for these fixes! Good catches. 

It would be good to squash these commits into one. Would you prefer to squash these commits yourself or would you prefer for me do it on merge?

glostis:
It's fine with me if you do it on merge.
Thanks!

>

snowman2:
Sounds good 👍. Mind if I add you to the contributors list?

glostis:
Sure, you can add me to the contributor's list!

PR #486, opened 2019-11-14
elevate +init= warning to FutureWarning; add UserWarning to CRS.to_proj4()

### LINKED ISSUE #365: Add future warning for export to PROJ string/dict

snowman2:
- [x] Closes #365
 - [x] Tests added

cc: @jorisvandenbossche @djhoese

### DIFF
--- a/docs/history.rst
+++ b/docs/history.rst
@@ -1,6 +1,11 @@
 Change Log
 ==========
 
+2.4.2
+~~~~~
+* Elevate +init= warning to FutureWarning (pull #486)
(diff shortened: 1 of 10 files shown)


snowman2:
> Do FutureWarnings show up by default?

Yes, they should show up by default from the tests I have done.

PR #489, opened 2019-11-21
add explicit initialisation for all 4 components of projlonlatlin and…

### LINKED ISSUE #481: 6 failing test cases for i686 platform during fedora 32/rawhide build
#### Problem description

During the build of pyproj for Fedora rawhide (32) for pyproj release 2.4.1 I switched on the new pytest test system.
The build went well for all supported platforms, but running the test suite failed on the i686 platform because of 6 failed test cases.
The log file for this build is available on:
https://koji.fedoraproject.org/koji/taskinfo?taskID=38858317

Details for failing tests start at line 631 in the build.log file.

#### Expected Output

all tests should pass

#### Environment Information
py.test-3 output
============================= test session starts ==============================
platform linux -- Python 3.8.0, pytest-4.6.6, py-1.8.0, pluggy-0.12.0
rootdir: /builddir/build/BUILD/pyproj-2.4.1
collected 325 items
...

proj version 6.2.0 (as packaged for fedora-rawhide)

#### Installation method

from source: pyproj-2.4.1.tar.gz as downloaded from pypi

jdkloe:
… projxyin. This fixes the test failures on the 32-bit platform for release 2.4.1

Closes #481 

I found this one by trial and error, and to me it seems this is either an undocumented feature of libproj or just a plain bug. Anyway, for pyproj this solved the test failures that I observed on 32-bit.

### DIFF
--- a/pyproj/_proj.pyx
+++ b/pyproj/_proj.pyx
@@ -86,6 +86,14 @@ cdef class Proj:
                         with gil:
                             raise ProjError("projection_undefined")
                     continue
+
+                # init all 4 fields in the the input PJ_COORD variable
+                # before use, otherwise 32-bit pyproj may show strange errors
+                projlonlatin.v[0] = 0.0
+                projlonlatin.v[1] = 0.0
+                projlonlatin.v[2] = 0.0
+                projlonlatin.v[3] = 0.0
+
                 if proj_angular_input(self.projobj, PJ_FWD):
                     projlonlatin.uv.u = _DG2RAD * lonsdata[iii]
                     projlonlatin.uv.v = _DG2RAD * latsdata[iii]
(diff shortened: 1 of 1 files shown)


snowman2:
Nice catch @jdkloe!

Would you mind trying out:
```python
cdef PJ_COORD projxyin = proj_coord(0, 0, 0, float("nan"))
```
in place of the current solution?

In PROJ they use `HUGE_VAL` for time if it is not provided. Since `Proj` does not support time, I think changing that would be a good idea as well.
 See:
https://github.com/OSGeo/PROJ/blob/8394cacb65cf0b28d75d319f1979f6c2a3cfd7d1/src/4D_api.cpp#L380-L383

PR #500, opened 2019-12-30
update CICD builds with 6.3.0

snowman2:
- [x] Part of #487


snowman2:
Proj failures are OK and related to: https://github.com/OSGeo/PROJ/issues/1808. Need to update to be version specific.

PR #512, opened 2020-01-22
default from_name to search for all authorities & disable approximate match

snowman2:
- default from_name to search for all authorities
    Reasoning: it is used in `from_string` and `from_user_input`, so don't want to restrict it prematurely.
- disable approximate match
    Reasoning: Only 1 result is returned with how it is currently used and being the correct result is important. Sometimes it matches against part of the string and that is not always desirable. Enabling it would be more for a feature to list objects with the name that contain the string.


snowman2:
Again - ignore the Appveyor build for now.

PR #541, opened 2020-02-19
ENH: Added Proj.get_factors & made Proj subclass Transformer

### LINKED ISSUE #503: Meridian Convergence and Scale Factor
Hi All,

I joined GitHub to get help with this issue, related to:
https://github.com/pyproj4/pyproj/issues/144

I want to use pyproj in python to convert coordinates and get the meridian convergence angle and grid scale factor. I figured out from the proj command line to use the -V flag to get these values, but it would be great to be able to get these values directly with pyproj in python.

There was talk of adding this feature into pyproj, but snooping through the scripts I didn't see anything. Is there any plans of still adding it? 

Thanks.

### LINKED ISSUE #533: Drop pyproj.Proj
It is currently around for backwards compatibility with pyproj 1.x. However, no new features will be likely be added.

This is superseded by use of `CRS` and `Transformer` classes:

```python
cc = CRS(...)
transformer = Transformer.from_crs(cc.geodetic_crs, cc)
transformer.transform(...)
```

It is also requires conversion to a PROJ string, which may or may not be reliable:
https://github.com/pyproj4/pyproj/blob/ae074ee4be00e245e23d1e26d0a06c0205af1dae/pyproj/proj.py#L161-L168

The plan would be to deprecate its usage in a 2.x release with deprecation/future warning.

snowman2:
- [x] Closes #533
 - [x] Closes #503
 - [x] Tests added

The factors only work on a Proj - like object. So, should keep Proj around.


snowman2:
Need to do more investigation to ensure this is done properly.

PR #550, opened 2020-02-25
Remove duplicated lines from docstrings in geod.py

glostis:
This PR simply removes what looks like duplicated lines in 2 docstrings of the `pyproj.geod` module.

I've also noticed that three methods of the `Geod` class (namely `fwd`, `inv`, and `npts`) seem to be older than the other methods of the class, and don't have docstrings with properly formatted `Parameters` and `Returns` sections describing the function signatures.
I can update these docstrings if you want!


snowman2:
Thanks!

> I can update these docstrings if you want!

That would be great :+1:

PR #559, opened 2020-03-03
ENH: Added pyproj.transformer.Transform.__eq__()

snowman2:
- [x] Tests added


test/test_transformer.py
```diff
@@ -585,3 +585,11 @@ def test_transformer__area_of_interest__invalid(aoi_data_directory):
...
     with pytest.raises(ProjError):
         TransformerGroup(4326, 2964, area_of_interest=(-136.46, 49.0, -60.72, 83.17))
+
+
+def test_transformer_equals():
+    assert Transformer.from_crs(28356, 7856) == Transformer.from_crs(28356, 7856)
+
+
+def test_transformer_not_equals():
+    assert Transformer.from_crs(28356, 7856) != 22
```
jorisvandenbossche: You should probably also test with a non-equal Transformer object?

snowman2: Sounds like a good idea 👍. If you have a free moment to submit a PR, that would be helpful, but by not required.

PR #566, opened 2020-03-05
bugfix re issue #564, numpy.bool_(False) is not False

### LINKED ISSUE #564: numpy.bool_ comparison yields incorrect results
#### Code Sample, a copy-pastable example if possible
```python
from pyproj import Proj  # pyproj 2.5.0
import numpy as np  # numpy 1.18.1

south = np.array(50) < 0

p = Proj(proj='utm', zone=32, ellipsis='WGS84', datum='WGS84', units='m', south=south)
print(p)
```
Outputs
```Proj('+proj=utm +zone=32 +south +datum=WGS84 +units=m +no_defs', preserve_units=True)```

#### Problem description

When passing numpy.bool_(False) to Proj( ... ,south=south), it activates the south flag anyway. I believe this relates to [CRS.py Line 51-56](https://github.com/pyproj4/pyproj/blob/65105c8f5e8ac76dcca5a19c8e227c9c2eb0fd19/pyproj/crs/crs.py#L51). I think this is because comparing numpy.bool_ types with ```is``` [results in unexpected results](https://github.com/numpy/numpy/issues/5942). 

#### Expected Output
No +south flag
``Proj('+proj=utm +zone=32 +datum=WGS84 +units=m +no_defs', preserve_units=True)``

#### Environment Information
```python
    pyproj: 2.5.0
      PROJ: 6.3.1
  data dir: <>

System:
    python: 3.8.1 (tags/v3.8.1:1b293b6, Dec 18 2019, 23:11:46) [MSC v.1916 64 bit (AMD64)]
executable: <>
   machine: Windows-10-10.0.17763-SP0

Python deps:
       pip: 19.2.3
setuptools: 41.2.0
    Cython: 0.29.15

```
#### Installation method
 - pip wheel

jranalli:
- [x] Closes #564 
 - [x] Tests added

### DIFF
--- a/docs/history.rst
+++ b/docs/history.rst
@@ -9,6 +9,7 @@ Change Log
 * ENH: Updated :attr:`pyproj.crs.CRS.axis_info` to pull all relevant axis information from CRS (issue #557)
 * ENH: Added :meth:`pyproj.transformer.Transform.__eq__` (issue #559)
 * ENH: Added :attr:`pyproj.crs.CRS.utm_zone` (issue #561)
+* BUG: Modify CRS dict test to accomodate numpy bool types. (issue #564)
 
 2.5.0
 ~~~~~
--- a/pyproj/crs/crs.py
+++ b/pyproj/crs/crs.py
@@ -50,7 +50,7 @@ def _prepare_from_dict(projparams: dict) -> str:
         # issue 183 (+ no_rot)
         if value is None or value is True:
             pjargs.append("+{key}".format(key=key))
-        elif value is False:
+        elif str(value) == str(False):
             pass
         else:
             pjargs.append("+{key}={value}".format(key=key, value=value))
--- a/test/test_proj.py
+++ b/test/test_proj.py
@@ -478,3 +478,12 @@ def test_get_factors__errcheck():
     transformer = Proj(3857)
     with pytest.raises(ProjError):
         transformer.get_factors(longitude=40, latitude=70, errcheck=True, radians=True)
+
+
+def test_numpy_bool_kwarg():
+    # Issue 546
+    south = np.array(50) < 0
+    proj = Proj(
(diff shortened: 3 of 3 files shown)


snowman2:
@jranalli, thanks for adding this. Looks good :+1:

Would you like to add a test for your specific use case in `test/test_proj.py` to ensure I this remains stable?

Maybe something like:

```python
from pyproj import Proj
import numpy as np

def test_numpy_bool_kwarg():
    south = np.array(50) < 0
    proj = Proj(proj='utm', zone=32, ellipsis='WGS84', datum='WGS84', units='m', south=south)
    assert "south" not in proj.srs
```

jranalli:
Sure. I would have added this sooner, but the test code was a little more complex than what I've ran into before and I wasn't sure how to. Anyway that's a good simple example and I added it in there. There was some [commentary](https://github.com/numpy/numpy/issues/12950) over at numpy about changing the __repr__ of their bool types, which would break my solution, so keeping it tested is probably a good idea.

There seems to be some type of lint error still showing up that I'm not clear on. Sorry, but I'm a bit new to contributing and working on formal projects that try to be professional. I'd be happy to fix it if you can tell what it's catching.

snowman2:
No worries - gotta start somewhere. Thanks for adding it in.

I think the lint error is the black formatting. You can fix it with: 
```
pip install black
black .
```

jranalli:
Looks like that got it! Thanks for the help.

snowman2:
Great 👍.
Are you familiar with squashing commits? If so, I usually like to let you squash them before merging. Or, if you don't mind me squashing on merge, I can do that.

jranalli:
So I thought I knew how based on instructions that I'd read, but it doesn't appear to have worked teh way I thought. Probably easier if you do it on the merge.

snowman2:
Sounds good. Thanks for your patience in this PR process and submitting a nice fix 👍

PR #570, opened 2020-03-09
Don't silently ignore kwargs when projparams are specified

### LINKED ISSUE #552: Keyword arguments to Proj() are silently ignored when projparams are specified
I came across this when trying to initialize a `Proj()` and include `geoidgrids`:

```
proj = pyproj.Proj('EPSG:3162', geoidgrids='CGG2013an83.gtx')
proj.crs.is_vertical
>>> False
```

This is confusing.  Why is it not taking my geoidgrids?  Did I get the file name/path wrong?

Turns out the positional argument `projparams` is mutually exclusive with any passed `**kwargs`:

https://github.com/pyproj4/pyproj/blob/632c39818e77f8226bbec0db2a3fbe7fc02f7d33/pyproj/proj.py#L143

Yet there is no indication that this is not a valid way to initialize it.

In my opinion it should throw an exception, and it could perhaps be made clearer in the documentation as well.  It's a minor user experience thing, but still something I had to spend time troubleshooting.  Happy to attempt a PR if others agree.

[(stackexchange thread that spawned this)](https://gis.stackexchange.com/q/352277/64255)

gberardinelli:
- [x] Closes #552
 - [x] Tests added

### DIFF
--- a/pyproj/crs/crs.py
+++ b/pyproj/crs/crs.py
@@ -37,9 +37,9 @@ from pyproj.exceptions import CRSError
 from pyproj.geod import Geod
 
 
-def _prepare_from_dict(projparams: dict) -> str:
+def _prepare_from_dict(projparams: dict, allow_json: bool = True) -> str:
     # check if it is a PROJ JSON dict
-    if "proj" not in projparams and "init" not in projparams:
+    if "proj" not in projparams and "init" not in projparams and allow_json:
         return json.dumps(projparams)
     # convert a dict to a proj string.
     pjargs = []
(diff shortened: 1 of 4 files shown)


snowman2:
@gberardinelli this looks great :+1:

Two questions for you:
1. Would you like to squash the commits yourself or do you mind if I do it on merge?
2. Would you mind adding a test in `test/crs/test_crs.py` that is identical to the one you added to test the `Proj` object in `test/test_proj.py` except that it would be a `CRS` object?

gberardinelli:
@snowman2, sorry for my slow responses.   Added a CRS test and squashed.

For whatever reason I'm getting failures from 4 existing tests in `test/crs/test_crs.py`, maybe related to my environment?

snowman2:
Thanks @gberardinelli!

PR #596, opened 2020-04-12
BUG: fix unequal (!=) with non-CRS type

jorisvandenbossche:
See https://github.com/geopandas/geopandas/pull/1339#discussion_r407188802

Currently, you get

```
In [34]: CRS("epsg:4326") != 4326  
Out[34]: True
```

which should be False

 - [ ] Closes #xxxx
 - [x] Tests added

### DIFF
--- a/pyproj/crs/crs.py
+++ b/pyproj/crs/crs.py
@@ -905,6 +905,9 @@ class CRS(_CRS):
     def __eq__(self, other: Any) -> bool:
         return self.equals(other)
 
+    def __ne__(self, other: Any) -> bool:
+        return not self == other
+
     def __reduce__(self) -> Tuple[Type["CRS"], Tuple[str]]:
         """special method that allows CRS instance to be pickled"""
         return self.__class__, (self.srs,)
--- a/test/crs/test_crs.py
+++ b/test/crs/test_crs.py
@@ -1114,6 +1114,13 @@ def test_srs__no_plus():
 
 def test_equals_different_type():
     assert CRS("epsg:4326") != ""
+    assert not CRS("epsg:4326") == ""
+
+    assert CRS("epsg:4326") != 27700
+    assert not CRS("epsg:4326") == 27700
+
+    assert not CRS("epsg:4326") != 4326
+    assert CRS("epsg:4326") == 4326
 
 
 def test_is_exact_same_different_type():


jorisvandenbossche:
I did a fix for CRS, but I suppose that we should in principle need to do it for all classes?

snowman2:
Thanks for the fix!

I wonder if it is an artifact of inheriting from a `cython` class :thinking: 

> I did a fix for CRS, but I suppose that we should in principle need to do it for all classes?

Yes, I think that should be checked that it works. Is that something you have time to add to this PR or should a separate issue be raised?

snowman2:
After some digging, I found out the reasoning. So the `Base` class has an `__eq__` method and I assume that the `__ne__` method was also auto-generated based on that.
https://github.com/pyproj4/pyproj/blob/9d7fb4ca4fc6f1897186ad9f186c653c2283dcb8/pyproj/_crs.pyx#L521-L524
So, when inheriting from the `Base` class, you also get the `__ne__` method that was auto-magically created for it instead of creating it's own `__ne__` method. I tested this out and found that when doing `!=` it calles the base classes `__eq__` method.

However, this is not an issue with any of the other classes, so a fix isn't necessary for them. As such, I think this is good to merge.

Thanks for the fix @jorisvandenbossche!

jorisvandenbossche:
Indeed, the other classes only check actual instances of the class, and don't try to coerce the value being compared to a CRS. 

So 

```
In [12]: crs = CRS("epsg:4326")    

In [13]: crs.datum   
Out[13]: 
DATUM["World Geodetic System 1984",
    ELLIPSOID["WGS 84",6378137,298.257223563,
        LENGTHUNIT["metre",1]],
    ID["EPSG",6326]]

In [14]: crs.datum == pyproj.crs.Datum.from_epsg(6326)   
Out[14]: True

In [16]: crs.datum == "EPSG:6326"   
Out[16]: False

In [17]: crs.datum == 6326     
Out[17]: False
```

So it's only if we would like to make this equality check more flexible here (similar as CRS) that we need to watch out for it.

snowman2:
> So it's only if we would like to make this equality check more flexible here (similar as CRS) that we need to watch out for it.

Interesting, I guess that with the `from_user_input` added to many of the crs module classes, flexibility could be added. My only hesitation would be the `Datum` class as it's `from_name` method could change behavior depending on the type of datum selected. So, it could say they aren't equal when they are. But, that is the only one I would hesitate on.

jorisvandenbossche:
I am personally fine with the other classes being less flexible in this regard (but consistency in behaviour could also be good, of course)

PR #602, opened 2020-04-16
DOC: Added guide for CRS compatibility in python geospatial ecosystem

### LINKED ISSUE #593: Add CRS.to_rasterio()
Related to:
- https://github.com/mapbox/rasterio/issues/1898
- https://github.com/corteva/rioxarray/issues/92

It would only work if you have `rasterio` installed (import rasterio in the method). Essentially, it would return a `rasterio.crs.CRS` object using the WKT format based on the `__gdal_version__`.

snowman2:
- [x] Closes #593

### DIFF
--- a/docs/crs_compatibility.rst
+++ b/docs/crs_compatibility.rst
@@ -1,0 +1,267 @@
+CRS Compatibility Guide for Geospatial Python
+==============================================
+
+This is meant to be a guide to help you along the way of you use :class:`pyproj.crs.CRS`
+with other Python Geospatial libraries.
+
+.. note:: WKT2 is the best format for storing your CRS according to the
+          `PROJ FAQ <https://proj.org/faq.html#what-is-the-best-format-for-describing-coordinate-reference-systems>`__.
+
+
+osgeo/gdal
+----------
+
+https://github.com/osgeo/gdal
+
+Converting from `osgeo.osr.SpatialReference` to `pyproj.crs.CRS`
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. code-block:: python
+
+    from osgeo.osr import SpatialReference
+    from pyproj.crs import CRS
+
+    osr_crs = SpatialReference()
+    osr_crs.ImportFromEPSG(4326)
+    proj_crs = CRS.from_wkt(osr_crs.ExportToWkt())
+
+
+Converting from `pyproj.crs.CRS` to `osgeo.osr.SpatialReference`
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. warning:: WKT2 is only supported in GDAL 3+
+
+.. code-block:: python
+
+    import osgeo
+    from osgeo.osr import SpatialReference
+    from pyproj.crs import CRS
+    from pyproj.enums import WktVersion
+
+    proj_crs = CRS.from_epsg(4326)
+
+    osr_crs = SpatialReference()
+    if osgeo.version_info.major < 3:
+        osr_crs.ImportFromWkt(proj_crs.to_wkt(WktVersion.WKT1_GDAL))
+    else:
+        osr_crs.ImportFromWkt(proj_crs.to_wkt())
+
+
+rasterio
+--------
+
+https://github.com/mapbox/rasterio
+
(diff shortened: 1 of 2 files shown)


docs/crs_compatibility.rst
```diff
@@ -0,0 +1,267 @@
...
+    proj_crs = CRS.from_epsg(4326)
+    if LooseVersion(rasterio.__gdal_version__) < LooseVersion("3.0.0")
+        rio_crs = RioCRS.from_wkt(proj_crs.to_wkt(WktVersion.WKT1_GDAL))
+    else:
+        rio_crs = RioCRS.from_wkt(proj_crs.to_wkt())
+
+fiona
+------
+
+https://github.com/mapbox/rasterio
```
jorisvandenbossche, line 107: wrong link here

docs/crs_compatibility.rst
```diff
@@ -0,0 +1,267 @@
...
+
+https://github.com/osgeo/gdal
+
+Converting from `osgeo.osr.SpatialReference` to `pyproj.crs.CRS`
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. code-block:: python
+
+    from osgeo.osr import SpatialReference
+    from pyproj.crs import CRS
```
jorisvandenbossche, line 22: I think most other examples in the docs do `from pyproj import CRS`, might be good to be somewhat consistent

docs/crs_compatibility.rst
```diff
@@ -0,0 +1,267 @@
...
+    proj_crs = CRS.from_epsg(4326)
+    globe = Globe(
+        ellipse=None,
+        semimajor_axis=proj_crs.ellipsoid.semi_major_metre,
+        semiminor_axis=proj_crs.ellipsoid.semi_minor_metre,
+        inverse_flattening=proj_crs.ellipsoid.inverse_flattening,
+    )
+    proj_dict = proj_crs.to_dict()
+    proj_dict["pm"] = proj_crs.prime_meridian.longitude
+    cart_crs = cCRS(proj_dict, globe=globe)
```
jorisvandenbossche, line 224: Personally I think doing `import cartopy.crs` and then `cartopy.crs.CRS(proj_dcit,globe=globe)` is clearer to read than something as `cCRS` (which is a rather "obscure" abbreviation ;))

docs/crs_compatibility.rst
snowman2: Good catch!

docs/crs_compatibility.rst
snowman2: That is a good point. I changed it to be `import cartopy.crs as ccrs` as that is consistent with what is in their docs.

docs/crs_compatibility.rst
snowman2: I am a big fan of consistency! Here is why some things are a little inconsistent.

The access pattern `from pyproj import CRS` was to be consistent with how things were done in `pyproj` 1 as the `Proj` and `Geod` classes were defined in the `pyproj.__init__.py`. It is also simplifies things when using the `Transformer` class as you only have to do one import `from pyproj import CRS, Transformer`. This is a common access pattern and so it is nice to simplify things a bit there. It is also nice to not have to type an extra 4 characters to import the `CRS` class in the terminal :).

The access pattern `from pyproj.crs import CRS` is consistent with the community `CRS` access patterns (`rasterio.crs` and `cartopy.crs`), which is why I chose to use it in this document. It is also nice to use this pattern as a one line import when accessing the other CRS objects such as `Datum` (https://github.com/pyproj4/pyproj/pull/596#issuecomment-614448662). For example, `from pyproj.crs import CRS, Datum`.

I like using them both personally and they both have their uses. They are both consistent within their own realm of usage. However, they can be inconsistent with each other. :man_shrugging:

snowman2:
@jorisvandenbossche thanks for reviewing the document - that was really helpful!

docs/crs_compatibility.rst
jorisvandenbossche: My opinionated(!) take:

> The access pattern from pyproj.crs import CRS is consistent with the community CRS access patterns

That are packages that are about much more than a CRS, so for them it makes more sense to put it in a submodule, than it does for pyproj, IMO. The CRS is the core object users of pyproj should be using, so I think ideally that is just top-level.

> which is why I chose to use it in this document.

Especially for this guide, with all the different CRS objects coming from different packages, I think doing a `import pyproj` and `pyproj.CRS` might actually be more readable. Because then, it is directly clear that it is about the pyproj CRS object, without needing to check the imports

Issue #599, opened 2020-04-14
PROJ 7.1.0


snowman2:
https://github.com/OSGeo/PROJ/milestone/22

Expected release: July 01, 2020

Relevant to `pyproj`

- https://github.com/OSGeo/PROJ/pull/2144 (`Add proj_degree_input() and proj_degree_output() `)
- https://github.com/OSGeo/PROJ/pull/2065 (`Add proj_get_units_from_database()`)
  - This deprecates `proj_list_angular_units` and `proj_list_linear_units`

Currently, `pyproj` is using these deprecated methods `proj_list_angular_units` and `proj_list_linear_units`. So, I think that `pyproj` 3 should require `PROJ` 7.1+, and update to be future compatible.

mmodenesi:
The required release `7.1.0` is not yet available. When trying to `pip install -e .` following the CONTRIBUTING.md guides, one gets

```
(devel) root@20585a61b60d:/src/pyproj# pip install -e .
Obtaining file:///src/pyproj
  Installing build dependencies ... done
  Getting requirements to build wheel ... error
  ERROR: Command errored out with exit status 1:
   command: /opt/conda/envs/devel/bin/python /opt/conda/envs/devel/lib/python3.8/site-packages/pip/_vendor/pep517/_in_process.py get_requires_for_build_wheel /tmp/tmpyuelewwh
       cwd: /src/pyproj
  Complete output (1 lines):
  ERROR: Minimum supported proj version is 7.1.0, installed version is 7.0.0. For more information see: https://pyproj4.github.io/pyproj/stable/installation.html
  ----------------------------------------
ERROR: Command errored out with exit status 1: /opt/conda/envs/devel/bin/python /opt/conda/envs/devel/lib/python3.8/site-packages/pip/_vendor/pep517/_in_process.py get_requires_for_build_wheel /tmp/tmpyuelewwh Check the logs for full command output.
```

sebastic:
There are two test failures with PROJ 7.1.0-rc2:
```
=================================== FAILURES ===================================
_________________ test_transformer__operations__scope_remarks __________________

    def test_transformer__operations__scope_remarks():
        transformer = TransformerGroup(28356, 7856).transformers[0]
        assert transformer.scope is None
>       assert [op.scope for op in transformer.operations] == [
            None,
            "Conformal transformation of GDA94 coordinates that have been derived through "
            "GNSS CORS.",
            None,
        ]
[...]

PR #608, opened 2020-04-30
Add deprecation warnings for get_units_map and get_angular_units_map

snowman2:
- [x] Related to units changes in #599 (https://github.com/OSGeo/PROJ/pull/2065)
 - [x] Tests updated

### DIFF
--- a/pyproj/_list.pyx
+++ b/pyproj/_list.pyx
@@ -2,6 +2,7 @@ include "proj.pxi"
 
 from collections import namedtuple
 from enum import IntEnum
+import warnings
 
 from pyproj.compat import cstrencode, pystrdecode
 from pyproj.enums import PJType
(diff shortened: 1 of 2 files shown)


pyproj/_list.pyx
```diff
@@ -74,6 +75,12 @@ def get_units_map():
     dict:
         Units supported by PROJ
     """
+    warnings.warn(
+        "The behavior of 'pyproj.get_units_map' is deprecated "
+        "and will change in version 3.0.0.",
```
jorisvandenbossche, line 80: Small comment: it's not very clear what the consequence of this is: what will change? What should the user do instead when seeing this warning?

snowman2: The user really cannot do much here as the change will just happen. The exact change isn't 100% decided, but will definitely be different.

jorisvandenbossche: Then it is maybe a bit too soon to raise a warning? As this will just be confusing, I might think

snowman2: Well, I wanted to have something. Not sure if there are going to be any more 2.x releases after this next one.

Issue #612, opened 2020-05-02
ENH: Add support for `radians` for Proj & Transformer.from_pipeline


snowman2:
With this addition: https://github.com/OSGeo/PROJ/pull/2144

We are able to support input as radians. Currently, only degrees are supported.

snowman2:
Related to: https://github.com/pyproj4/pyproj/issues/611#issuecomment-622949375

Since `Proj` used to accept `kwargs` and `radians` was an input argument in `pyproj` v1, I am thinking that it would be worthwhile in 2.6.1 to add the `radians` argument and raise a warning if it is set to True that it is currently ignored, but will work with `pyproj` 3+.

PR #613, opened 2020-05-02
Add warning for radians flag with Proj class

snowman2:
- [x] Start of #612
 - [x] Tests added

### DIFF
--- a/pyproj/_transformer.pyx
+++ b/pyproj/_transformer.pyx
@@ -424,8 +424,8 @@ cdef class _Transformer(Base):
             return
         if radians and self.is_pipeline:
             warnings.warn(
-                "radian input with pipelines is not supported and may result "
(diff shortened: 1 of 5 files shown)


snowman2:
@djhoese, just a ping to let you know about this change - it means you shouldn't need to worry about the previous ping in the release issue.

djhoese:
Thanks! As you know, pyresample plans on transitioning to using new pyproj features "soon". The only reason this custom class exists that was passing the old radians kwarg was to get degrees when given degrees for a latlong projection. Thanks for making this a warning instead for now.

Just curious, the support for radians in pyproj 3, does that change the degrees -> degrees default for lat/lon projections?

snowman2:
The default of degrees will remain in pyproj 3. The only change will be that you can use radians if you want.

PR #615, opened 2020-05-04
TST: Simplify logic if shapely is available, remove old comments

mwtoews:
Since the start of 2020, Shapely 1.7.0 has binary wheels for Windows and Python 3.8, so some of the comments and logic used for testing is outdated. Furthermore, the test logic should only skip certain test if shapely is not installed in the test environment.


snowman2:
@mwtoews thanks for updating the logic here :+1:

PR #620, opened 2020-05-05
Fix typo

habi:
I stumbled over https://pyproj4.github.io/pyproj/stable/installation.html and found this typo :)


snowman2:
Thanks!

habi:
Very welcome.

PR #625, opened 2020-05-07
ENH: Support objects with __array__ method

### LINKED ISSUE #573: ENH: Support obects with __array__ method
https://docs.scipy.org/doc/numpy/reference/arrays.interface.html

Had the thought, putting here to look into.

snowman2:
- [x] Closes #573
 - [x] Tests added

This includes at least:
- pandas.Series
- xarray.DataArray
- dask.array.Array

### DIFF
--- a/pyproj/utils.py
+++ b/pyproj/utils.py
@@ -32,10 +32,7 @@ def _copytobuffer(xx: Any) -> Tuple[Any, bool, bool, bool]:
     istuple = False
     # check for pandas.Series, xarray.DataArray or dask.array.Array
     if hasattr(xx, "__array__") and callable(xx.__array__):
-        try:
-            xx = xx.__array__(dtype="d")
-        except TypeError:
-            pass
+        xx = xx.__array__()
 
     # if it's a numpy array scalar convert to float
     # (array scalars don't support buffer API)
--- a/pyproj/utils.py
+++ b/pyproj/utils.py
@@ -30,7 +30,14 @@ def _copytobuffer(xx: Any) -> Tuple[Any, bool, bool, bool]:
     isfloat = False
     islist = False
     istuple = False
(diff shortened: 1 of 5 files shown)


pyproj/utils.py
```diff
@@ -30,7 +30,14 @@ def _copytobuffer(xx: Any) -> Tuple[Any, bool, bool, bool]:
     isfloat = False
     islist = False
     istuple = False
-    # first, if it's a numpy array scalar convert to float
+    # check for pandas.Series, xarray.DataArray or dask.array.Array
+    if hasattr(xx, "__array__") and callable(xx.__array__):
+        try:
+            xx = xx.__array__(dtype="d")
```
djhoese: What does `dtype="d"` do here? Give you a byte array?

snowman2: It converts to a double-like:
```python
>>> import xarray
>>> da = xarray.DataArray([1])
>>> da.dtype
dtype('int64')
>>> da.__array__(dtype="d")
array([1.])
>>> da.__array__(dtype="d").dtype
dtype('float64')
```

snowman2: I just removed it as the `dtype` kwarg isn't always there for all methods. Actually, I guess it is just an arg for numpy:
```python
>>> na = numpy.array([1])
>>> na.dtype
dtype('int64')
>>> na.__array__("d")
array([1.])
>>> na.__array__("d").dtype
dtype('float64')
>>> na.__array__(dtype="d").dtype
TypeError: __array__() takes no keyword arguments
```

```
__array__(...) method of numpy.ndarray instance
    a.__array__(|dtype) -> reference if type unchanged, copy otherwise.
```

snowman2: Either way, it does not need it as `astype` is called later.

PR #628, opened 2020-05-09
DOC: Add transformation grid documentation page

snowman2:

### DIFF
--- a/docs/transformation_grids.rst
+++ b/docs/transformation_grids.rst
@@ -1,0 +1,92 @@
+.. _transformation_grids:
+
+Transformation Grids
+=====================
+
+Transformation grids are necessary when you are performing datum transformations.
+
+More information about the data available is located under the PROJ
+`resource files <https://proj.org/resource_files.html#transformation-grids>`__
+documentation.
+
+`pyproj` API for managing the :ref:`data_directory`
+
+.. warning:: pyproj 2 includes datumgrid 1.8 in the wheels. pyproj 3 will not include any datum grids.
+
+
+Downloading data
+----------------
+
+PROJ 7+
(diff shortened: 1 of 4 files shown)


docs/transformation_grids.rst
```diff
@@ -0,0 +1,92 @@
+.. _transformation_grids:
+
+Transformation Grids
+=====================
+
+Transformation grids are necessary when you are performing datum transformations.
```
jorisvandenbossche, line 6: They are not "necessary" (transformations also work if the grids are not available?), but make the transformation more accurate?

docs/transformation_grids.rst
```diff
@@ -0,0 +1,92 @@
...
+
+What grids to download?
+-----------------------
+
+- Only using the :obj:`pyproj.crs.CRS` or :obj:`pyproj.Geod` classes? Then no grids are needed.
+
+- Have a machine that can hold and extra 500 MB - 1 GB of data? Then downloading all grids shouldn't be an issue.
+
+- Have a machine with limited space, a great network connection, and PROJ 7+? Look into `PROJ network <https://proj.org/usage/network.html>`__ capabilities.
+
```
jorisvandenbossche, line 72: With PROJ 7, it will also download the subset of grids you need (and cache those) automatically?

snowman2: By default it isn't enabled. You need to enable it. However, if you installed it via conda-forge it is enabled by default.

docs/transformation_grids.rst
snowman2: That is true. I guess it depends on your definition of necessary.

Issue #572, opened 2020-03-11
ENH: Replicate `projsync` for pyproj


snowman2:
Related to: https://github.com/OSGeo/PROJ/issues/2028

Two parts:
1. Python API
2. CLI

Currently, without `pj_context_set_user_writable_directory`, this means that the download can only go to the user writable directory via the `PROJ` API.

The alternative would be to create a python based downloader, but the logic may differ from PROJ.
 - To get the user writable directory, could use https://github.com/ActiveState/appdirs
 - To download the file, have many options available and can manually download to user directory, the data directory, or whichever is specified by the user.

PR #636, opened 2020-05-14
ENH: Added pyproj.datadir.get_user_data_dir

snowman2:
- [x] Will be helpful for #572
 - [x] Tests added


pyproj/_datadir.pyx
```diff
@@ -51,3 +51,19 @@ cdef void pyproj_context_initialize(
 def pyproj_global_context_initialize():
     proj_log_func(NULL, NULL, pyproj_log_function)
     set_context_data_dir(NULL)
+
+
+def get_user_data_dir(bint create: bool = True) -> str:
+    """
+    Parameters
```
jorisvandenbossche, line 58: Can you indicate here in the docstring how this is different from `get_data_dir` ?

snowman2: Good thinking. Added some information.

PR #648, opened 2020-05-28
ENH: Add transformation grid sync API/CLI

### LINKED ISSUE #572: ENH: Replicate `projsync` for pyproj

snowman2:
- [x] Closes #572
 - [x] Tests added

Why not use click for the CLI?
- Currently no external dependencies and it is nice that way.
- The CLI isn't all that complex, so not really needed.

Why not use PROJ download methods?
- Only downloads to the user data directory.
- I like adding the `.part` usage when downloading & checking the file.
- The behavior differed from the `grid.available` logic in `Transformer.download_grids` due to it's usage of the `cache.db` for checking if the download was needed.
- Possibly remove need for requiring PROJ to be compiled with `curl` if grids pre-downloaded?


snowman2:
Failures related to this update: https://github.com/OSGeo/PROJ/issues/2237. To be addressed in another PR.

PR #649, opened 2020-05-29
TST: Update tests with scope & remarks

snowman2:
- [x] Related to https://github.com/OSGeo/PROJ/issues/2247


snowman2:
Failure is due to core dump on unrelated test ...

PR #663, opened 2020-06-20
CLN: PJ_WKT2_2018->PJ_WKT2_2019; remove ContextManager; ctypedef

snowman2:
- PJ_WKT2_2018 is deprecated: https://github.com/OSGeo/PROJ/blob/af8334a5c8f6ec7d98e7f7ce0564ed4684896426/src/proj.h#L772-L773

- ContextManager is from pyproj 2.3 single context implementation and no longer needed

- Changed external declarations to use `ctypedef` instead of `cdef`


snowman2:
Tests pass - coveralls server is having issues.

PR #673, opened 2020-06-30
ENH: Added pyproj.set_use_global_context

### LINKED ISSUE #661: Instantiating many proj objects much slower than 1.9.6
I have a tool in pyproj 1.9.6 that I use for doing a "reverse lookup" of projections. Given an x/y in unknown CRS and a known longitude/latitude, this finds the projections which place that x/y closest to the known longitude/latitude. This is helpful when trying to track down an unknown coordinate system. 

This requires instantiating thousand of `proj`/`Proj` objects, but it only takes a few seconds in pyproj 1.9.6. Recently I wanted to upgrade to more recent versions of PROJ and GDAL, but this tool is now taking a few minutes, about 50 times longer in pyproj 2.x:
- [This gist](https://gist.github.com/dmahr1/a68dfa44f06d19a1aa2f1f81f64d5fb7) takes 2.32 seconds to run with pyproj 1.9.6.
- [This nearly identical gist](https://gist.github.com/dmahr1/01b561b166068b755e034b6e9e18e01f) takes 168 seconds to run with pyproj 2.6.1

I know that a lot changed in the underlying PROJ C++ library between pyproj 1.9.6 and 2.x. **But is there any way to restore the fast instantiation of the `proj`/`Proj` objects?** The projections don't have to be exact - just close enough to this reverse lookup tool. Also, I am willing to serialize/pickle the `proj` objects if that would help, though my understanding was that that didn't work with Python C extensions.

snowman2:
- [x] Closes #661
 - [x] Tests added


snowman2:
Hmmm, initially had speedup that was noticeable. Not seeing it anymore. Will need to investigate.

PR #698, opened 2020-08-27
BLK: Update to latest black & move target version to pyproject.toml

snowman2:
Looks like stickler is stuck on `19.10b0`

### DIFF
--- a/pyproj/geod.py
+++ b/pyproj/geod.py
@@ -570,10 +570,11 @@ class Geod(_Geod):
         -------
         (float, float):
             The geodesic area (meters^2) and permimeter (meters) of the polygon.
-       """
+        """
         try:
             return self.polygon_area_perimeter(  # type: ignore
-                *geometry.xy, radians=radians,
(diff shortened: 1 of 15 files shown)


pyproj/geod.py
```diff
@@ -570,10 +570,11 @@ def geometry_area_perimeter(
         -------
         (float, float):
             The geodesic area (meters^2) and permimeter (meters) of the polygon.
-       """
+        """
         try:
             return self.polygon_area_perimeter(  # type: ignore
-                *geometry.xy, radians=radians,
+                *geometry.xy,
+                radians=radians,
```
jorisvandenbossche, line 577: BTW, the reason those are now put on multiple lines (while they fit on a single) is the trailing comma AFAIU. If you want to keep it on a single line, you can remove the comma (not that it is a big issue though ;))

snowman2: Nice - that is a good tip. Thanks :+1:

PR #717, opened 2020-10-09
TST: Test EPSG 10 PROJ changes (PART 2)

snowman2:
Related: https://github.com/OSGeo/PROJ/pull/2371


snowman2:
The main change that might need to be looked at is: `test_datum__from_user_input`

PR #742, opened 2020-11-04
Specify units of measuring methods

underchemist:
Specifies that the unit of the output for `Geod.line_length` and `Geod.line_lengths` is meters.

### DIFF
--- a/pyproj/geod.py
+++ b/pyproj/geod.py
@@ -352,7 +352,7 @@ class Geod(_Geod):
         """
         .. versionadded:: 2.3.0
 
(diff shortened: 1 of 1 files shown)


snowman2:
Thanks :+1:

snowman2:
@underchemist, it looks like the commit is not recognizing your username for some reason.

underchemist:
I did this quickly from my work laptop which doesn't have my personal git config setup, all good.

snowman2:
Sounds good. To double check, you don't mind merging it in as?

underchemist:
yep no problem

PR #752, opened 2020-11-26
Delete .gitattributes

micahcochran:
This file is a a relic from when PROJ's code (vendor code) was in the pyproj repo.  Github's linguist had to be told to ignore that code when calculating the amount of Python versus C code for it percent Language in the repo (PROJ dwarfs pyproj in lines of code).


snowman2:
Thanks 👍 🦃

PR #783, opened 2021-03-01
TST: mask warning in test_cf_from_numpy_dtypes

snowman2:
Continuation of #775

### DIFF
--- a/test/crs/test_crs_cf.py
+++ b/test/crs/test_crs_cf.py
@@ -51,19 +51,20 @@ def test_cf_from_numpy_dtypes():
         "latitude_of_projection_origin": numpy.int(45),
     }
     crs = CRS.from_cf(cf)
-    assert crs.to_dict() == {
-        "datum": "WGS84",
-        "lat_0": 45,
-        "lat_1": 60,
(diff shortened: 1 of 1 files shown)


test/crs/test_crs_cf.py
```diff
@@ -51,19 +51,20 @@ def test_cf_from_numpy_dtypes():
         "latitude_of_projection_origin": numpy.int(45),
```
jorisvandenbossche, line 51: Instead of masking the warning, I think the "proper" fix is to change the above to `numpy.int64(45)`

snowman2: The warning masked is the one about losing CRS information raised by pyproj when exporting to PROJ string.  That is good to know about the numpy deprecation. If you have a moment to raise an issue or open a quick PR with the fix, that would be helpful. Thanks 👍

jorisvandenbossche: Ah, yes, stupid me, of course the warning you were fixing wasn't coming from here, since the pytest.warns is only below ;)

(but indeed, something to fix anyway, and I see you already did a PR!)

Issue #782, opened 2021-02-26
REF: Making pyproj completely threadsafe


snowman2:
pyproj CRS/Proj/Transformer objects aren't meant to be shared across threads, but they can exist if created within the thread ([ref](https://pyproj4.github.io/pyproj/stable/advanced_examples.html#multithreading)). Sharing the same object across threads can cause issues (#589, https://github.com/corteva/geocube/issues/41)

The only way to entirely solve this is to be able to have a single PROJ context per thread and to be able to dynamically switch to that context depending on the thread they are in.

Current concerns to overcome:
1. How to be able to have a thread local context? The `threading.local` object is a possibility. However it may require transitioning from cython to python to cython types. This has caused stability issues in the past (#412), so need to be careful.
2. How to dynamically swap contexts based on the thread? The `PJ *` objects have a context assigned at creation. Would need to use `proj_assign_context` each time you use the context object.

snowman2:
This kills making the objects thread safe: https://github.com/OSGeo/PROJ/issues/1047

Seems the `PJ*` is not thread safe. So, you would have to construct a `PJ*` per thread per object. It would simpler and more stable to just tell users of `pyproj` to use `threading,local` to construct an object per thread.

Since this is the case, `pyproj` really cannot do anything at this time.

snowman2:
Just had another thought, The `CRS` and `Transformer` classes have both a Python and a Cython version.

Following the logic in https://github.com/geopandas/geopandas/pull/1846,

Instead of having the Python `CRS` class inherit from the Cython `_CRS` class, it could instead store the Cython class on a `threading.local` in the class.

```python
class CRSLocal(threading.local):
    def __init__(self):
        self._crs = None  # Initialises in each thread

class CRS:
    self._local = CRSLocal()
...
```

 Then it could use the `srs` attribute to create a `_CRS` in each thread.

```python
@property
def _local_crs(self):
    if self._local._crs is None:
        self._local._crs = _CRS(self.srs)
    return self._local._crs
```
A similar concept could be applied to the `Transformer` since it already stores the Cython class as a property.

jorisvandenbossche:
> Instead of having the Python `CRS` class inherit from the Cython `_CRS` class, it could instead store the Cython class on a `threading.local` in the class.

Indeed, that's what I tried to suggest in https://github.com/geopandas/geopandas/issues/1842#issuecomment-786477955 ;) 

[...]

PR #793, opened 2021-03-06
REF: Multithread safe CRS

snowman2:
- [x] Part of #782
 - [x] Tests added

### DIFF
--- a/docs/history.rst
+++ b/docs/history.rst
@@ -4,6 +4,7 @@ Change Log
 3.1.0
 -----
 * DEP: Minimum supported Python version 3.7 (issue #790)
+* REF: Multithread safe CRS (issue #782)
 
 3.0.1
 -----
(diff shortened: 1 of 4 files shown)


jorisvandenbossche:
Cool! Will try to give this a go tomorrow

jorisvandenbossche:
This looks good!

As mentioned in https://github.com/geopandas/dask-geopandas/pull/36, locally this fixed the decoding errors we see with dask-geopandas. 

Maybe one suggestion: if you want to limit the duplication of the docstrings (because now all the attributes/methods that needed to be added to dispatch to `self._crs` have a copy of the docstring), I think it should be relatively easy to use a decorator that automatically adds the docstring of the cython class. 

Something like

```
def inherit_doc():
    """
    A decorator adding the docstring from the equivalent `_CRS` method.
    """

    def decorator(decorated):
        original_method = getattr(_CRS, decorated.__name__, None)
        if original_method:
            doc = original_method.__doc__ or ""
        else:
            doc = ""

        decorated.__doc__ = doc
        return decorated

    return decorator
```

snowman2:
Thanks for testing! I am going to merge now and deal with the docstring later. It is something that I agree should be looked into.

Issue #754, opened 2020-12-02
ENH: Add filter kwargs to Transformer.from_crs


snowman2:
- https://github.com/OSGeo/PROJ/pull/2477
- https://github.com/OSGeo/PROJ/pull/2488

AUTHORITY, ACCURACY and ALLOW_BALLPARK

Requires PROJ 8+

snowman2:
This requires using `proj_create_crs_to_crs_from_pj`. This caused issues in the past https://github.com/pyproj4/pyproj/pull/412/commits/7a063920edb1edb5c22b06cdb17bebfdfa733802 and would have to find a stable solution for this to work moving forward.

snowman2:
Good reference: https://proj.org/development/reference/functions.html#c.proj_create_crs_to_crs_from_pj

PR #412, opened 2019-08-27
PROJ 6.2+ required; remove global context; autoclose database

snowman2:
- [x] Closes #411
 - [x] Closes #408
 - [x] Closes #403
 - [x] Closes #426
 - [x] Fully documented, including `history.rst` for all changes and `api/*.rst` for new API

PR #800, opened 2021-03-14
ENH: Use proj_create_crs_to_crs_from_pj for creating transformer PJ*

snowman2:
Needed for #754

Attempted this in the past and had to revert ([ref](https://github.com/pyproj4/pyproj/commit/7a063920edb1edb5c22b06cdb17bebfdfa733802)) due to instability in #412.

@cgohlke would you mind testing this with Windows to see if it works or if we should wait on this change?


snowman2:
Have an alternative in #803 that should be stable as it is essentially what is in PROJ now. However, if this PR worked, it would be preferred because it reduces the need to re-create additional `PJ*` objects. However, due to the instability introduced in the past, I am hesitant to proceed with this.

PR #802, opened 2021-03-18
REF: Multithread safe Transformer

snowman2:
- [x] Related to #782
 - [x] Tests added


pyproj/transformer.py
```diff
@@ -180,14 +242,45 @@ class Transformer:
...
             )
-        self._transformer = base_transformer
+
+        self._local = TransformerLocal()
+        if isinstance(transform_init, _Transformer):
+            # this is from TransformerGroup
+            # and should only be used from there
+            self._transform_init: Union[
+                TransformerFromCRS, TransformerFromPipeline
+            ] = TransformerFromPipeline(transform_init.definition)
```
snowman2, line 266: After thinking on it, I am not sure I like this solution. I am thinking making an `UnsafeTransformer` class for the `TransformerGroup` where it always returns the original transformer + a warning in the docstrings about not being thread-safe would be a better approach.

Issue #809, opened 2021-03-29
would it be possible to have the CRS bounds in its original unit ? 


12rambau:
Let's assume I want to get the bounds of the projection 'EPSG:3857' in meters (unit of the 3857 projection)

In the current implementation of pyproj and with my little knowledge I would do the following : 
```python
from pyproj import CRS, Transformer

in_crs = CRS.from_epsg(4326)
out_crs = CRS.from_epsg(3857)
x_min, y_min, x_max, y_max = out_crs.area_of_use.bounds

proj = Transformer.from_crs(4326, 3857, always_xy=True)
bl = proj.transform(x_min, y_min)
tr = proj.transform(x_max, y_max)
```

that's a lot of steps just to extract the bounds in the original proj unit. 
Would it be possible to add a method to retrieve the projection in its original unit ? 

I would be happy to PR this very code in PR if you agree with this way of computing it. 

> maybe I'm foolish and there is already a way to retrieve the bounds...

snowman2:
To transform the boundary properly, would likely need a method similar to:
https://rasterio.readthedocs.io/en/latest/api/rasterio.warp.html#rasterio.warp.transform_bounds

snowman2:
PROJ does something similar:
https://github.com/OSGeo/PROJ/blob/5fb47f125c25751bd1b1018849de9e7ee69107f4/src/4D_api.cpp#L974-L1043

snowman2:
How to get projected bounds with #810:
```python
from pyproj import CRS, Transformer
crs = CRS("EPSG:3857")
transformer = Transformer.from_crs(crs.geodetic_crs, crs, always_xy=True)
transformer.transform_bounds(*crs.area_of_use.bounds)
```

12rambau:
awesome ! thanks a lot

snowman2:
Release is coming soon. Hopefully by the end of the week 🤞

PR #810, opened 2021-03-31
ENH: Add Transformer.transform_bounds

snowman2:
- [x] Related to #809
 - [x] Tests added


snowman2:
Seems like Windows doesn't like the current implementation. Works on linux 🤷‍♂️

snowman2:
There are some edge cases mentioned here: https://github.com/OSGeo/PROJ/pull/2616. But, I think this is a good enough start.

PR #841, opened 2021-05-05
add `[inv|fwd]_intermediate()` - to calculate intermediate points (like improved `Geod.npts()`)

idanmiara:
pyproj/geod.py - add a parameter `return_points` to Geod.npts() to determinate the return type; improve return type hint
pyproj/geod.py - add a parameters `initial_idx`, `terminus_idx` to Geod.npts()
pyproj/[geod.py, _geod.pyi, _geod.pyx] - add `inv_intermediate_by_npts` functions that uses given buffers `out_lons`, `out_lats` (and `out_azis`) buffers instead of creating new buffers
pyproj/_geod.pxd - add interface for `geod_lineinit`
pyproj/[geod.py, _geod.pyi, _geod.pyx] - add `inv_intermediate_by_del`, `fwd_intermediate_by_npts`, `fwd_intermediate_by_del` functions, similar to `inv_npts` but using forward calculation
test/test_geod.py - add tests for new functions

## Related PR

https://github.com/pyproj4/pyproj/pull/825/

### DIFF
--- a/pyproj/_geod.pyx
+++ b/pyproj/_geod.pyx
@@ -199,42 +199,6 @@ cdef class Geod:
 
     @cython.boundscheck(False)
     @cython.wraparound(False)
-    def _npts(
-        self,
-        double lon1,
-        double lat1,
-        double lon2,
-        double lat2,
-        int npts,
-        bint radians=False,
-        int initial_idx=1,
-        int terminus_idx=1,
-    ):
-        """
-        given initial and terminus lat/lon, find npts intermediate points.
-        returns lons, lats buffers
-        """
-        res = \
-            self._inv_or_fwd_intermediate(
-                lon1=lon1,
-                lat1=lat1,
-                lon2_or_azi1=lon2,
-                lat2_or_nan=lat2,
-                npts=npts,
-                del_s=0,
-                radians=radians,
-                initial_idx=initial_idx,
-                terminus_idx=terminus_idx,
-                flags = GEOD_INTER_FLAG_AZIS_DISCARD,
-                out_lons=None,
-                out_lats=None,
-                out_azis=None,
-            )
-
-        return res.lons, res.lats
-
-    @cython.boundscheck(False)
-    @cython.wraparound(False)
     def _inv_or_fwd_intermediate(
         self,
         double lon1,
--- a/pyproj/_geod.pyx
+++ b/pyproj/_geod.pyx
@@ -161,9 +161,12 @@ cdef class Geod:
         double lat2,
         int npts,
         bint radians=False,
+        int initial_idx=1,
+        int terminus_idx=1,
     ):
         """
         given initial and terminus lat/lon, find npts intermediate points.
+        returns lons, lats buffers
         """
         cdef Py_ssize_t iii
         cdef double del_s
@@ -172,11 +175,13 @@ cdef class Geod:
         cdef double plon2
         cdef double plat2
         cdef geod_geodesicline line
+
         cdef array.array array_template = array.array("d", [])
-        cdef array.array lats = array.clone(array_template, npts, zero=False)
-        cdef array.array lons = array.clone(array_template, npts, zero=False)
-        cdef PyBuffWriteManager lats_buff = PyBuffWriteManager(lats)
-        cdef PyBuffWriteManager lons_buff = PyBuffWriteManager(lons)
+        cdef array.array out_lons = array.clone(array_template, npts, zero=False)
+        cdef array.array out_lats = array.clone(array_template, npts, zero=False)
+
+        cdef PyBuffWriteManager lats_buff = PyBuffWriteManager(out_lats)
+        cdef PyBuffWriteManager lons_buff = PyBuffWriteManager(out_lons)
 
         with nogil:
             if radians:
@@ -189,18 +194,18 @@ cdef class Geod:
                 &line, &self._geod_geodesic, lat1, lon1, lat2, lon2, 0u,
             )
             # distance increment.
-            del_s = line.s13 / (npts + 1)
+            del_s = line.s13 / (npts + initial_idx + terminus_idx - 1)
             # loop over intermediate points, compute lat/lons.
-            for iii in range(1, npts + 1):
-                s12 = iii * del_s
+            for iii in range(0, npts):
+                s12 = (iii + initial_idx) * del_s
                 geod_position(&line, s12, &plat2, &plon2, &pazi2)
                 if radians:
                     plat2 *= _DG2RAD
                     plon2 *= _DG2RAD
-                lats_buff.data[iii - 1] = plat2
-                lons_buff.data[iii - 1] = plon2
+                lats_buff.data[iii] = plat2
+                lons_buff.data[iii] = plon2
 
-        return lons, lats
+        return out_lons, out_lats
 
     @cython.boundscheck(False)
     @cython.wraparound(False)
--- a/pyproj/_geod.pyx
+++ b/pyproj/_geod.pyx
@@ -2,15 +2,61 @@ include "base.pxi"
 
 cimport cython
 from cpython cimport array
+from libc.math cimport ceil, isnan, round
 
 import array
+from collections import namedtuple
 
 from pyproj.compat import cstrencode, pystrdecode
+from pyproj.enums import GeodIntermediateFlag
 from pyproj.exceptions import GeodError
 
 geodesic_version_str = (
     f"{GEODESIC_VERSION_MAJOR}.{GEODESIC_VERSION_MINOR}.{GEODESIC_VERSION_PATCH}"
 )
+
+GeodIntermediateReturn = namedtuple(
+    "GeodIntermediateReturn", ["npts", "del_s", "dist", "lons", "lats", "azis"]
+)
+
+GeodIntermediateReturn.__doc__ = """
+.. versionadded:: 3.1.0
+
+Geod Intermediate Return value (Named Tuple)
+
+Parameters
+----------
+
+npts: int
+    number of points
+del_s: float
+    delimiter distance between two successive points
+dist: float
+    distance between the initial and terminus points
+out_lons: Any
+    array of the output lons
+out_lats: Any
+    array of the output lats
+out_azis: Any
+    array of the output azis
+"""
+
+
+cdef int GEOD_INTER_FLAG_DEFAULT = GeodIntermediateFlag.DEFAULT
+
+cdef int GEOD_INTER_FLAG_NPTS_MASK = GeodIntermediateFlag.NPTS_MASK
+cdef int GEOD_INTER_FLAG_NPTS_ROUND = GeodIntermediateFlag.NPTS_ROUND
+cdef int GEOD_INTER_FLAG_NPTS_CEIL = GeodIntermediateFlag.NPTS_CEIL
+cdef int GEOD_INTER_FLAG_NPTS_TRUNC = GeodIntermediateFlag.NPTS_TRUNC
+
+cdef int GEOD_INTER_FLAG_DEL_S_MASK = GeodIntermediateFlag.DEL_S_MASK
+cdef int GEOD_INTER_FLAG_DEL_S_RECALC = GeodIntermediateFlag.DEL_S_RECALC
+cdef int GEOD_INTER_FLAG_DEL_S_NO_RECALC = GeodIntermediateFlag.DEL_S_NO_RECALC
+
+cdef int GEOD_INTER_FLAG_AZIS_MASK = GeodIntermediateFlag.AZIS_MASK
+cdef int GEOD_INTER_FLAG_AZIS_DISCARD = GeodIntermediateFlag.AZIS_DISCARD
+cdef int GEOD_INTER_FLAG_AZIS_KEEP = GeodIntermediateFlag.AZIS_KEEP
+
 
 cdef class Geod:
     def __init__(self, double a, double f, bint sphere, double b, double es):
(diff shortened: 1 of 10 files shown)


idanmiara:
Maybe I'll change this flag into an Enum so other return types could be supported in the future, like (lons, lats, ranges, azimuths)

snowman2:
@idanmiara, I am planning on doing a release soon since PROJ 8.0.1 was just released. Do you want to try to include this in this release or do you need more time?

idanmiara:
> @idanmiara, I am planning on doing a release soon since PROJ 8.0.1 was just released. Do you want to try to include this in this release or do you need more time?

Hi, thanks for your letting me know!
I'm trying to complete it now.

idanmiara:
I had a few other related ideas, I've implemented them in `geod.py` but I see that I can probably implement them much more efficient in Cython by modifying `_geod.Geod._npts`:
1. add a flag `add_points: int` add an option to add initial and/or terminus points to the output vector (0= default- add neither, 1=add first; 2=add last; 3= add both)
2. add an optional `az: float` parameter to allow to calculate the line by using `az` instead of `lon2, `lat2` using `geod_lineinit` (like before https://github.com/pyproj4/pyproj/pull/825)
3. add optional `del_s: float` and `n_ceil: bool` that can be an alternative way to calculate the number of points number (https://github.com/pyproj4/pyproj/pull/825/files#diff-884caf2af6e10fc8e0b305d38b427edc7b43d9ed3ec739aed1be2de94049ab7dR192):
```
    npts = dist / del_s
    if n_ceil:
        npts = math.ceil(npts)
    npts = int(npts)

    dist = npts * del_s
```
4. optionally add a `pazi2` vector to the output (https://github.com/pyproj4/pyproj/pull/825/files#diff-884caf2af6e10fc8e0b305d38b427edc7b43d9ed3ec739aed1be2de94049ab7dR196) by adding another return type option

I've never used Cython before. from what I see it's not too complicated. Do you have some pointers on how to start besides the Cython official tutorials? How do you debug it?

I'm thinking if I can make it to the next release, when are you planning on making it?

snowman2:
> Do you have some pointers on how to start besides the Cython official tutorials? How do you debug it?

That sounds like a good place to start. Also, looking through the existing code could be helpful.

For debugging, I usually rebuild the cython after each change with:

`pip install -e .` or `python setup.py build_ext --inplace`

And then test it like normal python code.

snowman2:
> I'm thinking if I can make it to the next release, when are you planning on making it?

No rush. If you are regularly working on it, it can wait until you are done.

snowman2:
One thought I had was to potentially break out the different types of inputs/responses into different methods. This will add clarify for expected input/output to the user and simplify the logic for each method.

idanmiara:
Hi @snowman2 
I'm almost done.
Would you like to review the APIs before I'm finalizing the tests?

snowman2:
Thanks for working on this. I will need to take a more detailed look at it later, but here are the initial things I have:
- I think it makes sense for the method names to start with `npts_` instead of end with it. Thoughts?
- Instead of modifying existing tests, Mind adding new tests? The more you can break them out into individual test cases, the better. It helps narrow down what is failing and why. Here are some examples [ref](https://github.com/pyproj4/pyproj/blob/master/test/crs/test_crs.py).

idanmiara:
> Thanks for working on this. I will need to take a more detailed look at it later, but here are the initial things I have:
> 
> * I think it makes sense for the method names to start with `npts_` instead of end with it. Thoughts?
I'm guessing `npts` is a short for `number of points`? I thought that a better to describe what the function does is `intermediate`.
I named the new functions with this in mind. 
`by_npts` means by number of points
`by_del` means by `del_s` (delimited?)

idanmiara:
@snowman2 
I think it's ready!
I've tried to document and test everything. 
I didn't make separate test functions as I test the functions against each other, it would have made things more complicated for me to test them individually.

pyproj/geod.py
```diff
@@ -336,17 +340,265 @@ def npts(
...
+
+        Similar to inv_intermediate_by_npts(),
+        but using npts as input as appose to del_s
+
+        Similar to npts(), but using the given buffers for the output.
+
+
+        Parameters
+        ----------
+        out_lons: array, :class:`numpy.ndarray`
```
snowman2, line 411: `pyproj` isn't strictly tied to `numpy`. It does support `numpy`, but it also supports quite a variety of other types as well. Initializing a python array within the function and returning it allows for this to be useful to a wider range of users and simplifies usage. Users can always convert the output to a numpy array if desired.

pyproj/geod.py
```diff
@@ -336,17 +340,265 @@ def npts(
...
+        out_azis: Any,
+        lon1: float,
+        lat1: float,
+        lon2: float,
+        lat2: float,
+        npts: int,
+        initial_idx: int = 1,
+        terminus_idx: int = 1,
+        radians: bool = False,
+    ) -> GeodIntermediateReturn:
```
snowman2, line 392: I am thinking that a namedtuple would be useful in place of the type `GeodIntermediateReturn`. This way you can store the `npts`, `del_s` `out_lats`, `out_lons`, and `out_azis` with named accessors for clarity. Here is an example [ref](https://github.com/pyproj4/pyproj/blob/2d12d5aa719ea989c8f6abdb81916d43200fe9c3/pyproj/database.pyx#L116-L261).

snowman2:
> I didn't make separate test functions as I test the functions against each other, it would have made things more complicated for me to test them individually.

I would definitely appreciate it if you did separate things out and add new tests as it makes maintenance much easier when things break. But, it won't block this from being merged.

snowman2:
> I named the new functions with this in mind.
> by_npts means by number of points
> by_del means by del_s (delimited?)

Looking at it now it makes sense. I have two thoughts:
1. Change user facing instance of `del_s` to `dist` for consistency with `inv/fwd` functions.
2. Do you think it makes sense to make one of `inv_intermediate_by_del` or `inv_intermediate_by_npts` just `inv_intermediate`?

idanmiara:
> > I didn't make separate test functions as I test the functions against each other, it would have made things more complicated for me to test them individually.
> 
> I would definitely appreciate it if you did separate things out and add new tests as it makes maintenance much easier when things break. But, it won't block this from being merged.

I would do my best to do that shortly after 3.1.0 is released.

pyproj/geod.py
idanmiara: The reason behind this choice was to make it as efficient as possible so the user can reuse this buffer for successive calls.
I will try to use a solution that would address both use cases.

idanmiara:
@snowman2 Any more comments?

pyproj/types.py
```diff
@@ -0,0 +1,18 @@
+from collections import namedtuple
+
+GeodIntermediateReturn = namedtuple(
```
snowman2, line 3: Mind moving this to _geod.pyx since it is no longer a type?

snowman2:
Thanks for keeping up with the changes. One thing that would be good to add is `.. version added:: 3.1` in the docstrings.

pyproj/types.py
idanmiara: > Mind moving this to _geod.pyx since it is no longer a type?

It is used in `geod.py` as well.

pyproj/types.py
```diff
@@ -0,0 +1,18 @@
...
+    "GeodIntermediateReturn", ["npts", "del_s", "lons", "lats", "azis"]
+)
+
+GEOD_INTER_FLAG_NPTS_MASK = 0xF
```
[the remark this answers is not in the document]
idanmiara: I guess I could use https://docs.python.org/3/library/enum.html#enum.IntFlag...

pyproj/types.py
snowman2: > It is used in geod.py as well.

Should be able to import it from `_geod` without too much trouble.

pyproj/types.py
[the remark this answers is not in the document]
snowman2: Looks like a good option.

snowman2:
@jswhit, if you have time to do a quick review of the geodesic functionality/internals in the cython code, I would appreciate it. If you are busy, definitely understand.

snowman2:
@idanmiara, to add yourself as a contributor, the [all-cotnributors add](https://allcontributors.org/docs/en/cli/usage#all-contributors-add) CLI is pretty helpful. There may need to be some tweaks after you add yourself to the formatting of the README and to update the count on the contributors badge in the README,

pyproj/types.py
idanmiara: OK, done.

pyproj/types.py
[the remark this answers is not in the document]
idanmiara: done

idanmiara:
> @idanmiara, to add yourself as a contributor, the [all-cotnributors add](https://allcontributors.org/docs/en/cli/usage#all-contributors-add) CLI is pretty helpful. There may need to be some tweaks after you add yourself to the formatting of the README and to update the count on the contributors badge in the README,

done

pyproj/geod.py
```diff
@@ -273,7 +283,10 @@ def npts(
...
         npts: int,
         radians: bool = False,
-    ) -> List:
+        initial_idx: int = 1,
+        terminus_idx: int = 1,
+        return_points: bool = True,
+    ) -> Union[List[Tuple[float, float]], Tuple[Sequence[float], Sequence[float]]]:
         """
         Given a single initial point and terminus point, returns
         a list of longitude/latitude pairs describing npts equally
```
snowman2, line 292: Mind adding:
```
.. versionadded:: 3.1 initial_idx, terminus_idx, return_points
```

pyproj/_geod.pyx
```diff
@@ -12,6 +15,43 @@ geodesic_version_str = (
...
+cdef int GEOD_INTER_FLAG_DEL_S_MASK = 0xF0
+cdef int GEOD_INTER_FLAG_DEL_S_RECALC = 0x00
+cdef int GEOD_INTER_FLAG_DEL_S_NO_RECALC = 0x10
+
+cdef int GEOD_INTER_FLAG_AZIS_MASK = 0xF00
+cdef int GEOD_INTER_FLAG_AZIS_DISCARD = 0x000
+cdef int GEOD_INTER_FLAG_AZIS_KEEP = 0x100
+
+
+class GeodIntemediateFlag(IntFlag):
```
snowman2, line 38: Mind moving this to `pyproj.enums` and adding a reference to it in `docs/api/enums.rst`?

idanmiara: done

pyproj/geod.py
idanmiara: done

idanmiara:
@snowman2 finally all the tests pass...

idanmiara:
I've squashed some commits to make the history cleaner.

pyproj/_geod.pyx
```diff
@@ -161,46 +184,151 @@ cdef class Geod:
...
+                terminus_idx=terminus_idx,
+                flags = GEOD_INTER_FLAG_AZIS_DISCARD,
+                out_lons=out_lons,
+                out_lats=out_lats,
+                out_azis=None,
+            )
+
+        return res.lons, res.lats
+
+    @cython.boundscheck(False)
```
snowman2, line 215: Since you allow users to pass in their own arrays, we can no longer disable the boundscheck [ref](https://cython.readthedocs.io/en/latest/src/userguide/numpy_tutorial.html#tuning-indexing-further).

> if you ‘’do’’ happen to access out of bounds you will in the best case crash your program and in the worst case corrupt data

I am wondering what the performance is for the two different scenarios:
1. Passing in numpy array with boundscheck enabled.
2. Returning `array.array` with boundscheck disabled and convert to numpy array later.

idanmiara: https://github.com/pyproj4/pyproj/pull/841/files#diff-884caf2af6e10fc8e0b305d38b427edc7b43d9ed3ec739aed1be2de94049ab7dR310

Isn't this check enough?

pyproj/_geod.pyx
```diff
@@ -161,46 +184,151 @@ cdef class Geod:
...
         cdef array.array array_template = array.array("d", [])
-        cdef array.array lats = array.clone(array_template, npts, zero=False)
-        cdef array.array lons = array.clone(array_template, npts, zero=False)
-        cdef PyBuffWriteManager lats_buff = PyBuffWriteManager(lats)
-        cdef PyBuffWriteManager lons_buff = PyBuffWriteManager(lons)
+        if (out_lons is None) \
+                or (out_lats is None) \
+                or (out_azis is None and store_az):
+            if npts == 0:
+                raise GeodError(
```
snowman2, line 255: If you define these earlier without initializing them:
```
        cdef PyBuffWriteManager lons_buff
        cdef PyBuffWriteManager lats_buff
        cdef PyBuffWriteManager azis_buff
```
Are you able to move this section after your npts check and this mode will work for all scenarios?

pyproj/_geod.pyx
snowman2: I missed that, thanks for pointing it out. That is good to have. I am still curious about the performance difference between the previous scenarios (minus the boundscheck part).

idanmiara: > I missed that, thanks for pointing it out. That is good to have. I am still curious about the performance difference between the previous scenarios (minus the boundscheck part).

That would be interesting to investigate...
My Cython training is just from this PR, I guess I could try to dig deeper later about it after I get more familiar with the language.

pyproj/_geod.pyx
idanmiara: good point.
I'm understanding that `array.clone` allocates a Python array, Could I call it inside a `with nogil` section ? 
If not, would there be a performance penalty to make another `with nogil` section after this call ?

snowman2: > Could I call it inside a with nogil section

You would have to call `with gil:` when you allocate the arrays.

> would there be a performance penalty to make another with nogil section after this call ?

There could be a minor penalty if you are doing multi-threading. Single threaded applications won't be impacted.

idanmiara: > > Could I call it inside a with nogil section
> 
> You would have to call `with gil:` when you allocate the arrays.
> 
> > would there be a performance penalty to make another with nogil section after this call ?
> 
> There could be a minor penalty if you are doing multi-threading. Single threaded applications won't be impacted.

OK, I've implemented your suggestion.
it's hard for me to evaluate if this case (`inv` + `del_s` + no input array) justifies the performance penalty for all the other cases, but I couldn't find a more elegant solution but to write two versions of this function.
I guess I could investigate that further in another PR.

pyproj/_geod.pyx
idanmiara: I've been thinking about your scenarios, I think the best way to put it into a test would be to make different versions of the `inv_or_fwd_intermediate` function, two as you suggested plus my previous version that doesn't use `with gil` for the case (`inv` + `del_s` + no input array).
I think this PR is already mature enough and this last version of `inv_or_fwd_intermediate` function is "one size fits all", with only a potential minor performance loss over my previous version.
I don't want these tests to push the release schedule of v3.1 further so I'm squashing my last commit and will open a new PR after the release of v3.1 to test these performance issues.

pyproj/_geod.pyx
idanmiara: see another comment in the conversation above.

pyproj/geod.py
```diff
@@ -273,13 +280,23 @@ def npts(
...
         """
+        .. versionadded:: 3.1.0 initial_idx, terminus_idx
+                                out_lons, out_lats, return_points,
+
         Given a single initial point and terminus point, returns
         a list of longitude/latitude pairs describing npts equally
         spaced intermediate points along the geodesic between the
         initial and terminus points.
 
+        Similar to inv_intermediate(), but with less options.
```
snowman2, line 298: With the addition of `inv_intermediate`, do we need the `return_points` kwarg?

pyproj/_geod.pyx
snowman2: I did some basic tests and it seems like the performance improves a bit when using numpy arrays. So, probably good to leave it as is.

pyproj/geod.py
```diff
@@ -336,17 +353,333 @@ def npts(
...
+            Latitudes(s) of the intermediate point(s)
+            If None then buffers would be allocated internnaly
+        out_azis: array, :class:`numpy.ndarray`, optional
+            az12(s) of the intermediate point(s)
+            If None then buffers would be allocated internnaly
+            unless requested otherwise by the flags
+
+        Returns
+        -------
+        GeodIntermediateReturn: NamedTuple
```
snowman2, line 527: Instead of adding the definition on each return, thoughts about adding `__doc__` to the `GeodIntermediateReturn` as it is done [here](https://github.com/pyproj4/pyproj/blob/2d12d5aa719ea989c8f6abdb81916d43200fe9c3/pyproj/database.pyx#L128-L149) and then adding a reference to `GeodIntermediateReturn` in the docs in `docs/api/geod.rst`?

snowman2:
I see a commit where you say:``".. versionadded:: 3.1.0" with ".. versionadded:: 3.1" for consistency` that modifies the `crs/transformer` files. Would you mind moving this to a separate PR for clarity? And, if you are feeling extra motivated, you could do a search for `versionadded` and make all of them consistent in that PR, but definitely not necessary.

pyproj/_geod.pyx
snowman2: Sounds good. Thanks :+1:

snowman2:
@idanmiara thanks for your patience with this PR! I think it is very close to being ready for merge.

pyproj/geod.py
idanmiara: I don't mind as I will indeed use `inv_intermediate` instead.

I also thought about removing the redundant `_geod._npts` (and implemente `geod.npts` on top of `_inv_or_fwd_intermediate`, as it starts with `_` I understand it's not part of the official API - should I remove it ?

pyproj/geod.py
idanmiara: OK. 
Should I move `GeodIntermediateReturn` to someplace else (new `pyproj.types.py`?)
Otherwise how could I declare it correctly in `_geod.pyi` ? I've tried `from pyproj.geod import GeodIntermediateReturn` before but it fails, and as it is now I get incorrect type hints from the IDE (as it sees it as `NamedTuple` as I decalred in `_geod.pyi`.

idanmiara:
> I see a commit where you say: `".. versionadded:: 3.1.0" with ".. versionadded:: 3.1"` for consistency that modifies the `crs/transformer` files. Would you mind moving this to a separate PR for clarity? And, if you are feeling extra motivated, you could do a search for `versionadded` and make all of them consistent in that PR, but definitely not necessary.

Sure.

idanmiara:
> @idanmiara thanks for your patience with this PR! I think it is very close to being ready for merge.

Thank you, I highly appreciate your thorough review!

pyproj/geod.py
snowman2: No need to move it to a new place. Here an example of declaring the tuple type: [ref](https://github.com/pyproj4/pyproj/blob/2d12d5aa719ea989c8f6abdb81916d43200fe9c3/pyproj/database.pyi#L25).

pyproj/geod.py
snowman2: Either way is fine. I will leave that decision up to you.

pyproj/geod.py
idanmiara: nice.

pyproj/_geod.pyx
```diff
@@ -161,46 +207,147 @@ cdef class Geod:
...
+        int terminus_idx=1,
+        object out_lons=None,
+        object out_lats=None,
     ):
         """
         given initial and terminus lat/lon, find npts intermediate points.
+        returns lons, lats buffers
+        """
+        res = \
+            self._inv_or_fwd_intermediate(
```
snowman2, line 220: I don't see `_npts` used anywhere. Mind removing it?

idanmiara: I thought to leave it for backwards compatibility. 
Is the policy not to maintain backwards compatibility with names that prefixed with `_` ?

pyproj/geod.py
```diff
@@ -273,13 +280,22 @@ def npts(
         lat2: float,
         npts: int,
         radians: bool = False,
+        initial_idx: int = 1,
+        terminus_idx: int = 1,
+        out_lons: Any = None,
```
snowman2, line 285: I think it would simplify things (reduce number of tests needed to ensure everything works properly) and be more clear to leave `out_lons`/`out_lats` as options on `inv_intermediate` and not on `npts`. Thoughts?

pyproj/_geod.pyx
snowman2: > Is the policy not to maintain backwards compatibility with names that prefixed with _ ?

If it is prefixed with `_`, it is private should only be used internally by `pyproj` (if someone does use it, it is at their own risk). So, backwards compatibility doesn't need to be maintained.

pyproj/geod.py
idanmiara: I don't have a strong opinion here.
I would imagine that whoever cares about performance would not use this function anyway because of `list(zip())` and would use `inv_intermediate` instead.
Would you like me to remove them?

snowman2: > Would you like me to remove them?

That would be great. Thanks :+1:

idanmiara: Done

snowman2:
@idanmiara this looks good. Ready to merge?

idanmiara:
> @idanmiara this looks good. Ready to merge?

Absolutely!

snowman2:
Thanks @idanmiara :+1:

PR #848, opened 2021-05-18
changing constructor-methods from staticmethods to classmethods

### LINKED ISSUE #847: CRS classmethods changed to staticmethods
At some point pyproj used classmethods for constructor-type methods in pyproj.crs.CRS. In [this commit](https://github.com/pyproj4/pyproj/commit/6296de2be98319be3540e852869024298d63acc4#diff-5c31eacd92cd28f19d968f4995e604f2f015a7adbc83866fcc34a65eeb530489) this was changed to use staticmethods. The [linked issue](https://github.com/pyproj4/pyproj/issues/389) does not seem to mention this change (unless I missed it). In any case, this change effectively breaks subclassing, as in this example:

```python
import pyproj


class myCRS(pyproj.crs.CRS):
    def print_foo(self):
        print("foo-bar!")


proj = pyproj.Proj("+proj=etmerc +ellps=GRS80 +lon_0=13 +lat_0=2 +x_0=0 +y_0=0 +k_0=1")
crs = myCRS.from_proj4(proj.definition_string())
print(type(crs))
print(isinstance(crs, myCRS))
crs.print_foo()
```

#### Expected Output

expect:

```
<class '__main__.myCRS'>
True
foo-bar!
```

what you get:

```
<class 'pyproj.crs.crs.CRS'>
False
...
AttributeError: 'CRS' object has no attribute 'print_foo'
```

#### Environment Information

Any ( we had been using pyproj v2.4.1, and just upgraded to 3.0.1)

#### Installation method

build from source

nmaxwell:
changes use of staticmethods to classmethods in constructor-methods in Transformer and CRS. I left from_cf in CRS, as there is some logic in there on whether to return a CRS, a GeographicCRS, a DerivedGeographicCRS, or a CompoundCRS. It may require some rethinking on how handle that properly with classmethods. 

 - [ x] Closes #847
 - [ x] Tests added

### DIFF
--- a/pyproj/crs/crs.py
+++ b/pyproj/crs/crs.py
@@ -1489,12 +1489,13 @@ class _CleanlyInheritableCRS(CRS):
 
 # remove from_* factory methods in _CleanlyInheritableCRS. They can't be deleted, as that would delete them from the
 # parent CRS. Best we can do is override them with a property which raises AttributeError.
-for item in CRS.__dict__:
+for item in CRS.__dict__.keys():
+    if item.startswith("from_"):
 
-    def _override(self):
-        raise AttributeError(f"method '{item}' is not accessible from this class")
+        def _override(self, item=item):
+            raise AttributeError(f"method '{item}' is not accessible from this class")
 
-    setattr(_CleanlyInheritableCRS, item, property(_override))
+        setattr(_CleanlyInheritableCRS, item, _override)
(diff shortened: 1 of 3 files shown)


snowman2:
Thanks for taking a stab at this :+1: . The issue has been addressed by #902.

PR #872, opened 2021-07-02
Update FAQ answer about CRS formats.

### LINKED ISSUE #789: "What is the best format to store the CRS information?" in FAQ is not answering the question
https://pyproj4.github.io/pyproj/latest/gotchas.html#what-is-the-best-format-to-store-the-crs-information is explaining that PROJ strings are a bad idea, but provide no answer what is the proper format

Answer seems to be in https://proj.org/faq.html#what-is-the-best-format-for-describing-coordinate-reference-systems

Not making PR as I am unsure whatever it would be better to change question "Why PROJ strings are a bad idea?" or add an aswer.

direvus:
The answer now directly addresses the question, and also aligns with the
advice in the PROJ and geopandas documentation.

Fixes #789.

 - [x] Closes #789 

### DIFF
--- a/docs/gotchas.rst
+++ b/docs/gotchas.rst
@@ -10,8 +10,12 @@ Also see:
   - `PROJ FAQ <https://proj.org/faq.html>`__
 
 
-What is the best format to store the CRS information?
+What are the best formats to store the CRS information?
 -----------------------------------------------------
+
+In general, **Well-Known Text (WKT)** or **Spatial Reference ID (SRID)** are
+the preferred formats to describe a CRS.
+
 PROJ strings can be lossy for storing CRS information.
 If you can avoid it, it is best to not use them.
 Additionally, PROJ strings will likely not be supported
--- a/docs/gotchas.rst
+++ b/docs/gotchas.rst
@@ -13,8 +13,12 @@ Also see:
 What are the best formats to store the CRS information?
 -----------------------------------------------------
 
-In general, **Well-Known Text (WKT)** or **Spatial Reference ID (SRID)** are
-the preferred formats to describe a CRS.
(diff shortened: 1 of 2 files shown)


docs/gotchas.rst
```diff
@@ -10,8 +10,12 @@ Also see:
   - `PROJ FAQ <https://proj.org/faq.html>`__
 
 
-What is the best format to store the CRS information?
+What are the best formats to store the CRS information?
 -----------------------------------------------------
+
+In general, **Well-Known Text (WKT)** or **Spatial Reference ID (SRID)** are
```
jorisvandenbossche, line 16: I would maybe add something like "(such as EPSG codes)" after the spatial reference ID mention, to make it clearer what this is?

direvus: Yeah I could add that.  If you're concerned about readers not knowing what SRID means, would links to the relevant Wikipedia pages be appropriate here?

snowman2: Thanks for this 👍. Also, mind adding this note from the PROJ docs: "If possible WKT2 should be used.".

> would links to the relevant Wikipedia pages be appropriate here?

EPSG codes are the most common, so adding the reference to them will likely be sufficient. But, adding a link would be fine if you would like to.

direvus: Thanks for the feedback, I've updated the PR with those suggestions.

snowman2: Perfect, thanks 👍. Any issues with me squash and merging?

direvus: Squash is fine by me.

snowman2:
Thanks @direvus

PR #886, opened 2021-08-03
BUG: Update GeodIntermediateFlag for future Python compatibility

### LINKED ISSUE #855: pyproj  3.1.0 build fails for python 3.10.0b1
#### Problem description

For the next Fedora cycle all packaged python packages are currently being tested against a preliminary version of python 3.10.0. 
The build is performed in a mock sandbox environment. During this testing it was found that the pyproj package release 3.1.0 fails to build in this environment.

For details of the reported crash see: https://bugzilla.redhat.com/show_bug.cgi?id=1964955

The cause has been tracked down to the pyproj/enum.py module. This one sub-classes enum.IntFlag to create GeodIntermediateFlag, and it seems it is no longer allowed to define this sub-class using a non-consecutive range of bits.
This is an intended change in python 3.10, and was introduced with this commit: 
https://github.com/python/cpython/commit/7aaeb2a3d682ecba125c33511e4b4796021d2f82 but it seems not well documented yet.

A possible solution is to define additional constants that fill the gap between the 3 defined groups of bits in this class.
This patch has been submitted now for Fedora to let the pyproj build pass again. For details see: https://src.fedoraproject.org/rpms/pyproj/blob/rawhide/f/pyproj-enums.patch

If you wish I can provide a pull request that fixes the issue, but as stated in the bugzilla bug report, this seems a backwards incompatible change in Python, so it could as well be reverted before he final python 3.10.0 release. So maybe it is better to wait for that one to happen before modifying the pyproj code here on github.

snowman2:
- [x] Closes #855

@jdkloe, from https://bugs.python.org/issue44242, I noticed this comment:
```
Rather than make such masks containing unknown bits, this would be best practice if you want to use STRICT, correct?

NPTS_ROUND = 0x0
NPTS_CEIL = 0x1
NPTS_TRUNC = 0x2
NPTS_MASK = NPTS_ROUND | NPTS_CEIL | NPTS_TRUNC

Otherwise, if your input may have unknown bits, use CONFORM.
```
So, I updated accordingly. Are you able to verify whether this works with the  Python 3.10.0b1? I am thinking it would be a good idea to add this update now so it isn't an issue in the future if they decide to add this.

cc: @idanmiara


jdkloe:
>     * [x]  Closes [pyproj  3.1.0 build fails for python 3.10.0b1 #855](https://github.com/pyproj4/pyproj/issues/855)
>     * [x]  Fully documented, including `history.rst` for all changes and `api/*.rst` for new API
> 
> @jdkloe, from https://bugs.python.org/issue44242, I noticed this comment:
> 
> ```
> Rather than make such masks containing unknown bits, this would be best practice if you want to use STRICT, correct?
> 
> NPTS_ROUND = 0x0
> NPTS_CEIL = 0x1
> NPTS_TRUNC = 0x2
> NPTS_MASK = NPTS_ROUND | NPTS_CEIL | NPTS_TRUNC
> 
> Otherwise, if your input may have unknown bits, use CONFORM.
> ```
> 
> So, I updated accordingly. Are you able to verify whether this works with the Python 3.10.0b1? I am thinking it would be a good idea to add this update now so it isn't an issue in the future if they decide to add this.

@snowman2 Thanks for working on this issue. I tested again with the current fedora rawhide version which is  python 3.10.0b4.
Using this version I could confirm that:

- the pyproj 3.1.0 release compiles and installs without the need to patch the enum.py file, so it seems that indeed the change was reverted before the python 3.10 release.
- I tested your c0024cf commit (BUG: Update GeodIntermediateFlag for future Python compatibility) and also this version compiles and installs fine with this python version
- I tested your 9cafeac commit (REF: Make GeodIntermediateFlag masks internal only) and  again this version compiles and installs fine with this python version

So to me this looks good. I would be happy to have either of these in the next pyproj release.

snowman2:
Thanks for testing this out and for the report @jdkloe :+1:

idanmiara:
@jdkloe thanks for reporting.
@snowman2 thanks for taking care of it
👍

PR #887, opened 2021-08-04
BUG: Make datum name match exact in query_utm_crs_info

snowman2:
- Related to https://github.com/geopandas/geopandas/issues/2045
 - [x] Tests added


snowman2:
I did a check with all of the UTM CRS in the database and the names all have "/" in the name, so this is a pretty safe change in the present state.

PR #908, opened 2021-08-23
CLN: remove numeric/numarrays support

snowman2:
These have been phased out by numpy. 
- https://wiki.python.org/moin/NumArray
- https://en.wikipedia.org/wiki/NumPy#History

### DIFF
--- a/pyproj/utils.py
+++ b/pyproj/utils.py
@@ -69,30 +69,16 @@ def _copytobuffer(xx: Any) -> Tuple[Any, bool, bool, bool]:
         if xx.shape == ():
             return _copytobuffer_return_scalar(xx)
         else:
-            try:
-                # typecast numpy arrays to double.
-                # (this makes a copy - which is crucial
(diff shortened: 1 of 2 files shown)


pyproj/utils.py
```diff
@@ -69,30 +69,16 @@ def _copytobuffer(xx: Any) -> Tuple[Any, bool, bool, bool]:
...
-                    return inx, False, False, False
-                except Exception:
-                    raise TypeError(
-                        "input must be an array, list, tuple, scalar, "
-                        "or have the __array__ method."
-                    )
+            # typecast numpy arrays to double.
+            # (this makes a copy - which is crucial
+            #  since buffer is modified in place)
+            xx.dtype.char
```
jorisvandenbossche, line 75: I don't think this is still needed?

snowman2: Good eye. See #909 as it is removed there.

PR #965, opened 2021-09-30
fix minor typos

bjlittle:
Well, we've all got to start somewhere...

... noticed some minor typos et al, so thought I'd suggest a change.

Thanks for `pyproj` btw, it's incredible! :beers:


snowman2:
Good eye! Thanks 👍

PR #974, opened 2021-10-08
DEP: Disable pypy tests & make python pin 3.8+

snowman2:
Continuation of #972

After some thought, I think not increasing the Python pin in setup.py will cause confusion/troubles for CPython users without wheels. So, I think pypy support will have to be dropped until pypy supports newer Python versions.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -44,10 +44,11 @@ jobs:
             python-version: 3.8
             python-implementation: python
             proj-version: 8.1
-          - os: ubuntu-latest
-            python-version: '*'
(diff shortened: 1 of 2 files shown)


mattip:
PyPy now supports python3.8. You can enable it in CI as `pypy-3.8`

snowman2:
Thanks @mattip 👍. Do you happen to know if it has been updated in conda-forge?

mattip:
3.8 is getting there, I need to finish [the PR to enable it on windows](https://github.com/conda-forge/pypy3.6-feedstock/pull/61)

PR #996, opened 2021-11-09
CI: Activate python 3.10 & skip shapely install for osx Python 3.10

snowman2:


snowman2:
Shapely skipped osx for Python 3.10, so that is likely the cause of failure: https://github.com/conda-forge/shapely-feedstock/pull/91

PR #1018, opened 2022-01-24
crs.py: remove @abstractmethod decorator

hemberger:
According to the documentation for the `@abstractmethod` decorator:

> Using this decorator requires that the class's metaclass is
> ABCMeta or is derived from it.

(source: https://docs.python.org/3/library/abc.html#abc.abstractmethod)

Since `CustomConstructorCRS` and `CRS` are not defined as abstract
classes, using the `@abstractmethod` decorator does not actually do
anything and should be removed.

 - [ ] Tests added

### DIFF
--- a/pyproj/crs/crs.py
+++ b/pyproj/crs/crs.py
@@ -7,7 +7,6 @@ import json
 import re
 import threading
 import warnings
-from abc import abstractmethod
 from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 
 from pyproj._crs import (
@@ -1591,7 +1590,6 @@ class CustomConstructorCRS(CRS):
     """
 
     @property
-    @abstractmethod
     def _expected_types(self) -> Tuple[str, ...]:
         """
         These are the type names of the CRS class


hemberger:
A couple notes about this PR:

1. Even though the python interpreter ignores `@abstractmethod` here, it can still cause a problem for linters. For example:
 
    ```python
    import abc 
    import pyproj.crs

    class Foo(pyproj.crs.CustomConstructorCRS, metaclass=abc.ABCMeta):
        pass

    Foo()
    ```

    This runs properly, but pylint will report the following error:
    ```
    $ pylint --errors-only test.py
    ************* Module test
    test.py:6:0: E0110: Abstract class 'Foo' with abstract methods instantiated (abstract-class-instantiated)
    ```
    While this is a false positive that might be fixable in pylint, it is still a symptom of `@abstractmethod` being used improperly.

2. An alternative solution would be to make `CustomConstructorCRS` an abstract class, but this seems too heavy-handed since it would break backwards compatibility for any class that derived from `CustomConstructorCRS`, but which has not implemented the `_expected_types` method (which has an entirely reasonable implementation in the `CustomConstructorCRS` class). This is the case with, for example, the `CRS` wrapper class in the cartopy module.

Please let me know if any changes need to be made to this PR to make it suitable for merging. Thank you for your time!

snowman2:
Thanks @hemberger :+1:

PR #1103, opened 2022-07-10
BUG: Ensure numpy masked arrays stay masked after projection

### LINKED ISSUE #1102: Projecting numpy masked arrays returns plain numpy arrays
While upgrading an older system from pyproj 2.6.1 to the 3.x series I found that when projecting arrays of coordinates (as a numpy masked array) that in pyproj 2.6.1 masked arrays of the projected coordinates were returned, however as of the 3.x series the arrays are no longer masked (i.e. "plain" numpy arrays).  I'm not sure if this an intentional change or if this is a regression of some kind.  I couldn't find a GH issue relating to this problem, so I thought I'd let you know about it.  I bisected the issue and found that commit where the problem first appears is 4ab3ff7cf2e3ff089509b921f103dd2fe57ddfda.

If you require any more information (beyond that mentioned below), please just let me know and I'll be more than happy to provide it.

#### Code Sample

Here's a sample piece of code which illustrates the problem (referred to as `test_projected_masked_arrays.py` further below):

```python
# -*- coding: utf-8 -*-

import numpy as np
from pyproj import Proj
from unittest import TestCase


class TestProjectMaskedArrays(TestCase):
    def test_projected_masked_array_is_masked(self):
        lat = np.ma.array(data=[30, 35, 40, 45], mask=[0, 0, 1, 0])
        lon = np.ma.array(data=[0, 5, 10, 15], mask=[0, 0, 1, 0])

        proj = Proj('+ellps=WGS84 +proj=stere +lat_0=75.0 '
                    '+lon_0=-14.0 +x_0=0.0 +y_0=0.0 +no_defs')

        x, y = proj(lon, lat)

        self.assertTrue(hasattr(x, "mask"))
        self.assertTrue(hasattr(y, "mask"))

# vim: expandtab shiftwidth=4 softtabstop=4
```

Setting up a virtual environment and installing base packages (assuming a Debian-bullseye system with Python 3.9):

```bash
$ virtualenv --python=/usr/bin/python3 venv
$ source venv/bin/activate
$ pip install numpy==1.21.6 pytest==7.1.2
$ pip install pyproj==2.6.1 && pytest test_projected_masked_arrays.py  # passes
$ pip install pyproj==3.0.0 && pytest test_projected_masked_arrays.py  # fails
# last good commit
$ pip install --force-reinstall git+https://github.com/pyproj4/pyproj.git@8eb145e13ba8133ba624fa2cc1cbc0a0733d69c0 && pytest test_projected_masked_arrays.py  # passes
# first bad commit
pip install --force-reinstall git+https://github.com/pyproj4/pyproj.git@4ab3ff7cf2e3ff089509b921f103dd2fe57ddfda && pytest test_projected_masked_arrays.py  # fails
```

#### Problem description

My expectation is that the masked array behaviour from the 2.x pyproj series would continue (this could be an incorrect expectation; I'm not sure!).  Specifically, i expect that projecting masked arrays would return masked arrays.

#### Expected Output

I would expect that after executing

```python
x, y = proj(lon, lat)
```

where `lon` and `lat` are numpy masked arrays, that `x` and `y` would have the `mask` attribute.  I.e. that both `hasattr(x, "mask")` and `hasattr(y, "mask")` return `True`.

#### Environment Information

 - Output from: `python -m pyproj -v`

```
pyproj info:
    pyproj: 3.0.dev0
      PROJ: 7.2.1
  data dir: /usr/share/proj

System:
    python: 3.9.2 (default, Feb 28 2021, 17:03:44)  [GCC 10.2.1 20210110]
executable: /venv/bin/python
   machine: Linux-5.10.0-0.bpo.8-amd64-x86_64-with-glibc2.31

Python deps:
       pip: 20.3.4
setuptools: 44.1.1
    Cython: None
```

 - PROJ version (`python -c "import pyproj; print(pyproj.proj_version_str)"`)

```
7.2.1
```

 - PROJ data directory (`python -c "import pyproj; print(pyproj.datadir.get_data_dir())"`)

```
/usr/share/proj
```

 - Python version (`python -c "import sys; print(sys.version.replace('\n', ' '))"`)

```
3.9.2 (default, Feb 28 2021, 17:03:44)  [GCC 10.2.1 20210110]
```

 - Operation System Information (`python -c "import platform; print(platform.platform())"`)

```
Linux-5.10.0-0.bpo.8-amd64-x86_64-with-glibc2.31
```

#### Installation method

Via `pip` in a virtual environment.

paultcochrane:
As noted in #1102, projecting numpy masked arrays returned numpy masked
arrays in the 2.x series of pyproj.  This behaviour changed in commit
4ab3ff7 as part of the 3.x series, where a "plain" numpy ndarray was returned.  The change
implemented here ensures that projecting numpy masked arrays returns
masked arrays as was previously the case.

I'm fairly sure that the fix I've implemented here does actually fix the issue mentioned in #1102; nevertheless, please review my changes carefully as I'm not very familiar with this codebase.

 - [x] Closes #1102 
 - [x] Tests added

### DIFF
--- a/pyproj/utils.py
+++ b/pyproj/utils.py
@@ -109,7 +109,14 @@ def _copytobuffer(xxx: Any, inplace: bool = False) -> Tuple[Any, DataType]:
         The copy of the data prepared for the PROJ API & Python Buffer API.
     """
     # check for pandas.Series, xarray.DataArray or dask.array.Array
-    if hasattr(xxx, "__array__") and callable(xxx.__array__):
+    # also handle numpy masked Arrays; note that pandas.Series also has a
+    # "mask" attribute, hence checking for simply the "mask" attr in that
+    # case isn't sufficient
+    if (
+        not hasattr(xxx, "hardmask")
+        and hasattr(xxx, "__array__")
+        and callable(xxx.__array__)
+    ):
         xxx = xxx.__array__()
 
     # handle numpy data
--- a/docs/history.rst
+++ b/docs/history.rst
@@ -10,6 +10,7 @@ Latest
 - REF: use regex to process PROJ strings in :meth:`.CRS.to_dict` (pull #1086)
 - BUG: :class:`.MercatorAConversion` defined only for lat_0 = 0 (issue #1089)
 - BUG: Add support for `PROJ_DATA` environment variable (issue #1097)
+- BUG: Ensure numpy masked arrays stay masked after projection (issue #1102)
 
 3.3.1
 -------
--- a/test/test_utils.py
+++ b/test/test_utils.py
@@ -42,6 +42,13 @@ def test__copytobuffer__numpy_array(in_arr):
     )
 
 
+def test__copytobuffer__numpy_masked_array():
+    in_arr = numpy.ma.array([1])
+    out_arr, dtype = _copytobuffer(in_arr)
+
+    assert isinstance(out_arr, numpy.ma.MaskedArray)
+
+
 def test__copytobuffer__fortran_order():
     data = numpy.ones((2, 4), dtype=numpy.float64, order="F")
     converted_data, dtype = _copytobuffer(data)


pyproj/utils.py
```diff
@@ -108,6 +108,11 @@ def _copytobuffer(xxx: Any, inplace: bool = False) -> Tuple[Any, DataType]:
     Tuple[Any, DataType]
         The copy of the data prepared for the PROJ API & Python Buffer API.
     """
+    # handle numpy masked Arrays; note that pandas.Series also has a "mask"
+    # attribute, hence checking for simply the "mask" attr isn't sufficient
+    if hasattr(xxx, "hardmask"):
+        return xxx, DataType.ARRAY
+
     # check for pandas.Series, xarray.DataArray or dask.array.Array
     if hasattr(xxx, "__array__") and callable(xxx.__array__):
```
snowman2, line 117: If you remove the line above and do this, does it still work?
```suggestion
    if not hasattr(xxx, "hardmask") and hasattr(xxx, "__array__") and callable(xxx.__array__):
```
I am wondering if dtypes and other functionality are handled properly otherwise.

paultcochrane: Thanks for the review! Yeah, I was wondering about things like that myself, but wasn't sure.  I'll give this a go and let you know.

paultcochrane: I've made your suggested change, which worked as you suggested, (I modified the comment to integrate the information about which attribute to check for masked arrays), reformatted the code using `black`[1], rebased to `main` and have force-pushed my changes, so everything should be good to go.

[1] BTW: the docs mention using `make lint`, but there's no such target there anymore.  Should that be added back in?

snowman2:
Thanks @paultcochrane :+1:

PR #1104, opened 2022-07-10
Fix issues in transformation grids docs

paultcochrane:
While reading through the transformation grids documentation, I noticed a few issues which might need to be fixed.  Since all are associated with the transformation grids documentation, they all part of one PR.  Nevertheless, I've made several commits so that they can be cherry-picked as desired.

This PR is submitted in the hope that it is useful; if you have any questions, comments or wish for any changes, please don't hesitate to let me know :-)


snowman2:
Thanks 👍

PR #1121, opened 2022-08-23
DEP: Add python 3.11 support

### LINKED ISSUE #1110: DEP: Python 3.11
https://devguide.python.org/versions/

Release: 2022-10-03

snowman2:
- [x] Closes #1110
 - [x] Tests added

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -102,7 +102,7 @@ jobs:
           cp "$VCPKG_INSTALLATION_ROOT/installed/${{ matrix.triplet }}/share/proj/data/"* ${GITHUB_WORKSPACE}/pyproj/proj_dir/share/proj/
 
       - name: Build wheels
-        uses: pypa/cibuildwheel@v2.8.1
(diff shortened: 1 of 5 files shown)


.github/workflows/release.yaml
```diff
@@ -102,7 +102,7 @@ jobs:
           cp "$VCPKG_INSTALLATION_ROOT/installed/${{ matrix.triplet }}/share/proj/data/"* ${GITHUB_WORKSPACE}/pyproj/proj_dir/share/proj/
 
       - name: Build wheels
-        uses: pypa/cibuildwheel@v2.8.1
+        uses: pypa/cibuildwheel@v2.9.0
```
snowman2, line 105: v2.9 adds Python 3.11 wheel builds

> CPython 3.11 wheels are now built by default - without the CIBW_PRERELEASE_PYTHONS flag. It's time to build and upload these wheels to PyPI! This release includes CPython 3.11.0rc1, which is guaranteed to be ABI compatible with the final release.

PR #1156, opened 2022-10-03
Fix/20221203/fixed dist info requirement for deploy

### LINKED ISSUE #1155: Missing dist-info breaks project on version 3.4.0
The new version 3.4.0 requires the presence of the METADATA file on dist-info folder, otherwise the project breaks and is unable to load.

https://github.com/pyproj4/pyproj/blob/bc7805ccc813b40ab28099ef305821300f149aad/pyproj/__init__.py#L70
This is the only place where this information is ever looked and in no way it is important or vital for the correct operation of the application.

This is a breaking change for all applications where the deployment strips metadata and other files to reduce size.

Scoppio:
Removes dependency of dist-info folder inside the `__init__.py` file, fixing break when running the package in "slim" environment due to the lack of the METADATA file.

 - [x] Closes #1155
 - [X] Fully documented, including `history.rst` for all changes

### DIFF
--- a/docs/history.rst
+++ b/docs/history.rst
@@ -4,6 +4,10 @@ Change Log
 Latest
 -------
 
+3.4.1
+-----
+- BUG: Changed so that the setup.cfg depends on the version code in the __init__.py instead of the other way around (issuue #1155)
+
 3.4.0
 -----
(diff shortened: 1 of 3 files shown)


snowman2:
@Scoppio, thanks 👍

PR #1162, opened 2022-10-14
REF: Use upper case EPSG code when creating CRS

snowman2:
The EPSG database uses the uppercase version. This is for consistency with `to_authority`.


pyproj/crs/crs.py
```diff
@@ -147,7 +147,7 @@ def _prepare_from_authority(auth_name: str, auth_code: Union[str, int]):
 
 
 def _prepare_from_epsg(auth_code: Union[str, int]):
-    return _prepare_from_authority("epsg", auth_code)
+    return _prepare_from_authority("EPSG", auth_code)
```
snowman2, line 150: This is the main change when creating CRS `from_epsg` integer.

PR #1165, opened 2022-10-24
ENH: Add return_back_azimuth kwarg to fwd, inv, fwd_intermediate, inv_intermediate

### LINKED ISSUE #1163: Lack of consistency between `fwd` and `fwd_intermediate` azimuth output
#### Problem description

I've noticed an inconsistency between the azimuth output of `fwd` and `fwd_intermediate`. In the example below, I compute the lon/lat/azi  when moving 1km from London with an initial forward azimuth of 45 degrees with the both methods.  As expected from the documentation, `fwd` returns the backward azimuth whereas `fwd_intermediate` returns the forward azimuth of the terminus point. For consistency and to avoid confusion, I suggest `fwd_intermediate` also return the backward azimuth.

#### Code Sample

```python
from pyproj import Geod, enums

wgs84 = Geod(ellps="WGS84")

# Inputs 
london = (-0.118092, 51.509865) #lon/lat
dist = 1e3  # m
bearing = 45  # deg
flags = enums.GeodIntermediateFlag(0).AZIS_KEEP  # get fwd_intermediate to return azimuths

# fwd
lon, lat, azi = wgs84.fwd(*london, bearing, dist)

# fwd_intermediate (return first and last point)
res = wgs84.fwd_intermediate(*london, bearing, 2, dist, flags=flags, terminus_idx=0, initial_idx=0)

print("lon, lat, azimuth from fwd             :", lon, lat, azi)
print("lon, lat, azimuth from fwd_intermediate:", res.lons[-1], res.lats[-1], res.azis[-1])
```

which outputs:

```console
lon, lat, azimuth from fwd             : -0.10790547005142687 51.51622011204643 -134.99202649526126
lon, lat, azimuth from fwd_intermediate: -0.10790547005142687 51.51622011204643 45.007973504738736
```

#### Environment Information

pyproj info:
    pyproj: 3.4.0
     PROJ: 9.1.0

#### Installation method

Installed via pip

idanmiara:
https://github.com/pyproj4/pyproj/issues/1163

 - [X] Closes #1163
 - Tests added
   - [ ] inv (also test warning)
   - [ ] inv_intermediate (also test warning)
   - [ ] fwd (also test warning)
   - [ ] fwd_intermediate (also test warning)
   - [ ] reverse_azimuth
   - [ ] reverse_azimuth_arr
 - Fully documented
   - [x] `history.rst`
   -  `api/*.rst` for new API
     - [ ] reverse_azimuth
     - [ ] reverse_azimuth_arr


pyproj/utils.py
```diff
@@ -151,3 +152,40 @@ def _convertback(data_type: DataType, inx: Any) -> Any:
     if data_type == DataType.TUPLE:
         return tuple(inx)
     return inx
+
+
+def reverse_azimuth(azi: float, radians: bool = False) -> float:
```
snowman2, line 157: I would say these functions are a better fit for `geod.py` than a generic pyproj utility.

snowman2: It looks like they are only used for testing. Thoughts about moving them to the testing module instead?

snowman2: It may be more performant to move them to `_geod.pyx` and import into `_geod.py`.

idanmiara: > It may be more performant to move them to `_geod.pyx` and import into `geod.py`.

makes sense, would you mind doing that after this merge as I don't have cython installed and I don't have more time to invest in this...

idanmiara:
@snowman2 I think this is ready. 
I could reorganize the commits if you approve this for merging.

snowman2 (approved):
Thanks @idanmiara :+1:

snowman2:
I am not going to merge it quite yet. 3.4.1 has not been released yet and this is targeted for 3.5. 3.4.1 may be replaced by 3.5, but not sure yet.

pyproj/geod.py
```diff
@@ -240,9 +240,17 @@ def __init__(self, initstring: Optional[str] = None, **kwargs) -> None:
...
 
     def fwd(  # pylint: disable=invalid-name
-        self, lons: Any, lats: Any, az: Any, dist: Any, radians: bool = False
+        self,
+        lons: Any,
+        lats: Any,
+        az: Any,
+        dist: Any,
+        radians: bool = False,
+        return_back_azimuth: bool = True,
```
snowman2, line 249: I have been thinking about this and am thinking it would be a good idea to warn users of
the upcoming change.

This change can initially be released with:
```
return_back_azimuth: Optional[bool] = None,
```
and 
```python
if return_back_azimuth is None:
    return_back_azimuth = False
    warnings.warn(
        "The forward azimuth is currently being returned by default. "
        "In the future, the back azimuth will be returned. To opt-in "
        "to the new behavior, set return_back_azimuth=True. "
        "Otherwise, to avoid breaking changes, set return_back_azimuth=False "
        "for pyproj 3.5+."
    )
```
Then, in a future release, the default can be changed to True.

Thoughts?

snowman2: Another idea is to change the default and raise a warning similar to the above if the user has not selected True or False so they know that they are getting the back azimuth now.

idanmiara: > Another idea is to change the default and raise a warning similar to the above if the user has not selected True or False so they know that they are getting the back azimuth now.

Yes, sounds good.

snowman2: Great. @idanmiara, is this something you have time to add to this PR?

idanmiara: yes. added and rebased.

snowman2:
Thanks @idanmiara for the changes. We're getting closer. I updated the initial comment on the issue with a checklist of testing & documentation updates that would be good to add. Let me know if you have time to add those.

idanmiara:
> Thanks @idanmiara for the changes. We're getting closer. I updated the initial comment on the issue with a checklist of testing & documentation updates that would be good to add. Let me know if you have time to add those.

All good now?

pyproj/geod.py
```diff
@@ -1067,3 +1115,49 @@ def __eq__(self, other: Any) -> bool:
             return False
 
         return self.__repr__() == other.__repr__()
+
+
+try:
+    import numpy as np  # type: ignore  # pylint: disable=import-outside-toplevel
+
+    def reverse_azimuth(azi: Any, radians: bool = False) -> Any:
```
snowman2, line 1123: This function does not appear to behave the same as the one below and changes in place based on the existence of numpy. You should be able to re-write the function to work on a numpy array without needing numpy. I recommend using that approach as it is easier to test and ensure that the behavior is consistent with or without numpy.

snowman2: This may be useful:
- https://docs.python.org/3/library/math.html#math.copysign
- https://stackoverflow.com/a/1986776/11622836

idanmiara: Yes, I thought about using `copysign`, but how would you apply this to a numpy array?
I couldn't find a way to write that function that will work on numpy array and on scalars without importing numpy.

snowman2: Some functions in the math module work on numpy arrays, but that one unfortunately doesn't. You will likely have to do a workaround for numpy arrays. Something like this:

```python
def sign(data):
    data_sign = data.copy()
    neg = data < 0
    data_sign[neg] = -1
    data_sign[~neg] = 1
    return data_sign
```

idanmiara: This function would work on numpy array but not on scalars, so if you write two versions of the functions, why not use the built `sign` method as I did?

snowman2: It is an equivalent amount of code and easier to test/maintain if you don't use numpy.

snowman2: Actually, I have an idea that I think we both will like. Will share later when I am not on my phone.

idanmiara: I see that numpy has also `copysign` so maybe now it's better.

snowman2: My main goal is to remove numpy from the core codebase. Otherwise, the tests and CI will need to be refactored to support both testing with and without numpy. The added complexity doesn't sound fun to maintain.

Here is an alternative approach: https://github.com/talos-gis/pyproj/pull/1

It uses cython instead of numpy and it supports all of the input scalar/array types that the geodesic routines support. This is easier to test and handles more input types in a performant way when numpy isn't present.

idanmiara: Great, thanks for looking into this 👍

snowman2:
This looks great - thanks @idanmiara :+1:. I think this is ready to go. Would you like to clean up the commit history or would you mind if I squashed on merge?

idanmiara:
> This looks great - thanks @idanmiara 👍. I think this is ready to go. Would you like to clean up the commit history or would you mind if I squashed on merge?

Thanks for your help!
I've cleaned the commit history.

PR #1169, opened 2022-10-29
WHL: Add win32 to build_wheels matrix

### LINKED ISSUE #1168: WHL: win32 wheels
#### Problem description

Until version 3.3.1, it is possible to find `win32` wheels on PyPI for the supported Python versions, see:
https://pypi.org/project/pyproj/3.3.1/#files

Since version 3.4.0, the `win32` wheels are not available anymore, see:
https://pypi.org/project/pyproj/3.4.0/#files

These means that for Python 3.11 there is no `win32` wheel for any `pyproj` version, and any library installation depending on `pyproj` will fail for Python 3.11 on `win32` due to lack of precompiled wheels (e.g. [`basemap`](https://github.com/matplotlib/basemap)).

#### Expected Output

I would expect `win32` wheels for `pyproj` 3.4.0 for the supported Python versions on PyPI.

I am not sure if this support drop was actually intended, since I saw some slight reorganisation in the `.github/workflows` folder. I can provide some support to fix this if desired.

#### Environment Information

- pyproj version: 3.4.0
- Python version: 3.11.0
- OS information: Windows-10-10.0.22621-SP0

#### Installation method
 - pip wheel

molinav:
This PR extends the GitHub workflow build matrix to include one more item with variables `{os: "windows-2019": arch: "auto32", triplet: "x86-windows"}`, similarly to the already present win64 build. 

- [X] Closes #1168
- [ ] Tests added
  - No new tests, but I checked that the current tests pass on win32 (as run by cibuildwheel).
  - I have not updated `history.rst` because I was not sure if this addition was that relevant (wheels were already there in `pyproj` 3.3.1, so this PR simply brings them back).

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -75,6 +75,11 @@ jobs:
           triplet: "x64-windows"
           vcpkg_cache: "c:\\vcpkg\\installed"
           vcpkg_logs: "c:\\vcpkg\\buildtrees\\**\\*.log"
+        - os: "windows-2019"
+          arch: "auto32"
+          triplet: "x86-windows"
+          vcpkg_cache: "c:\\vcpkg\\installed"
+          vcpkg_logs: "c:\\vcpkg\\buildtrees\\**\\*.log"
 
     steps:
       - uses: actions/checkout@v2


snowman2:
Thanks @molinav :+1:

PR #1178, opened 2022-11-11
BUG: Fix memory leak in CRS.list_authority

skogler:
List elements were not properly disposed of when iterating the results of proj_identify. This fixes this and simplifies the cleanup logic.

 - [ ] Tests added (not sure how to test this automatically)

You can reproduce this issue with the following snippet (look at the memory usage):

```python
from pyproj import CRS

for i in range(10000):
    CRS.from_epsg(3857).to_epsg()
```

### DIFF
--- a/pyproj/_crs.pyx
+++ b/pyproj/_crs.pyx
@@ -2883,9 +2883,13 @@ cdef class _CRS(Base):
                             out_confidence_list[iii]
                         )
                     )
+                # at this point, the auth name is copied and we can release the proj object
+                proj_destroy(proj)
+                proj = NULL
         finally:
-            for iii in range(num_proj_objects):
-                proj_destroy(proj_list_get(self.context, proj_list, iii))
+            # If there was an error we have to call proj_destroy
+            # If there was none, calling it on NULL does nothing
+            proj_destroy(proj)
             proj_list_destroy(proj_list)
             CRSError.clear()
         return authority_list
--- a/docs/history.rst
+++ b/docs/history.rst
@@ -8,6 +8,7 @@ Latest
 -----
 - BUG: Changed so that the setup.cfg depends on the version code in the __init__.py instead of the other way around (issuue #1155)
 - BUG: Fix :meth:`.CRS.to_cf` for Pole rotation GRIB convention (pull #1167)
+- BUG: Fix :meth:`.CRS.to_authority` memory leak (pull #1178)
 - REF: Use upper case EPSG code when creating CRS (pull #1162)
 
 3.4.0


pyproj/_crs.pyx
```diff
@@ -2834,15 +2834,19 @@ cdef class _CRS(Base):
         # get list of possible matching projections
         cdef PJ_OBJ_LIST *proj_list = NULL
         cdef int *c_out_confidence_list = NULL
-        cdef int num_proj_objects = -9999
+        cdef int num_proj_objects = 0
```
snowman2, line 2837: This looks like the only change needed.

pyproj/_crs.pyx
```diff
@@ -2834,15 +2834,19 @@ cdef class _CRS(Base):
...
         cdef bytes b_auth_name
         cdef char *user_auth_name = NULL
         cdef int iii = 0
+        cdef PJ * proj = NULL
```
snowman2, line 2841: Please don't change the location of the cdef for these. It is preferred to define these closer to where they are going to be used.

pyproj/_crs.pyx
```diff
@@ -2854,24 +2858,16 @@ cdef class _CRS(Base):
             )
             if proj_list != NULL:
                 num_proj_objects = proj_list_get_count(proj_list)
-            if c_out_confidence_list != NULL and num_proj_objects > 0:
```
snowman2, line 2857: Please don't move this.

snowman2: It is best to cleanup objects as soon as they are no longer needed.

pyproj/_crs.pyx
```diff
@@ -2883,9 +2879,14 @@ cdef class _CRS(Base):
                             out_confidence_list[iii]
                         )
                     )
+                # at this point, the auth name is copied and we can release the proj object
+                proj_destroy(proj)
+                proj = NULL
         finally:
-            for iii in range(num_proj_objects):
```
snowman2, line 2887: This is here because the entire list may not be iterated through and not all of the objects will be accessed due to the confidence filter.

pyproj/_crs.pyx
snowman2: Actually, this seems safe:
```python
>>> list(range(-9999))
[]
```

skogler: Oh, I did not know you can do this, its a bit strange though I think. Do you prefer the -9999 default or the 0? I would tend to think 0 is a bit more natural for most readers.

snowman2: It would be fine to change to 0 for readability.

pyproj/_crs.pyx
skogler: As I understand the proj docs, every time you call `proj_list_get` you will get a copy of the element that you need to clean up.

pyproj/_crs.pyx
skogler: I can split it into two try/finally blocks again, but then the error handling is a bit more complex and the code is harder to read. But whatever you prefer is fine.

pyproj/_crs.pyx
skogler: Yeah, see https://github.com/OSGeo/PROJ/blob/master/src/iso19111/c_api.cpp#L8313

snowman2: Yes, you are correct. I see it calls `pj_obj_create` on each call. So, you only need to delete on access.

pyproj/_crs.pyx
skogler: Pushed a new change that is more minimal and does not touch this.

snowman2:
Thanks @skogler :+1:

skogler:
Happy to help, thanks for the free software :)

skogler:
@snowman2 I searched for other usages of `proj_list_get` and found one here: https://github.com/pyproj4/pyproj/blob/main/pyproj/_transformer.pyx#L204

What do you think about that one? It seems to me that https://github.com/pyproj4/pyproj/blob/main/pyproj/_transformer.pyx#L116 should probably call `roj_destroy` on all the child transformer objects?

snowman2:
> What do you think about that one?

I believe that should be fine. The `_Transformer` or `CoordinateOperation` object handles de-allocation of the objects.

skogler:
Ah, I see, because Base is doing, it, you are right!

snowman2:
Mind rebasing from main? I added a fix (#1179) that should allow the CICD tests to run to completion now.

skogler:
Sure, done.

snowman2:
Thanks @skogler :+1:

PR #1187, opened 2022-11-27
BUG: Fix Transform options array length 

shadchin:
After #1123 length options should be increased


snowman2:
Good catch. Thanks @shadchin 👍

PR #1201, opened 2022-12-16
ENH: Add inplace kwarg to Geod.fwd and Geod.inv

greglucas:
This adds the ability to reuse the arrays passed in and out of the geod functions, following the same idea as the Transformers. I noticed this while profiling some of Cartopy's code and we could potentially get some speedup over there from this. We spend a lot of time in `_copytobuffer`.

I'm not sure if you would want the inplace applied to all potential inputs (i.e. `ind` as well) or just the outputs. Seems best to make it all to avoid the copy if possible.

 - [x] Tests added

### DIFF
--- a/pyproj/geod.py
+++ b/pyproj/geod.py
@@ -240,7 +240,13 @@ class Geod(_Geod):
         )
 
     def fwd(  # pylint: disable=invalid-name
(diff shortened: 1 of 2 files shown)


snowman2:
Looks good. Thanks :+1:

PR #1202, opened 2022-12-16
ENH: Add a fast-path for inplace transforms

greglucas:
Rather than opening an issue, I figured it would be easier to push up a PR with code for discussion purposes. This is looking at micro-optimizations when repeatedly calling `Transformer.transform()` on a single point at a time (yes, it is way better to make a large array first, but unfortunately that isn't always feasible). When transforming small arrays, there is a very large overhead (~40%) associated with calling into `_copytobuffer`/`_convertback` on every input/output array.

I'm wondering if anyone has any thoughts on making a fast-path for inplace arrays to bypass the utils functions? Right now the code is very friendly and if a user requests inplace, but the array doesn't work as input array it will gracefully transform it to an array that works, but this adds quite a few checks along the way that are inconsequential for large arrays, but make a difference for small arrays.

Essentially what I want is access to the private Cython inplace `_transform()` method, so perhaps another way to handle this would be to expose that in a public way somehow?

Benchmarking code:

```python
import array

import numpy as np
from pyproj import Transformer

transformer = Transformer.from_crs(2263, 4326)
x_coords = np.random.randint(80000, 120000)
y_coords = np.random.randint(200000, 250000)

x_np = x_coords * np.ones(1, dtype=np.float64)
y_np = y_coords * np.ones(1, dtype=np.float64)

x_list = x_np.tolist()
y_list = y_np.tolist()

x_arr = array.array('d', x_list)
y_arr = array.array('d', y_list)

# Repeat the calculation a bunch of times
for i in range(200000):
    # transformer.transform(x_np, x_np, inplace=False)
    # transformer.transform(x_np, x_np, inplace=True)
    # transformer.transform(x_list, y_list)
    # transformer.transform(x_arr, y_arr, inplace=False)
    transformer.transform(x_arr, y_arr, inplace=True)
```

Then running:
```
python -m cProfile -o prof-arr test_script.py
snakeviz prof-arr
```

 - [ ] Closes #xxxx
 - [ ] Tests added

### DIFF
--- a/pyproj/transformer.py
+++ b/pyproj/transformer.py
@@ -789,6 +789,22 @@ class Transformer:
         '33  98'
 
         """
+        if inplace and getattr(xx, "typecode", "") == "d":
+            self._transformer._transform(
+                xx,
+                yy,
+                inz=zz,
+                intime=tt,
+                direction=direction,
+                radians=radians,
+                errcheck=errcheck,
+            )
+            out = (xx, yy)
+            if zz is not None:
+                out += (zz,)
+            if tt is not None:
+                out += (tt,)
+            return out
         # process inputs, making copies that support buffer API.
         inx, x_data_type = _copytobuffer(xx, inplace=inplace)
         iny, y_data_type = _copytobuffer(yy, inplace=inplace)
--- a/pyproj/geod.py
+++ b/pyproj/geod.py
@@ -298,6 +298,9 @@ class Geod(_Geod):
         scalar or array:
             Back azimuth(s)
         """
+        if inplace:
+            self._fwd(lons, lats, az, dist, radians=radians)
+            return lons, lats, az
         # process inputs, making copies that support buffer API.
         inx, x_data_type = _copytobuffer(lons, inplace=inplace)
         iny, y_data_type = _copytobuffer(lats, inplace=inplace)
@@ -368,6 +371,9 @@ class Geod(_Geod):
             Distance(s) between initial and terminus point(s)
             in meters
         """
+        if inplace:
+            self._inv(lons1, lats1, lons2, lats2, radians=radians)
+            return lons1, lats1, lons2
         # process inputs, making copies that support buffer API.
         inx, x_data_type = _copytobuffer(lons1, inplace=inplace)
         iny, y_data_type = _copytobuffer(lats1, inplace=inplace)


pyproj/transformer.py
```diff
@@ -789,6 +789,22 @@ def transform(  # pylint: disable=invalid-name
         '33  98'
 
         """
+        if inplace and getattr(xx, "typecode", "") == "d":
```
snowman2, line 792: Thoughts about moving this to `_copytobuffer` instead?

snowman2: You could move it to the top of the function so it would skip everything else if true.

snowman2:
Another alternative to consider is to add a method called `transform_point` that is optimized for transforming scalars.

pyproj/transformer.py
greglucas: Yeah, I tried that too and it unfortunately still adds quite a bit of overhead with both of those. 1.5 seconds vs 1 second on my laptop here.

Although, both of those would be "easy" immediate gains.

greglucas:
> Another alternative to consider is to add a method called transform_point that is optimized for transforming scalars.

I like this idea! It could be a bit confusing that you'd need a buffer array to pass to it for the speed though? (i.e. `_copytobuffer_returnscalar` wouldn't be good enough I don't think) I also want the fast-path for Geod fwd/inv too...

snowman2:
> I like this idea! It could be a bit confusing that you'd need a buffer array to pass to it for the speed though? (i.e. `_copytobuffer_returnscalar` wouldn't be good enough I don't think)

I was thinking of adding `_transform_point` to the cython transformer as well. Under the hood it could use [proj_trans](https://proj.org/development/reference/functions.html#c.proj_trans) instead of what `_transform` is using for arrays. This should add significant speedups as you wouldn't require any transformations between types or type checking. You could only pass in floats for `x,y,x,t` and get floats back.

>  I also want the fast-path for Geod fwd/inv too...

That is a good goal. I suggest addressing that issue separately.

greglucas:
OK, I did some benchmarking on the single-point versions, and casting immediately to a single-value float array and using those in the generics is still a massive improvement. I'm sure more can be done by calling into a different Cython function too...

see here: https://github.com/pyproj4/pyproj/pull/1203

PR #1203, opened 2022-12-16
ENH: Add single point versions of transform, fwd, inv

greglucas:
When transforming single points, this can add a significant speed benefit.

 - [x] Tests added
 - Fully documented
 - [ ] including `history.rst` for all changes
 - [x] `api/*.rst` for new API

### DIFF
--- a/pyproj/geod.py
+++ b/pyproj/geod.py
@@ -17,6 +17,7 @@ __all__ = [
 ]
 
 import math
+from array import array
 from typing import Any, Dict, List, Optional, Tuple, Union
 
 from pyproj._geod import Geod as _Geod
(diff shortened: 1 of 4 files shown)


pyproj/geod.py
```diff
@@ -310,6 +311,61 @@ def fwd(  # pylint: disable=invalid-name
...
+        Accepted numeric scalar:
+
+        - :class:`int`
+        - :class:`float`
+        - :class:`numpy.floating`
+        - :class:`numpy.integer`
+
+        Parameters
+        ----------
+        lon: scalar or array
```
snowman2, line 338: I think this is something to look out for in the new docstrings.
```suggestion
        lon: scalar
```

test/test_transformer.py
```diff
@@ -1505,6 +1505,18 @@ def test_4d_transform__inplace__numpy__int():
     assert tarr[0] == 2019
 
 
+def test_transform_single_point():
```
snowman2, line 1508: Might not be a bad idea to parametrize `z` and `t` in this test for the cases that they are None (i.e. not provided).

pyproj/transformer.py
```diff
@@ -820,6 +821,108 @@ def transform(  # pylint: disable=invalid-name
...
+        radians: bool = False,
+        errcheck: bool = False,
+        direction: Union[TransformDirection, str] = TransformDirection.FORWARD,
+    ) -> Tuple[float, float]:
+        ...
+
+    @overload
+    def transform_point(  # pylint: disable=invalid-name
+        self,
+        x: float,
```
snowman2, line 838: It would be good for theses kwargs to be consistent with `transform`.

snowman2:
See #1204 for cython version of `Transformer.transform_point`.

greglucas:
Closing in favor of #1204. Will open a separate PR for the Geod point functions.

idanmiara:
> Closing in favor of #1204. Will open a separate PR for the Geod point functions.

https://github.com/pyproj4/pyproj/pull/1206

Just adding the reference here. Thanks for this work @greglucas @snowman2 !

PR #1204, opened 2022-12-17
PERF: Optimize point transformations

snowman2:
Related: #1202
```python
In [1]: from pyproj import Transformer

In [2]: transformer = Transformer.from_crs(4326, 3857, always_xy=True)

In [3]: %timeit transformer.transform(1, 2)
4.69 µs ± 37.9 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)

In [4]: %timeit transformer.transform_point(1, 2)
2.08 µs ± 20.2 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```
Version in #1203:
```python
In [5]: %timeit transformer.transform_point(1, 2)
2.99 µs ± 132 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```

### DIFF
--- a/pyproj/_transformer.pyx
+++ b/pyproj/_transformer.pyx
@@ -720,6 +720,58 @@ cdef class _Transformer(Base):
 
     @cython.boundscheck(False)
     @cython.wraparound(False)
+    def _transform_point(
+        self,
+        double inx,
+        double iny,
+        double inz,
+        double intime,
+        object direction,
+        bint radians,
+        bint errcheck,
+    ):
+        if self.id == "noop":
+            return inx, iny, inz, intime
+
+        cdef PJ_DIRECTION pj_direction = get_pj_direction(direction)
+        cdef PJ_COORD projxyout
+        cdef PJ_COORD projxyin = proj_coord(inx, iny, inz, intime)
+        with nogil:
+            # degrees to radians
+            if not radians and proj_angular_input(self.projobj, pj_direction):
+                projxyin.uv.u *= _DG2RAD
+                projxyin.uv.v *= _DG2RAD
+            # radians to degrees
+            elif radians and proj_degree_input(self.projobj, pj_direction):
+                projxyin.uv.u *= _RAD2DG
+                projxyin.uv.v *= _RAD2DG
+
+            proj_errno_reset(self.projobj)
+            projxyout = proj_trans(self.projobj, pj_direction, projxyin)
+            errno = proj_errno(self.projobj)
+            if errcheck and errno:
+                with gil:
+                    raise ProjError(
+                        f"transform error: {proj_context_errno_string(self.context, errno)}"
+                    )
+            elif errcheck:
+                with gil:
+                    if ProjError.internal_proj_error is not None:
+                        raise ProjError("transform error")
+
+            # radians to degrees
+            if not radians and proj_angular_output(self.projobj, pj_direction):
+                projxyout.xy.x *= _RAD2DG
+                projxyout.xy.y *= _RAD2DG
+            # degrees to radians
+            elif radians and proj_degree_output(self.projobj, pj_direction):
+                projxyout.xy.x *= _DG2RAD
+                projxyout.xy.y *= _DG2RAD
+        ProjError.clear()
+        return projxyout.xyzt.x, projxyout.xyzt.y, projxyout.xyzt.z, projxyout.xyzt.t
+
+    @cython.boundscheck(False)
+    @cython.wraparound(False)
     def _transform_sequence(
         self,
         Py_ssize_t stride,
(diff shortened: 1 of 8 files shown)


pyproj/transformer.py
```diff
@@ -820,6 +821,105 @@ def transform(  # pylint: disable=invalid-name
...
+        radians=False,
+        errcheck=False,
+        direction=TransformDirection.FORWARD,
+    ):
+        """
+        Optimized to transform single point between two coordinate systems.
+
+        See: :c:func:`proj_trans`
+
+        .. versionadded:: 2.5.0
```
greglucas, line 875:
```suggestion
        .. versionadded:: 3.5.0
```

pyproj/transformer.py
```diff
@@ -820,6 +821,105 @@ def transform(  # pylint: disable=invalid-name
...
+        xx,
+        yy,
+        zz=None,
+        tt=None,
+        radians=False,
+        errcheck=False,
+        direction=TransformDirection.FORWARD,
+    ):
+        """
+        Optimized to transform single point between two coordinate systems.
```
greglucas, line 871:
```suggestion
        Optimized to transform a single point between two coordinate systems.
```

pyproj/transformer.py
```diff
@@ -820,6 +821,105 @@ def transform(  # pylint: disable=invalid-name
...
+
+        See: :c:func:`proj_trans`
+
+        .. versionadded:: 2.5.0
+
+        Accepted numeric scalar:
```
greglucas, line 877: Do you want to say `numbers.Real` instead of explicitly listing these out? https://docs.python.org/3/library/numbers.html#numbers.Real

greglucas (commented):
Nice! This works great for me.

pyproj/transformer.py
snowman2: Thanks :+1:

pyproj/transformer.py
snowman2: Thanks :+1:

pyproj/transformer.py
snowman2: Good suggestion. I added it as the type and kept the list as examples.

snowman2:
With #1208 

```
In [3]: transformer = Transformer.from_crs(4326, 3857, always_xy=True)

In [4]: %timeit transformer.transform(1, 2)
1.44 µs ± 38.6 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

In [5]: %timeit transformer.transform(1, 2)
1.42 µs ± 14.1 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

In [6]: %timeit transformer.transform(1, 2)
1.44 µs ± 19.1 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```

pyproj/_transformer.pyx
```diff
@@ -712,6 +712,91 @@ cdef class _Transformer(Base):
...
+        bint radians,
+        bint errcheck,
+    ):
+        """
+        Optimized to transform a single point between two coordinate systems.
+        """
+        cdef double coord_x = inx
+        cdef double coord_y = iny
+        cdef double coord_z = 0
+        cdef double coord_t = HUGE_VAL
```
greglucas, line 733: I guess it doesn't matter if these are attempted to cast here before the instance checks below and will raise a TypeError if it can't cast?

greglucas (approved):
Looks good to me! I get similar timings to you, and not a major slowdown for trying this before going down the other path. So, I do like this better than adding a new method, and theoretically downstream users should be able to get immediate benefits without needing to change the call sites.

pyproj/_transformer.pyx
snowman2: Yes, that is correct.

snowman2: This the only reason for the isinstance check is for 1D numpy arrays.

test/test_transformer.py
```diff
@@ -27,6 +27,24 @@
...
+@pytest.fixture(
+    params=[
+        float,
+        np.array,
+        _make_longer_array,
+    ]
+)
+def transform_point_type(request):
+    """
+    Test both Transformer.transform and Transformer.transform_point
```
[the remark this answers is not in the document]
snowman2: I think I clarified that better now...

snowman2:
> I do like this better than adding a new method, and theoretically downstream users should be able to get immediate benefits without needing to change the call sites.

I agree

pyproj/_transformer.pyx
greglucas: even more minimal than that... single-value numpy arrays is the only reason :( Definitely seems like a poor decision to allow the array->float cast, but I doubt they would want to change upstream, so we're stuck with this I guess.

PR #1205, opened 2022-12-18
PERF: Faster transform direction if already passed an Enum

greglucas:
Currently, `Enum.create()` adds quite a bit of overhead to the transform calls, so we probably don't want to call it when unnecessary. Here I added `isinstance()` checks to handle it in a completely backwards compatible way.

```python
In [1]: from pyproj import Transformer

In [2]: transformer = Transformer.from_crs(2263, 4326, always_xy=True)

In [3]: %timeit transformer.transform(1, 2, direction="forward")
7.96 µs ± 32 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)

In [4]: %timeit transformer.transform(1, 2, direction="FORWARD")
4.18 µs ± 33.9 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)

In [5]: %timeit transformer.transform(1, 2)
3.26 µs ± 21.6 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```

 I think it might be even nicer to remove `Enum.create()` entirely and move over to a `StrEnum` / `BaseEnum(str, Enum)` and then the dictionary lookup could be either `"FORWARD"` or `TransformDirection.FORWARD` with the caveat that you'd lose the case insensitive conversion, but the code would be a bit cleaner everywhere then.

### DIFF
--- a/pyproj/_transformer.pyx
+++ b/pyproj/_transformer.pyx
@@ -104,6 +104,16 @@ dy_dphi: List[float]
     Partial derivative of coordinate.
 """
 
+cdef PJ_DIRECTION get_pj_direction(object direction) except *:
+    # optimized lookup to avoid creating a new instance every time
+    # gh-1205
+    try:
+         return _PJ_DIRECTION_MAP[direction]
+    except KeyError:
+        direction = TransformDirection.create(direction)
+        return _PJ_DIRECTION_MAP[direction]
+
+
 cdef class _TransformerGroup:
     def __cinit__(self):
         self.context = NULL
@@ -627,9 +637,7 @@ cdef class _Transformer(Base):
     ):
         if self.id == "noop":
             return
-
-        tmp_pj_direction = _PJ_DIRECTION_MAP[TransformDirection.create(direction)]
-        cdef PJ_DIRECTION pj_direction = <PJ_DIRECTION>tmp_pj_direction
+        cdef PJ_DIRECTION pj_direction = get_pj_direction(direction)
         cdef PyBuffWriteManager xbuff = PyBuffWriteManager(inx)
         cdef PyBuffWriteManager ybuff = PyBuffWriteManager(iny)
 
@@ -718,8 +726,7 @@ cdef class _Transformer(Base):
     ):
         if self.id == "noop":
             return
-        tmp_pj_direction = _PJ_DIRECTION_MAP[TransformDirection.create(direction)]
-        cdef PJ_DIRECTION pj_direction = <PJ_DIRECTION>tmp_pj_direction
+        cdef PJ_DIRECTION pj_direction = get_pj_direction(direction)
         # private function to itransform function
         cdef double *x
         cdef double *y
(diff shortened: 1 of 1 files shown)


snowman2:
StrEnum looks nice, but is Python 3.11+.

snowman2:
I wonder if it would be simpler to add the change there: https://github.com/pyproj4/pyproj/blob/e345729e4283535acf43da40e3679a4e8fae6152/pyproj/enums.py#L14

```python
if isinstance(item, cls):
    return item
```

greglucas:
Adding the change directly to Enums would be cleaner, but does add a bit of a speed hit still... (I assume going from Cython -> Python -> Cython adds some overhead?)


> StrEnum looks nice, but is Python 3.11+.

I think we could get away with just subclassing string here and defining:
`class TransformDirection(str, BaseEnum)`
there isn't much more to the stdlib implementation:
https://github.com/python/cpython/blob/1cf3d78c92eb07dc09d15cc2e773b0b1b9436825/Lib/enum.py#L1263

I think the question if you want to go this route is whether losing case sensitivity is a deal breaker or not?

snowman2:
To make the code simpler, thoughts about adding this function:
```python
cdef PJ_DIRECTION get_pj_direction(object direction) except *:
    if not isinstance(direction, TransformDirection):
        direction = TransformDirection.create(direction)
    return _PJ_DIRECTION_MAP[direction]
```
and doing this:
```python
cdef PJ_DIRECTION pj_direction = get_pj_direction(direction)
```

snowman2:
> I think we could get away with just subclassing string

Have you tested the performance of this implementation?

greglucas:
> To make the code simpler, thoughts about adding this function:

:+1: That seems like it would be nice!

However, I think we should decide on the next comment first.

> >  I think we could get away with just subclassing string

> Have you tested the performance of this implementation?

Yes, it is actually faster because you don't need an isinstance check. You can go right into your dictionary with both `TransformDirection.FORWARD` and the string `"FORWARD"` because `"FORWARD" == TransformDirection.FORWARD` and both versions will match the key. (thus negating the need for the lookup above)

Downsides are that `"forward" != TransformDirection.Forward` so random cased strings won't work (probably good for Enums actually) and the nice error message of valid options are no longer there.

snowman2:
How much of a performance improvement do you get compared to this PR? I think that if it is significant, I don't think that being case insensitive is as important.

snowman2:
I am wondering if this would be faster and still allow the error message and support case insensitive input:

```python
cdef PJ_DIRECTION get_pj_direction(object direction) except *:
    try:
         return _PJ_DIRECTION_MAP[direction]
    except KeyError:
        direction = TransformDirection.create(direction)
        return _PJ_DIRECTION_MAP[direction]
```

greglucas:
I like your most recent suggestion of the try/except inside a helper function. We can add `StrEnum` later too if desired.

pyproj/_transformer.pyx
```diff
@@ -104,6 +104,14 @@ dy_dphi: List[float]
     Partial derivative of coordinate.
 """
 
+cdef PJ_DIRECTION get_pj_direction(object direction) except *:
+    try:
```
snowman2, line 108: Mind making a note here that this is for performance and reference this PR (https://github.com/pyproj4/pyproj/pull/1205)? This is for a future programmer that wonders why it is done this way when the code could have been simpler.

greglucas: Yep, good suggestion!

snowman2: Thanks!

snowman2:
Thanks @greglucas :+1:

PR #1206, opened 2022-12-19
PERF: Optimize for single point in Geod fwd/inv functions

greglucas:
This adds a fast-path for scalar inputs by trying out a double-only implementation, and falling back if any input is not a double. An alternative is to add `Geod.fwd_point()` and `Geod.inv_point()` functions as well, but I kind of like this transparent form for simplicity of users only needing one function and only a minor overhead for arrays where you are already going through some slower paths.

A little under 10x improvement for scalars.
```python
In [1]: from pyproj import Geod

In [2]: g = Geod(ellps="WGS84")

In [3]: %timeit g.fwd([0], [0], [1], [1])
4.76 µs ± 28.5 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)

In [4]: %timeit g.fwd(0, 0, 1, 1)
588 ns ± 4.57 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```

 - [x] Tests added

### DIFF
--- a/pyproj/_geod.pyx
+++ b/pyproj/_geod.pyx
@@ -168,6 +168,63 @@ cdef class Geod:
 
     @cython.boundscheck(False)
     @cython.wraparound(False)
+    def _fwd_point(
+        self,
+        object lon1in,
+        object lat1in,
+        object az1in,
+        object s12in,
+        bint radians=False,
+        bint return_back_azimuth=True,
+    ):
+        """
+        Scalar optimized function
+        forward transformation - determine longitude, latitude and back azimuth
+        of a terminus point given an initial point longitude and latitude, plus
+        forward azimuth and distance.
+        if radians=True, lons/lats are radians instead of degrees.
+        """
+        cdef double plon2, plat2, pazi2
+
+        # We do the type-checking internally here rather than declaring double as the
+        # type into the function due to automatically casting length-1 arrays to float
+        # that we don't want to return scalar for.
+        # Ex: float(np.array([0])) works and we don't want to accept numpy arrays
+        cdef double lon1 = lon1in
+        cdef double lat1 = lat1in
+        cdef double az1 = az1in
+        cdef double s12 = s12in
+        for x in (lon1in, lat1in, az1in, s12in):
+            if not isinstance(x, (float, int)):
+                raise TypeError("Scalar input is required for point based functions")
+
+        with nogil:
+            if radians:
+                lon1 = _RAD2DG * lon1
+                lat1 = _RAD2DG * lat1
+                az1 = _RAD2DG * az1
+            geod_direct(
+                &self._geod_geodesic,
+                lat1,
+                lon1,
+                az1,
+                s12,
+                &plat2,
+                &plon2,
+                &pazi2,
+            )
+            # back azimuth needs to be flipped 180 degrees
+            # to match what PROJ geod utility produces.
+            if return_back_azimuth:
+                pazi2 =_reverse_azimuth(pazi2, factor=180)
+            if radians:
+                plon2 = _DG2RAD * plon2
+                plat2 = _DG2RAD * plat2
+                pazi2 = _DG2RAD * pazi2
+        return plon2, plat2, pazi2
+
+    @cython.boundscheck(False)
+    @cython.wraparound(False)
     def _inv(
         self,
         object lons1,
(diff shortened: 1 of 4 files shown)


snowman2:
Nice! I am starting to think that following this pattern in #1204 may make sense as well.

greglucas:
I think it would be nice to add this pattern to the `transform()` call as well in #1204. Then downstream users wouldn't need to call a separate function at all and their code would benefit immediately. In addition though, since there is a separate PROJ function for this, do you want to still expose it as `transform_point()` or just leave that as an implementation detail within `transform()`?

pyproj/geod.py
```diff
@@ -298,6 +298,15 @@ def fwd(  # pylint: disable=invalid-name
         scalar or array:
             Back azimuth(s)
         """
+        if not inplace:
```
snowman2, line 301: I don't think this `inplace` check is needed. The reason is that inplace still won't work for scalars in the latter code.

greglucas: I thought the same thing, but the test for `inplace` types fails with numpy arrays of length-1.

Maybe Cython does some auto-casting and allows `np.array([1], dtype=np.float64)` to pass into a `double x` function?

snowman2: That is a gotcha, but kind of makes sense. This is concerning for any array of length 1 as it will be passed into this function by default and get back a scalar instead of an array. Any thoughts on how to handle that?

greglucas: Hm... I'm not sure how to prevent the cast/raise and a quick search didn't turn up anything super obvious.

I think this is only an issue for numpy-like arrays?
```python
float(np.array([0]))  # scalar
float(array.array("d", [0]))  # ValueError
```

snowman2: That is really unfortunate that it is the case. I wonder why numpy allows that.
Due to this behavior, I am thinking it would be better to have a separate interface for the optimized point functions.

greglucas: Yes, this is unfortunate :( Apparently you can typecheck during a typecast in cython `<MyType?>my_type`, but I can't figure out how to do that via the function call... I am not knowledgeable in Cython at all.

Another option which adds a small bit of overhead, but can get rid of the try/except is:
```python
        if (isinstance(lons, (float, int))
            and isinstance(lats, (float, int))
            and isinstance(az, (float, int))
            and isinstance(dist, (float, int))):
```

what are your thoughts about that as an option versus the extra function? (~800 ns in this implementation versus 600 ns with the try/except)

`all(...)` adds a bit of overhead as well, so it is better to check each arg explicitly.

snowman2: https://cython.readthedocs.io/en/latest/src/userguide/extension_types.html#type-testing-and-casting

> To explicitly test the type of an object, use the isinstance() builtin function. For known builtin or extension types, Cython translates these into a fast and safe type check that ignores changes to the object’s __class__ attribute etc., so that after a successful isinstance() test, code can rely on the expected C structure of the extension type and its C-level attributes (stored in the object’s C struct) and [cdef](https://cython.readthedocs.io/en/latest/src/userguide/language_basics.html#cdef)/@cfunc methods.

snowman2: > what are your thoughts about that as an option versus the extra function? (~800 ns in this implementation versus 600 ns with the try/except)

Looks like checking for `float`|`int` types may not be desirable if you are coming from numpy:

```python
>>> import numbers
>>> isinstance(2, numbers.Real)
True
>>> import numpy
>>> isinstance(numpy.int64(2), numbers.Real)
True
>>> isinstance(numpy.array([2]), numbers.Real)
False
>>> isinstance(numpy.int64(2), int)
False
```

greglucas: Inside Cython, if I declare the inputs to the function as `double lons` and then `print(type(lons))` it is always class float, so it has already been cast upon entering the function. Is that suggesting to use `object lons` and then `if not isinstance(lons, (float, int)): raise TypeError; else: double lon = <double>lons`? That seems pretty verbose, but doable if we want.

I'm not following why that the typecheck from numpy would be undesirable? Worst case if you're passing in a numpy scalar int, then you'll just go down the old path because you've failed the isinstance check. If you pass in a `np.float64(2)` numpy scalar, we will return a standard Python scalar instead, but it looks like that already happens in the current methods.

I just pushed up the isinstance() checks from the Python side in a separate commit.

snowman2: >  Is that suggesting to use object lons and then if not isinstance(lons, (float, int)): raise TypeError; else: double lon = <double>lons? That seems pretty verbose, but doable if we want.

Does that have any significant impact on performance?

greglucas: I think it is slightly faster to do that, back down around 600ish ns again. Although the code isn't quite as nice in the C layer, but that is probably OK.

pyproj/_geod.pyx
```diff
@@ -146,6 +146,63 @@ cdef class Geod:
...
+        forward azimuth and distance.
+        if radians=True, lons/lats are radians instead of degrees.
+        """
+        cdef double plon2, plat2, pazi2, lon1, lat1, az1, s12
+
+        # We do the type-checking internally here rather than declaring in the function
+        # due to automatically casting length-1 arrays to float that we don't want
+        for x in (lon1in, lat1in, az1in, s12in):
+            if not isinstance(x, (float, int)):
+                raise TypeError(f"Function only accepts scalar input")
```
snowman2, line 172:
```suggestion
                raise TypeError("Function only accepts scalar input")
```

pyproj/_geod.pyx
```diff
@@ -146,6 +146,63 @@ cdef class Geod:
...
+                lon1,
+                az1,
+                s12,
+                &plat2,
+                &plon2,
+                &pazi2,
+            )
+            # back azimuth needs to be flipped 180 degrees
+            # to match what PROJ geod utility produces.
+            if pazi2 > 0:
```
snowman2, line 196: Mind rebasing from master and replacing this if/elif with:
```python
           pazi2 = _reverse_azimuth(pazi2, factor=180)
```

greglucas: Yep, should be rebased and squashed down to a single commit now.

snowman2:
My main hesitation to merge is the amount of code added versus the amount of code tested. The tests you added are a good start. However, I am thinking it may be a good idea to update the geodesic tests with the `scalar_and_array` fixture before merging this.

snowman2:
Thanks @greglucas 👍

greglucas:
:+1: Thanks, @snowman2!

FYI that this all works really nicely over on Cartopy without any update needed to the code at all. The benchmark case I was running went from 7s on `pyproj==3.4` to 1.1s on `pyproj==main` so roughly 6-7x improvement!

snowman2:
That's great 🎉

dopplershift:
Thanks for doing all this @greglucas !

PR #1208, opened 2022-12-20
PERF: Use isinstance in get_pj_direction

snowman2:
Related: #1205

@greglucas, this seems to be consistently faster for some reason. Can you verify?


greglucas (approved):
:+1: I agree it seems to be a bit faster for both cases (default instance, and "forward"). A few hundred ns or so...

snowman2:
Thanks for verifying 👍

PR #1209, opened 2022-12-20
PERF: Additional performance updates

greglucas:
This is whittling down the last two easy to find performance bottlenecks using snakeviz. See the image below.

image

Commit 1: Removes the __hash__ calls to enum from the dict access. It is small enough and only used in one place that we can inline it.
Commit 2: (far right purple box is `ProjError.clear()`) Removes the class attribute of the internal_proj_error and moves that into Cython to take care of all the error messages without needing to access Python objects while within C.


pyproj/_datadir.pyx
```diff
@@ -19,7 +19,8 @@ _USE_GLOBAL_CONTEXT = strtobool(os.environ.get("PYPROJ_GLOBAL_CONTEXT", "OFF"))
 # static user data directory to prevent core dumping
 # see: https://github.com/pyproj4/pyproj/issues/678
 cdef const char* _USER_DATA_DIR = proj_context_get_user_writable_directory(NULL, False)
-
+# Store the message from any internal PROJ errors
+cdef str _INTERNAL_PROJ_ERROR = None
```
greglucas, line 23: I'm not sure how this behaves across threads compared to the previous implementation of using class attributes to store the information. Do I need to grab a context here when using this?

snowman2: It looks like the type is string, so I imagine that it still uses the Python gil. Due to that, I believe that the behavior here should be consistent with the previous behavior.

snowman2:
This is great! How much does it improve the performance?

greglucas:
I go from about 1.5 us -> 1.1 us / iteration of transform_point (rebased on your branch). Maybe see the results of calling it 2,000,000 times to get the cumulative. Here are the figures below relative to the original one included above, although I'm not sure we really get _that_ much of a speedup.

After the first commit:
image

After the second commit:
image

pyproj/_datadir.pyx
greglucas: Another thing to think about is whether you want `get_proj_error()` and `clear_proj_error()` to leak into the public modules or whether you want those kept as private functions.

snowman2: I think that keeping them private is a good idea.

snowman2:
Every little speedup counts. I think that this is a worthwhile change. Thanks for working on this 👍

pyproj/exceptions.py
```diff
@@ -1,27 +1,25 @@
...
-            )
-            ProjError.clear()
+        proj_error = _get_proj_error()
+        if proj_error is not None:
+            error_message = f"{error_message}: (Internal Proj Error: {proj_error})"
+            _clear_proj_error()
         super().__init__(error_message)
 
     @staticmethod
     def clear() -> None:
```
snowman2, line 18: I think you can safely remove this method.

snowman2: Actually, I think it is fine to ignore that in this PR.

PR #1231, opened 2023-01-15
geod.inv - support for mixed scalar and vector inputs and output selection

idanmiara:
This PR is (trying) to add the option to call `inv` with vectors that are the same length or with length=1 (for example, from (lon1, lat1) to [(lon2, lon3), (lat2, lat3)] and also to select which outputs to return (for example only azi1 and dist)

### DIFF
--- a/pyproj/_geod.pyi
+++ b/pyproj/_geod.pyi
@@ -46,6 +46,9 @@ class Geod:
         lats1: Any,
         lons2: Any,
         lats2: Any,
+        pazi1_out: int,
+        pazi2_out: int,
+        ps12_out: int,
         radians: bool = False,
         return_back_azimuth: bool = False,
     ) -> None: ...
--- a/pyproj/_geod.pyx
+++ b/pyproj/_geod.pyx
@@ -243,6 +243,9 @@ cdef class Geod:
         object lats1,
         object lons2,
         object lats2,
+        int pazi1_out,
+        int pazi2_out,
+        int ps12_out,
         bint radians=False,
         bint return_back_azimuth=True,
     ):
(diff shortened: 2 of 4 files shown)


idanmiara:
@snowman2 WDYT about this idea?

Could you advise about the cython - in this PR in general and specifically this error:
https://ci.appveyor.com/project/jswhit/pyproj-0e5a7/builds/45944305

snowman2:
@idanmiara, would you mind providing code snippets demonstrating the behavior you would like to see?

idanmiara:
> @idanmiara, would you mind providing code snippets demonstrating the behavior you would like to see?

my main motivation would be to mimic numpy broadcasting:
```
import numpy as np
from pyproj import Geod
g = Geod(ellps='clrk66')
lon1 = 30
lat1 = 36
lon2 = np.array([31, 31.1])
lat2 = np.array([37, 37.1])
azi1, azi2, distance = g.inv(lon1, lat1, lon2, lat2)
```

second motivation is to select which output I need:
```
import numpy as np
from pyproj import Geod
g = Geod(ellps='clrk66')
lon1 = np.array([30, 30.1])
lat1 = np.array([36, 36.1])
lon2 = np.array([31, 31.1])
lat2 = np.array([37, 37.1])
distance = g.inv(lon1, lat1, lon2, lat2, output_map=[False, False, False])
```

The second one is by product, `inv` outputs are saved inplace - then it might be more efficient to allow selecting less outputs in cases where not all of them are needed

snowman2:
> my main motivation would be to mimic numpy broadcasting

Can you explain why you would want to do that here versus in numpy?

idanmiara:
> > my main motivation would be to mimic numpy broadcasting
> 
> Can you explain why you would want to do that here versus in numpy?

How would you do that? Do you mean using `np.repeat` ?

snowman2:
@idanmiara, I understand that both of the changes you want are in the same code region and will likely involve code conflicts if you attempt two separate PRs at the same time. However, I ask that you separate the changeset into separate PRs for clarity in both the changes and the discussion as they are different items. If there are conflicts, I recommend we work through the PRs sequentially. I suggest starting with the broadcasting logic first in this PR.

idanmiara:
> @idanmiara, I understand that both of the changes you want are in the same code region and will likely involve code conflicts if you attempt two separate PRs at the same time. However, I ask that you separate the changeset into separate PRs for clarity in both the changes and the discussion as they are different items. If there are conflicts, I recommend we work through the PRs sequentially. I suggest starting with the broadcasting logic first in this PR.

The problem I ran into was that if the input is 2 arrays and 2 scalars then it wouldn't be possible to use 3 outputs as there are no 3 arrays to write the output to. I thought to avoid this problem by limiting the output number to the number of arrays that have the max length, thus adding the output selection, I could also potentially duplicate the data into another array in that case, but I thought that it would also defete the purpose not using `np.repeat` in the first place, and also add more complexity. do you have another idea for avoiding this problem?

idanmiara:
Also - could you point me to the reason for this following error ?
https://ci.appveyor.com/project/jswhit/pyproj-0e5a7/builds/45944305#L198

snowman2:
Is the performance/memory required for this operation with numpy currently causing you troubles?

idanmiara:
> Is the performance/memory required for this operation with numpy currently causing you troubles?

I've been working on some major performance refactor and this was one of the items to check on the list.
I was thinking this would be an easy fix but apparently it's not that simple so I'm thinking to close this one for now unless there are more people who are interested in this.

snowman2:
#1206 may be of interest to you if memory is your limitation.

snowman2:
Thank you for you time an effort considering this approach. I recommend using `numpy.full` if you have mixed scalar & vector inputs.

Closing due to inactivity.

PR #1288, opened 2023-05-17
BUG: fix Geod.npts NaN handling

### LINKED ISSUE #1282: Incorrect behavior of geod.npts
Hello,

This is my first post here, first and foremost I would like to thank you all for doing a fantastic work with this package:)

I am working on a project where pyproj is used and came across a strange behavior. I haven't checked every version to check where exactly did the issue start happening but it seems to be happening for the newest 3.5.0.

Consider the following:

```python
import Geod

geod = Geod(ellps="WGS84")
geod.npts(np.nan, 1.0, 2.0, np.nan, npts=1)  # returns [(np.nan, 2.0)] ?!
geod.npts(1.0, 2.0, 3.0, np.nan, npts=1)  # returns [(1.0, 2.0)] ?!
```

#### Problem description

I would expect `.npts` to always return a tuple of nans when called on input containing nans. This however doesn't seem to be the case. I checked other possibilities with 2x np.nan or 1x np.nan in the input but only these two seem to be problematic.

#### Expected Output

`[(np.nan, np.nan)]` for both

#### Environment Information
pyproj info:
    pyproj: 3.5.0
      PROJ: 9.2.0
  data dir: /usr/local/lib/python3.8/dist-packages/pyproj/proj_dir/share/proj
user_data_dir: /home/developer/.local/share/proj
PROJ DATA (recommended version): 1.13
PROJ Database: 1.2
EPSG Database: v10.082 [2023-02-06]
ESRI Database: ArcGIS Pro 3.1 [2023-19-01]
IGNF Database: 3.1.0 [2019-05-24]

System:
    python: 3.8.10 (default, Mar 13 2023, 10:26:41)  [GCC 9.4.0]
executable: /usr/bin/python3
   machine: Linux-5.15.90.1-microsoft-standard-WSL2-x86_64-with-glibc2.29

Python deps:
   certifi: 2022.6.15
    Cython: 0.29.14
setuptools: 58.4.0
       pip: 23.1

#### Installation method
poetry inside docker

What do you think?

snowman2:
- [x] Closes #1282
 - [x] Tests added

### DIFF
--- a/test/test_geod.py
+++ b/test/test_geod.py
@@ -5,7 +5,7 @@ from itertools import permutations
 
 import numpy
 import pytest
-from numpy.testing import assert_almost_equal
+from numpy.testing import assert_almost_equal, assert_array_equal
 
 from pyproj import Geod
 from pyproj.geod import GeodIntermediateFlag, reverse_azimuth
@@ -441,6 +441,21 @@ def test_geodesic_npts(include_initial, include_terminus):
     if include_terminus:
         expected_lonlats.append((_PORTLAND_LON, _PORTLAND_LAT))
     assert_almost_equal(lonlats, expected_lonlats, decimal=3)
+
+
+@pytest.mark.parametrize(
+    "input_data",
+    [
+        [1.0, 2.0, 3.0, math.nan],
+        [1.0, 2.0, math.nan, 4.0],
+        [1.0, math.nan, 3.0, 4.0],
+        [math.nan, 2.0, 3.0, 4.0],
+        [math.nan, math.nan, math.nan, math.nan],
+    ],
+)
+def test_geodesic_npts__nan(input_data):
+    geod = Geod(ellps="WGS84")
+    assert_array_equal(geod.npts(*input_data, npts=1), [(math.nan, math.nan)])
 
 
 @pytest.mark.parametrize(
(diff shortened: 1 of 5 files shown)


test/test_geod.py
```diff
@@ -443,6 +443,17 @@ def test_geodesic_npts(include_initial, include_terminus):
     assert_almost_equal(lonlats, expected_lonlats, decimal=3)
 
 
+def test_geodesic_npts__nan():
```
FactorizeD, line 446: I am wondering whether it wouldn't be nice to safeguard against all nan positions? You could use parametrize for that.

One way would be to do the following:

```
@pytest.mark.parametrize(lon_one, [1.0, np.nan])
@pytest.mark.parametrize(lat_one, [2.0, np.nan])
@pytest.mark.parametrize(lon_two, [3.0, np.nan])
@pytest.mark.parametrize(lat_two, [4.0, np.nan])
def test_geodesic_npts__nan(lon_one, lat_one, lon_two, lat_two):
    coords_list = [lon_one, lat_one, lon_two, lat_two]
    if np.isfinite(coords_list).all():
        return

    geod = Geod(ellps="WGS84")
    assert_array_equal(
        geod.npts(*coords_list, npts=1), [(math.nan, math.nan)]
    )
```

Alternatively, to omit the return statement:
```
from itertools import combinations

@pytest.mark.parametrize(nan_positions, list(combinations([0, 1, 2, 3], 1)) + list(combinations([0, 1, 2, 3], 2)) + list(combinations([0, 1, 2, 3], 3)) + list(combinations([0, 1, 2, 3], 4)))
def test_geodesic_npts__nan(nan_positions):
    geod = Geod(ellps="WGS84")
    coords_list = [1.0, 2.0, 3.0, 4.0]
    for position in nan_positions:
        coords_list[position] = math.nan

    assert_array_equal(
        geod.npts(*coords_list, npts=1), [(math.nan, math.nan)]
    )
```

snowman2: That sounds like a good idea.

snowman2: I went with a simpler approach that I think covers it.

PR #1295, opened 2023-05-28
DOC: Add a reference to #1206 in history

idanmiara:
I think that this is missing from: https://github.com/pyproj4/pyproj/pull/1206

 - [X] Fully documented, including `history.rst` for all changes and `api/*.rst` for new API


snowman2:
Thanks @idanmiara :+1:

Issue #1342, opened 2023-09-13
Request: release a patch version that pins cython<3.0


gwerbin:
Currently, it's impossible to build Pyproj 3.6.0 with PEP 517 build isolation because Pip will install Cython 3.x, which Pyproj 3.6.0 is incompatible with.

This can be easily fixed by releasing a Pyproj 3.6.1 or 3.6.0.post1 version that does simply adds `<3.0` to the Cython build dep: https://github.com/pyproj4/pyproj/compare/3.6.0...gwerbin:pin-cython-below-3.0?expand=1

I understand that the current master branch updates Pyproj to use _only_ Cython 3.x, but I'm not aware of a release date for it. This patch would be very useful for end users who need to build their own wheels (e.g. for PyPy), while the next release is still in development.

snowman2:
For most projects, a release is fairly simple as you mention. However, due to a requirement on maintainers free time, building wheels, and testing with downstream Linux release managers, it is not quite as simple for pyproj.

The next release of pyproj depends on python 3.12 compatibility updates from a release from numpy to ensure stability.

We welcome assistance preparing pyproj for the next release with Python 3.12 wheels.

gwerbin:
@snowman2 is it a matter of someone running through the release instructions [here](https://github.com/pyproj4/pyproj/blob/main/HOW_TO_RELEASE.md)? Or is there more to it? I'm happy to dedicate some time to the patch release, unless the next real release of Pyproj is right around the corner.

snowman2:
The blocker for the next release is #1330. Once that is ready, the release will follow. That likely depends on the next release of numpy, but there may be a way to get it to work now.

The instructions you linked to are correct. However, slightly out of date. The wheels are automatically uploaded except from Cirrus CI as those need to be manually uploaded to pypi currently.

djhoese:
@snowman2 I had some luck with the vispy project (builds wheels with a single Cython extension) with forcing the pre-release of numpy to be installed. I see you're already doing that, but it looks like for some reason it is trying to build numpy from source. Here's vispys cibuildwheel env vars:

https://github.com/vispy/vispy/blob/a1a639f33c59a16c8af2bf605a23b55210569f5e/.github/workflows/wheels.yml#L38-L39

If needed I can try to take a look tomorrow, but no guarantees on time. I also don't build 32-bit wheels for vispy (or any of my packages) and it looks like you do...how do you handle that with numpy not providing 32-bit wheels?

[...]

PR #1345, opened 2023-09-22
Cache PROJ library build when building wheels

djhoese:
As discussed in #1342 PROJ building takes about 10 minutes at the top of every wheel build workflow. This PR caches it with GitHub Actions to hopefully remove this requirement except for when PROJ (or its dependencies) are updated.

This workflow is taken from shapely's own CI.

 - [ ] Closes #xxxx
 - [ ] Tests added

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -116,6 +116,12 @@ jobs:
           mkdir -p ${GITHUB_WORKSPACE}/pyproj/proj_dir/share/proj
           cp "$VCPKG_INSTALLATION_ROOT/installed/${{ matrix.triplet }}/share/proj/"* ${GITHUB_WORKSPACE}/pyproj/proj_dir/share/proj/
 
+      - name: Cache PROJ build
+        uses: actions/cache@v3
+        with:
+          path: ${{ runner.temp }}/proj-${{ env.PROJ_VERSION }}
+          key: ${{ matrix.os }}-${{ matrix.arch }}-${{ env.PROJ_VERSION }}-${{ hashFiles('ci/proj-compile-wheels.sh') }}
+
       - name: Build wheels
         uses: pypa/cibuildwheel@v2.16
         env:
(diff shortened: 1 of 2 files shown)


djhoese:
I swear I write enough bash to not have made the mistakes I made in this PR. I don't think my coffee kicked in yet.

djhoese:
Damn I was so sure I had this working. It even restored the directory...but then it didn't use it.

djhoese:
Not sure why but throughout my testing the PROJ building has been improving in time from ~600s to ~480s. Maybe some download caching inside github? Not really sure.

Watching the current builds it seems like the cache is working.

djhoese:
...and nevermind it needs libtiff and sqlite too. I was only caching the PROJ directory.

djhoese:
@snowman2 FYI this works now, but has a lot of debug and ugly ways of doing things just as a proof of concept.

I'm very surprised at the variance in the build times both of PROJ itself and the pyproj wheel building. They seems to very by quite a bit. I've seen PROJ go between ~4 minutes to ~6 minutes through my testing.

One of the biggest surprises but one that still makes sense was that caching the libtiff and other PROJ deps in the cache directory meant I had to put it in LD_LIBRARY_PATH, but only on the linux system. Otherwise, `auditwheel` was failing to find them. One solution I thought of was to add the cache directory to `LD_RUN_PATH` versus depending on `LD_LIBRARY_PATH` so that I didn't need to carry aroundthis extra library path (it would be in the PROJ `.so`). My understanding is that auditwheel should bundle all the depended on libraries anyway so this shouldn't really matter. Does any of this concern you? Or do you have other things you'd like to see as I work on cleaning this code up?

snowman2:
Thanks @djhoese 👍 - I am looking forward to this update. I am busy this week, but should have a moment next week to take a look.

snowman2 (approved):

snowman2:
No concerns from my end. This change makes sense to me. Thank you for working on this :+1:

djhoese:
I just wish it was a bigger win for build times than it is.

PR #1354, opened 2023-10-26
PERF: drop GIL during long-running proj database calls

tpwrules:
These calls can take hundreds of milliseconds to access the PROJ databases, during which time no other Python code in any other thread can run. Use `nogil` to avoid this effect.

This should be safe for these calls as they all take a separate PROJ threading context.

It's quite probable there are other calls which could use this, but these negatively affected my application and I was able to identify them. I would welcome additions to the set of calls to wrap.

 - [ ] Closes #xxxx
 - [ ] Tests added

### DIFF
--- a/pyproj/_transformer.pyx
+++ b/pyproj/_transformer.pyx
@@ -330,14 +330,15 @@ cdef PJ* proj_create_crs_to_crs(
         ELSE:
             raise NotImplementedError("only_best requires PROJ 9.2+.")
 
-
-    cdef PJ* transform = proj_create_crs_to_crs_from_pj(
-        ctx,
-        source_crs,
-        target_crs,
-        area,
-        options,
-    )
+    cdef PJ* transform = NULL
+    with nogil:
+        transform = proj_create_crs_to_crs_from_pj(
+            ctx,
+            source_crs,
+            target_crs,
+            area,
+            options,
+        )
     proj_destroy(source_crs)
     proj_destroy(target_crs)
     if transform == NULL:
--- a/pyproj/database.pyx
+++ b/pyproj/database.pyx
@@ -69,7 +69,9 @@ def get_authorities():
         Authorities in PROJ database.
     """
     cdef PJ_CONTEXT* context = pyproj_context_create()
-    cdef PROJ_STRING_LIST proj_auth_list = proj_get_authorities_from_database(context)
+    cdef PROJ_STRING_LIST proj_auth_list = NULL
+    with nogil:
+        proj_auth_list = proj_get_authorities_from_database(context)
     if proj_auth_list == NULL:
         pyproj_context_destroy(context)
         return []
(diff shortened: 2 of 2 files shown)


pyproj/_transformer.pyx
```diff
@@ -330,14 +330,15 @@ cdef PJ* proj_create_crs_to_crs(
...
-    cdef PJ* transform = proj_create_crs_to_crs_from_pj(
-        ctx,
-        source_crs,
-        target_crs,
-        area,
-        options,
-    )
+    cdef PJ* transform = NULL
+    with nogil:
+        transform = proj_create_crs_to_crs_from_pj(
```
djhoese, line 335: Just for my own edification, does the declaration need to be moved outside the `with nogil` because Cython doesn't like it?

djhoese (approved):
I don't know enough about the thread-safety of PROJ, but I like the idea of this optimization. @tpwrules do you have any profiling results (pretty plots?) that show the execution time improvements you've seen by doing this?

Of course, something this low-level should really get approval from @snowman2 before merging.

pyproj/_transformer.pyx
tpwrules: Precisely. cdefs are only allowed directly within a function and not within a `with` or `try` block.

tpwrules:
It doesn't improve execution time, just makes other threads in the process more stable. It's a really hard thing to benchmark and measure unfortunately. I just was able to observe the effect on my program, but I can't really share it here.

jjimenezshaw:
> I don't know enough about the thread-safety of PROJ

PROJ does not deal with thread safety internally. It should be managed by the user via the "context". The same context must not be used simultaneously by two different threads.
(PROJ uses the "context" to keep there caches and any global initialization. So new context, new initialization. The access to the sqlite db -proj.db- is in read only mode, and managed by sqlite3 to avoid collisions)

pyproj/database.pyx
```diff
@@ -400,13 +408,15 @@ def get_units_map(str auth_name=None, str category=None, bint allow_deprecated=F
...
-        bool(allow_deprecated),
-        &num_units,
-    )
+    cdef PROJ_UNIT_INFO** db_unit_list = NULL
+    with nogil:
+        db_unit_list = proj_get_units_from_database(
+            context,
+            c_auth_name,
+            c_category,
+            bool(allow_deprecated),
```
snowman2, line 417: I am surprised cython didn't complain about using `bool` with `nogil`.

snowman2:
Thanks @tpwrules 👍

snowman2:
I would be interested to see if doing this with `proj_create` calls helps as well.

pyproj/database.pyx
tpwrules: I'm assuming it's valid if it did not complain, but I'm not 100% sure.

Issue #1111, opened 2022-07-24
DEP: Python 3.8 Support


snowman2:
Related to #930

Thinking about pyproj 3.4.x being the last series to support python 3.8.

Community references:

-  Python 3.11 - October 2022 [ref](https://devguide.python.org/versions/). #1110
-  [NEP-29](https://numpy.org/neps/nep-0029-deprecation_policy.html) says numpy drops support for Python 3.8 in April 2023.
- [rioxarray](https://github.com/corteva/rioxarray/issues/582)
- [rasterio](https://github.com/rasterio/rasterio/issues/2773)
- [geocube](https://github.com/corteva/geocube/issues/136)
- [xarray](https://github.com/pydata/xarray/pull/7461)

EpicWink:
Note Python 3.8 is the default Python for Ubuntu 20.04 (Focal), supported by Debian until May 2025. That's quite extended, so that's not my recommendation.

[PEP 569](https://peps.python.org/pep-0569/) says the last security release will be at least October 2024.

My suggestion is for Python 3.8 to be dropped when addressing failing tests in the Python 3.8 CI job becomes too much of a burden, or October 2024, whichever is later.

snowman2:
Stable versions of pyproj for Python 3.8 will continue to be on pypi for those users. Newer releases will have pins in the metadata to ensure disruption does not occur for those using the default python on LTS distributions.

If those users would like to use newer versions of packages in the Python ecosystem, the dead snakes ppa is a great resource: https://launchpad.net/~deadsnakes/+archive/ubuntu/ppa. pyproj uses it to test pyproj.

snowman2:
Another thing to consider is that Ubuntu and Python are funded by organizations that provide long term support. pyproj is maintained solely by contributors in their free time. So, the standard for support is a bit different.

PR #1357, opened 2023-11-08
DEP: Minimum supported Python version 3.10

snowman2:
Python 3.9 is no longer supported in the scientific python ecosystem: https://scientific-python.org/specs/spec-0000/

Related #1111

### DIFF
--- a/.github/workflows/build_docs.yaml
+++ b/.github/workflows/build_docs.yaml
@@ -23,7 +23,7 @@ jobs:
       - name: Setup Conda
         uses: s-weigand/setup-conda@v1
         with:
-          python-version: 3.9
+          python-version: '3.11'
           conda-channels: conda-forge
 
       - name: Install and Build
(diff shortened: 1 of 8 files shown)


djhoese:
Ah so pyproj follows scipy instead of numpy which doesn't drop 3.9 until April:

https://numpy.org/neps/nep-0029-deprecation_policy.html

djhoese:
Ah sorry, numpy drops 3.8 in April if I'm reading that right.

snowman2:
> Ah so pyproj follows scipy instead of numpy which doesn't drop 3.9 until April:
> 
> https://numpy.org/neps/nep-0029-deprecation_policy.html


The note at the top: "This NEP is superseded by the scientific python ecosystem coordination guideline [SPEC 0 — Minimum Supported Versions](https://scientific-python.org/specs/spec-0000/)."

jjimenezshaw:
Ubuntu LTS users (like me) will have only a few months to move to Ubuntu 24.04 (with python 3.12).
That applies as well for PROJ docker image.

snowman2:
> Ubuntu LTS users (like me) will have only a few months to move to Ubuntu 24.04 (with python 3.12). That applies as well for PROJ docker image.

The dead snakes ppa is a great resource: https://launchpad.net/~deadsnakes/+archive/ubuntu/ppa. pyproj uses it to test pyproj on Ubuntu with newer versions of Python in GitHub actions.

PR #1390, opened 2024-03-31
Add query_geodetic_crs_from_datum

### LINKED ISSUE #1389: Add proj_query_geodetic_crs_from_datum
In the scope of adding a new feature to https://spatialreference.org/ `base crs`, in a similar way it is done in [epsg.org](http://epsg.org/), I think I need [proj_query_geodetic_crs_from_datum](https://proj.org/en/9.4/development/reference/functions.html#c.proj_query_geodetic_crs_from_datum). (Please, correct me if I am wrong). 

The idea will be to get the geographic 3D from a geographic 2D, or a geocentric from a geographic 3D as `base crs`.

Something like EPSG:6318 -> EPSG:6319 -> EPSG:6317 

(For a projected and compound there are already tools in pyproj)

Looking at proj documentation, I think it is doable with `proj_query_geodetic_crs_from_datum`. It is the only missing piece.

Should I implement that function? Where in the code?

jjimenezshaw:
Adds new API to PROJ. `database.query_geodetic_crs_from_datum`

 - [x] Closes #1389
 - [x] Tests added

### DIFF
--- a/pyproj/database.pyx
+++ b/pyproj/database.pyx
@@ -514,6 +514,9 @@ def query_geodetic_crs_from_datum(
     list[CRS]
     """
 
+    if pj_type is not None and not isinstance(pj_type, PJType):
+        pj_type = PJType.create(pj_type)
+
     cdef const char* c_crs_type = NULL
     if pj_type is None:
         pass
--- a/pyproj/database.pyx
+++ b/pyproj/database.pyx
@@ -6,9 +6,14 @@ from collections import namedtuple
 from libc.stdlib cimport free, malloc
 
 from pyproj._compat cimport cstrdecode, cstrencode
-from pyproj._datadir cimport pyproj_context_create, pyproj_context_destroy
+from pyproj._datadir cimport (
+    _clear_proj_error,
+    pyproj_context_create,
+    pyproj_context_destroy,
+)
 
 from pyproj.aoi import AreaOfUse
+from pyproj.crs import CRS
 from pyproj.enums import PJType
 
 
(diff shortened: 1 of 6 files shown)


pyproj/database.pyx
```diff
@@ -483,3 +488,105 @@ def get_database_metadata(str key not None):
...
+    cdef const char* c_crs_auth_name = NULL
+    cdef const char* c_datum_auth_name = NULL
+    cdef const char* c_datum_code = NULL
+    cdef bytes b_crs_auth_name
+    cdef bytes b_datum_auth_name
+    cdef bytes b_datum_code
+
+    if crs_auth_name is not None:
+        b_crs_auth_name = cstrencode(crs_auth_name)
+        c_crs_auth_name = b_crs_auth_name
```
jjimenezshaw, line 550: There is no simpler and more compact way to do this?

pyproj/database.pyx
```diff
@@ -483,3 +488,105 @@ def get_database_metadata(str key not None):
...
+    list[_CRS]
+    """
+
+    cdef const char* c_crs_type = NULL
+    cdef bytes b_crs_type
+    cdef str crs_type = ""
+    if pj_type is None:
+        pass
+    elif pj_type == PJType.GEOCENTRIC_CRS:
+        crs_type = "geocentric"
```
snowman2, line 529: Since these are static (no user input), you can use this strategy to set the string:
```suggestion
        c_crs_type = b"geocentric"
```

pyproj/database.pyx
```diff
@@ -483,3 +488,105 @@ def get_database_metadata(str key not None):
...
+
+    Returns
+    -------
+    list[_CRS]
+    """
+
+    cdef const char* c_crs_type = NULL
+    cdef bytes b_crs_type
+    cdef str crs_type = ""
+    if pj_type is None:
```
snowman2, line 526:
```python
 cdef PJ_TYPE cpj_type = get_pj_type(pj_type)
```

pyproj/database.pyx
```diff
@@ -483,3 +488,105 @@ def get_database_metadata(str key not None):
...
+        pass
+    elif pj_type == PJType.GEOCENTRIC_CRS:
```
snowman2, line 528:
```suggestion
    elif pj_type is PJType.GEOCENTRIC_CRS:
```

pyproj/database.pyx
```diff
@@ -483,3 +488,105 @@ def get_database_metadata(str key not None):
...
+
+    Parameters
+    ----------
+    crs_auth_name: str | None
+        The authority name to filter by (e.g. EPSG, ESRI). None is all.
+    datum_auth_name: str
+        The authority of the datum
+    datum_code: str
+        Datum code
+    pj_type: pyproj.enums.PJType | None, optional
```
snowman2, line 514:
```suggestion
    crs_type: pyproj.enums.PJType | None, optional
```

pyproj/database.pyx
```diff
@@ -483,3 +488,105 @@ def get_database_metadata(str key not None):
...
+    )
+
+    if proj_list != NULL:
+        num_proj_objects = proj_list_get_count(proj_list)
+
+    cdef PJ* proj = NULL
+    try:
+        for iii in range(num_proj_objects):
+            proj = proj_list_get(context, proj_list, iii)
+            ret_list.append(_CRS(proj_as_wkt(context, proj, PJ_WKT2_2019, NULL)))
```
snowman2, line 581: The `_CRS` class is not in the public API. Probably better to return `CRS`.

pyproj/database.pyx
jjimenezshaw: I do not understand this suggestion. If the user sets `None`, then c_crs_type is NULL, and the C function will return without filtering by type.

pyproj/database.pyx
jjimenezshaw: thanks. done

pyproj/database.pyx
jjimenezshaw: done. Is there any difference?

pyproj/database.pyx
jjimenezshaw: Sure. Done.

pyproj/database.pyx
snowman2: The suggestiin is not in the exact location. It would be applied after that.

pyproj/database.pyx
snowman2: It is a minor difference that cython uses to be slightly more performant.

pyproj/database.pyx
jjimenezshaw: I think I understood what you mean. I added it, and some tests.

snowman2:
I tried to re-produce the error and couldn't with https://github.com/pyproj4/pyproj/compare/main...snowman2:pyproj:docker .

To debug, I suggest modifying the CI tests to specifically run the test that fails and trying these configurations:
- set [PROJ_DEBUG](https://proj.org/en/9.3/usage/environmentvars.html#envvar-PROJ_DEBUG) environment variable to 3
- add `-vv` flag to tests
- Install `gdb` and run `gdb -ex=r --args /venv/bin/python -B -m pytest test/test_database.py`

snowman2:
Also, it may not hurt to rebase this PR against the latest version.

jjimenezshaw:
> Also, it may not hurt to rebase this PR against the latest version.

I did rebase (after some problems), but the tests are not running. Is anything broken in the gh actions?

snowman2:
I was able to re-produce this with PROJ 9.3  (both docker & conda envs) and I did not encounter any errors with PROJ 9.4 on the docker image (https://github.com/pyproj4/pyproj/pull/1390#issuecomment-2065691985):
```
test/test_database.py 
Thread 1 "python" received signal SIGSEGV, Segmentation fault.
0x000075ff9a5b3079 in proj_list_get_count () from /lib/x86_64-linux-gnu/libproj.so.25
```
If we add logic in the function to not allow running in PROJ 9.3, I would be okay with adding this in.

test/test_database.py
```diff
@@ -268,3 +269,50 @@ def test_get_database_metadata():
 
 def test_get_database_metadata__invalid():
     assert get_database_metadata("doesnotexist") is None
+
+
+def test_query_geodetic_crs_from_datum():
```
snowman2, line 274: `pytest.mark.parametrize` could help simplify this test.

test/test_database.py
```diff
@@ -268,3 +269,50 @@ def test_get_database_metadata():
...
+
+    crss = query_geodetic_crs_from_datum("EPSG", "EPSG", "6269", None)
+    assert len(crss) == 1
+    assert crss[0].to_authority()[1] == "4269"
+
+    crss = query_geodetic_crs_from_datum(None, "EPSG", "6269")
+    assert len(crss) == 3  # EPSG, ESRI, OGC
+
+
+def test_query_geodetic_crs_from_datum_invalid():
```
snowman2, line 298: `pytest.mark.parametrize` could help simplify this test.

test/test_database.py
```diff
@@ -268,3 +269,50 @@ def test_get_database_metadata():
...
+    crss = query_geodetic_crs_from_datum(None, "EPSG", "11")
+    assert len(crss) == 0
+
+    crss = query_geodetic_crs_from_datum(None, "EPSG", "32632")
+    assert len(crss) == 0
+
+    crss = query_geodetic_crs_from_datum("foo-bar", "EPSG", "6269", None)
+    assert len(crss) == 0
+
+    with pytest.raises(ValueError):
```
snowman2, line 308: I recommend creating a separate test where `ValueError` is raised and use `pytest.mark.parametrize` to help simplify this test.

jjimenezshaw:
> If logic is added in the function to not allow running in PROJ 9.3, I would be okay with adding this in.

How can I do that? Is it done somewhere else in the code?

snowman2:
> > If logic is added in the function to not allow running in PROJ 9.3, I would be okay with adding this in.
> 
> How can I do that? Is it done somewhere else in the code?

This is no longer needed with https://github.com/pyproj4/pyproj/pull/1481

jjimenezshaw:
I had to change `.pre-commit-config.yaml`, otherwise I cannot do a commit.
I have an Ubuntu 24.04, and installed the conda environment as explained in https://github.com/jjimenezshaw/pyproj/blob/main/CONTRIBUTING.md

The error message is this:
```
An unexpected error has occurred: CalledProcessError: command: ('/home/jshaw/anaconda3/envs/pyproj-devel/bin/python', '-mvirtualenv', '/home/jshaw/.cache/pre-commit/repoc6h5_c3j/py_env-python3.11', '-p', 'python3.11')
return code: 1
stdout:
    RuntimeError: failed to find interpreter for Builtin discover of python_spec='python3.11'
stderr: (none)
Check the log at /home/jshaw/.cache/pre-commit/pre-commit.log
```

The installed python in the environment is 
``` bash
$ python3 --version
Python 3.13.7
```

snowman2:
Thanks @jjimenezshaw 👍

PR #1396, opened 2024-04-20
Fix tests.yaml for CI

jjimenezshaw:
As mentioned by @snowman2 in https://github.com/pyproj4/pyproj/issues/1390#issuecomment-2067054089


snowman2:
Thank you 👍

Issue #1316, opened 2023-06-30
CI: Nightly wheels?


snowman2:
https://scientific-python.org/specs/spec-0004/

snowman2:
https://anaconda.org/scientific-python-nightly-wheels/pyproj

PR #1419, opened 2024-07-22
PERF: thread local context

### LINKED ISSUE #1133: PERF: Thread-local global contex

### LINKED ISSUE #1418: Small memory leak
#### Code Sample, a copy-pastable example if possible

```python
import pyproj # yes this simple
```
Run this code in heaptrack, and notice that likely osgeo::proj::io::DatabaseContext::create should be freed.

```sh
heaptrack python -c "import pyproj"
```

#### Problem description

I am currently evaluating performance issues of a python program with respect to memory leaks. I have noticed that, while little, pyproj also has a contribution to it. It is about 2.6MB, small.

#### Expected Output

No memory allocation expected at the end of the program. But one may also wonder why the DatabaseContext is already created without _any_ calls. I think it may be more elegant to wait for the first call that requires it.

#### Environment Information
3.6.1 / 9.4.0 / 3.12.4 / Linux-6.9.9-arch1-1-x86_64-with-glibc2.39

#### Installation method
pacman

snowman2:
- [x] Closes #1133
 - [x] Closes #1418 
 - [x] Tests updated


snowman2:
Considering merging this for this release. I haven't seen any issues so far. If you have time to test it with other repos that use pyproj, that would be helpful.

snowman2:
#1316 may be helpful for testing...

snowman2:
https://anaconda.org/scientific-python-nightly-wheels/pyproj

PR #1441, opened 2024-09-29
MNT: Add too-many-positional-arguments to lint ignore rules

greglucas:
pylint 3.3.0 added a new rule that is causing failures. There is already an ignore rule for too many arguments, so also add the ignore rule for the positional specific arguments.


snowman2:
Thanks @greglucas 👍

PR #1442, opened 2024-09-29
CI: Pin micromamba due to CI Runner incompatibilities with newer versions

greglucas:
Looks like micromamba v2 release doesn't work on GitHub Actions runners currently, so pin the version back.

xref: https://github.com/mamba-org/setup-micromamba/issues/225


snowman2:
Thanks @greglucas 👍

PR #1450, opened 2024-10-09
MNT: Remove use of utcnow() and change to naive datetimes instead.

greglucas:
`utcnow()` is deprecated since Python 3.12 in preference of `now(UTC)`. However, here `st_time` actually returns a naive datetime, so doing `now(UTC)` returns a TypeError because you can't subtract a naive time from a timezone aware time. So, I think we should actually be using whatever the system is reporting for the timezone with just `now()`.

https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow


snowman2:
Thanks @greglucas 👍

PR #1452, opened 2024-10-09
Cython 3.1+ fixes

### LINKED ISSUE #1451: Upcoming Cython 3.1 str <-> bytes errors
Building with the latest Cython nightly wheel causes errors with strings that are expected to be bytes and vice-versa. I'm not sure if this is due to an implicit cast that was removed somewhere or something else. My guess is this may be related to: https://github.com/cython/cython/pull/6374

```bash
pip install --upgrade --pre --only-binary :all: -i https://pypi.anaconda.org/scientific-python-nightly-wheels/simple cython numpy
pip install .
python -m pytest
```

#### Problem description

```
____________________________________________________ ERROR collecting test/crs/test_crs.py ____________________________________________________
test/crs/test_crs.py:1670: in <module>
    CRS("EPSG:4979"),  # native 3D
pyproj/crs/crs.py:350: in __init__
    self._local.crs = _CRS(self.srs)
pyproj/_crs.pyx:2355: in pyproj._crs._CRS.__init__
    def __init__(self, const char *proj_string):
E   TypeError: expected bytes, str found
__________________________________________________ ERROR collecting test/test_datum_shift.py __________________________________________________
test/test_datum_shift.py:30: in <module>
    WGS84_PROJ = Proj(proj="latlong", datum="WGS84")
pyproj/proj.py:110: in __init__
    self.crs = CRS.from_user_input(projparams, **kwargs)
pyproj/crs/crs.py:503: in from_user_input
    return cls(value, **kwargs)
pyproj/crs/crs.py:350: in __init__
    self._local.crs = _CRS(self.srs)
pyproj/_crs.pyx:2355: in pyproj._crs._CRS.__init__
    def __init__(self, const char *proj_string):
E   TypeError: expected bytes, str found
```

#### Expected Output

No TypeErrors

#### Installation method

from source

greglucas:
This fixes the errors that came up during Cython3.1a testing. I'm not a Cython expert by any means, so let me know if there is a better way to handle any of these. Do you want to add a test/workflow for upcoming Cython releases?

Test with this nightly wheel release install locally:
```
pip install --upgrade --pre --only-binary :all: -i https://pypi.anaconda.org/scientific-python-nightly-wheels/simple cython numpy
pip install .
python -m pytest
```

 - [x] Closes #1451

### DIFF
--- a/pyproj/_geod.pyx
+++ b/pyproj/_geod.pyx
@@ -90,10 +90,7 @@ cdef class Geod:
         geod_init(&self._geod_geodesic, <double> a, <double> f)
         self.a = a
         self.f = f
-        # convert 'a' only for initstring
-        a_str = int(a) if a.is_integer() else a
-        f_str = int(f) if f.is_integer() else f
-        self.initstring = f"+a={a_str} +f={f_str}"
+        self.initstring = f"+{a=} +{f=}"
         self.sphere = sphere
         self.b = b
         self.es = es
--- a/.github/workflows/test_proj_latest.yaml
+++ b/.github/workflows/test_proj_latest.yaml
@@ -1,4 +1,4 @@
-name: Test PROJ Latest
+name: Test PROJ and Cython Latest
 
 on:
   push:
@@ -41,7 +41,7 @@ jobs:
         shell: bash
         run: |
           python -V
-          python -m pip install cython
+          python -m pip install --upgrade --pre --only-binary :all: -i https://pypi.anaconda.org/scientific-python-nightly-wheels/simple cython
           python -m pip install -e .
           python -m pip install -r requirements-test.txt
           pyproj -v
--- a/pyproj/_crs.pyx
+++ b/pyproj/_crs.pyx
@@ -2352,13 +2352,13 @@ cdef class _CRS(Base):
         self._coordinate_operation = None
         self._type_name = None
 
-    def __init__(self, const char *proj_string):
+    def __init__(self, str proj_string):
         self.context = pyproj_context_create()
         self._context_manager = get_context_manager()
         # initialize projection
         self.projobj = proj_create(
             self.context,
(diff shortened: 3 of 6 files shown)


pyproj/_geod.pyx
```diff
@@ -90,10 +90,7 @@ cdef class Geod:
         geod_init(&self._geod_geodesic, <double> a, <double> f)
         self.a = a
         self.f = f
-        # convert 'a' only for initstring
-        a_str = int(a) if a.is_integer() else a
-        f_str = int(f) if f.is_integer() else f
-        self.initstring = f"+a={a_str} +f={f_str}"
+        self.initstring = f"+{a=} +{f=}"
```
greglucas, line 93: Note that I took the easy route and changed the repr to a pure float for `Geod`. Does this matter? The actual usage is as a double, so presumably this is actually "more correct" to show to users.

snowman2: This change predates me. Is it relevant to this PR?

snowman2:
> Do you want to add a test/workflow for upcoming Cython releases?

Yes, that change is welcome.

greglucas:
>> Do you want to add a test/workflow for upcoming Cython releases?

> Yes, that change is welcome.

I added it to the PROJ_latest workflow to just install the latest Cython there as well. Let me know if you want it as an isolated workflow instead.

pyproj/_geod.pyx
greglucas: Yes, there is a borrowed Python reference error with it so the project won't even compile. We could change this to something like setting a temporary variable and expanding the if/else.

snowman2: I think that your fix makes the most sense to me.

snowman2:
Thanks @greglucas 👍

PR #1455, opened 2024-10-25
Setup ruff

tqa236:
I divide the MR into 2 commits:
- The first one sets up `ruff` and its config
- The second one fixes all sensible auto-fix errors

If you want to check with rules that `ruff` auto-fix, please checkout to the first commit, change `fix = false` and run `pre-commit run ruff --all-files`

I make the current config based on my experience, please feel free to reclassify the rules to "won't fix" or "to do"

 - [ ] Closes #xxxx
 - [ ] Tests added


snowman2:
Thanks @tqa236 👍

PR #1456, opened 2024-10-31
Disallow type error without code

tqa236:
I strengthen the `mypy` and `ruff` config a bit to disallow `type: ignore` with no code, found one case when the ignore can be removed


snowman2:
Thanks @tqa236 👍

PR #1471, opened 2025-01-29
DEP: Make shapely a mandatory dependency

### LINKED ISSUE #1470: Should shapely be added as an (optional?) dependency?
Some of the pyproj functionality requires `shapely`. E.g. [`Geod.geometry_area_perimeter`](https://pyproj4.github.io/pyproj/stable/api/geod.html#pyproj.Geod.geometry_area_perimeter). Nevertheless when you `pip install pyproj`, `shapely` does not get installed. I understand that most of the time when you install pyproj you will also install `shapely` or something else that will pull it, nevertheless, from a metadata point of view, shouldn't it be added at least as an optional dependency?

pmav99:
- [x] Closes #1470
 - [x] Tests added

### DIFF
--- a/docs/history.rst
+++ b/docs/history.rst
@@ -4,6 +4,7 @@ Change Log
 Latest
 ------
 
+- DEP: Make shapely a mandatory dependency (pull #1471)
 
 3.7.0
 ------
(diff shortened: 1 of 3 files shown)


snowman2:
I think that there was some confusion. The change is to make shapely optional.

snowman2:
Closing due to inactivity. I apologize for the confusion. If you want to list shapely as an optional dependency, that would be okay,

PR #1490, opened 2025-04-25
DOC: add note for WKT and CRS interaction

### LINKED ISSUE #1486: DOC: WKT & coordinate systems not compatible
Should probably document this better, but WKT & coordinate systems don't mix well together: https://github.com/OSGeo/PROJ/issues/1819

I recommend using `to_json`/`from_json` instead.

_Originally posted by @snowman2 in https://github.com/pyproj4/pyproj/discussions/1485#discussioncomment-12671241_

myl-dvl:
- [x] Closes #1486
 
image

---
It seems that it added my previous commits, I'm not sure what I did wrong.
Does it change anything for the PR ?

### DIFF
--- a/pyproj/_crs.pyx
+++ b/pyproj/_crs.pyx
@@ -569,7 +569,10 @@ cdef class CoordinateSystem(_CRSParts):
     """
     .. versionadded:: 2.2.0
 
-    Coordinate System for CRS
+    Coordinate System for CRS.
+
+    .. note:: | WKT and CRS are not compatible as standalone element (see `OSGeo/PROJ#1819 <https://github.com/OSGeo/PROJ/issues/1819>`_ for more details).
+              | We recommend using `to_json`/`from_json` for the import.
 
     Attributes
     ----------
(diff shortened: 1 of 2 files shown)


snowman2:
Probably helpful to note here that WKT doesn't work on import for CoordinateSystem and PROJ JSON is preferred: https://github.com/pyproj4/pyproj/blob/1fa11649f7f0adeda5ec4ad1e778bbea7dd6bf0b/pyproj/_crs.pyx#L514

myl-dvl:
> Probably helpful to note here that WKT doesn't work on import for CoordinateSystem and PROJ JSON is preferred:
> 
> https://github.com/pyproj4/pyproj/blob/1fa11649f7f0adeda5ec4ad1e778bbea7dd6bf0b/pyproj/_crs.pyx#L514


---
Something like this ?

image

snowman2:
> Something like this ?

Yes, that would be great to add to `from_user_input`.

myl-dvl:
> > Something like this ?
> 
> Yes, that would be great to add to `from_user_input`.

I'm not sure if you want the note here or in this method ([link](https://github.com/pyproj4/pyproj/blob/1fa11649f7f0adeda5ec4ad1e778bbea7dd6bf0b/pyproj/_crs.pyx#L507)).
Could you elaborate please ?
I will undraft this PR, if by luck it works for you^^

snowman2:
The original issue #1485 had confusion as the user was using `from_user_input`. So, I think that it would be good to ensure that we add a note inside that method. You can also have a note where you have it as well as someone may see it there instead.

PR #1492, opened 2025-04-25
DOC: Remove 'noqa: E501' from docstrings

adamjstewart:
These comments are visible in the API docs: https://pyproj4.github.io/pyproj/stable/api/crs/crs.html

The correct way to handle long lines in a docstring is to put the ignore at the end of the docstring.


snowman2:
Thanks 👍

PR #1494, opened 2025-04-29
pyproj.crs: add imports to __all__

adamjstewart:
This PR explicitly exports all classes in `pyproj.crs`, allowing type checkers like mypy to better use the type hints. Without this change, mypy will complain with an error like:
```console
> mypy --strict foo.py
foo.py:3: error: Module "pyproj.crs" does not explicitly export attribute "CRS"  [attr-defined]
```

### DIFF
--- a/pyproj/crs/__init__.py
+++ b/pyproj/crs/__init__.py
@@ -4,7 +4,7 @@ to the coordinate reference system (CRS) information through the CRS
 class.
 """
 
-from pyproj._crs import (  # noqa: F401  pylint: disable=unused-import
+from pyproj._crs import (
     CoordinateOperation,
     CoordinateSystem,
     Datum,
@@ -13,7 +13,7 @@ from pyproj._crs import (  # noqa: F401  pylint: disable=unused-import
     is_proj,
     is_wkt,
 )
-from pyproj.crs.crs import (  # noqa: F401  pylint: disable=unused-import
+from pyproj.crs.crs import (
     CRS,
     BoundCRS,
     CompoundCRS,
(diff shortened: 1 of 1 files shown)


adamjstewart:
@snowman2 can you review?

pyproj/crs/__init__.py
```diff
@@ -24,4 +24,24 @@
...
+    "CRS",
+    "BoundCRS",
+    "CompoundCRS",
+    "CustomConstructorCRS",
+    "DerivedGeographicCRS",
+    "GeocentricCRS",
+    "GeographicCRS",
+    "ProjectedCRS",
+    "VerticalCRS",
+    "CRSError",
```
snowman2, line 46: This import is for backwards compatibility.
```suggestion
```

pyproj/crs/__init__.py
```diff
@@ -24,4 +24,24 @@
     ProjectedCRS,
     VerticalCRS,
 )
-from pyproj.exceptions import CRSError  # noqa: F401  pylint: disable=unused-import
+from pyproj.exceptions import CRSError
+
+__all__ = [
+    "CoordinateOperation",
```
snowman2, line 30: CoordinateOperation, Datum, Ellipsoid, PrimeMeridian and here for backwards compatibility. I don't recommend adding them to `__all__`.

snowman2:
Thanks @adamjstewart 👍

pyproj/crs/__init__.py
```diff
@@ -25,3 +25,18 @@
     VerticalCRS,
 )
 from pyproj.exceptions import CRSError  # noqa: F401  pylint: disable=unused-import
+
+__all__ = [
+    "CoordinateSystem",
```
snowman2, line 30:
```suggestion
```

PR #1496, opened 2025-05-22
MNT: replace black and isort with ruff format and lint

mwtoews:
This maintenance PR does the following:

- Replace Black with the Ruff formatter
- Replace isort with Ruff's "I" lint check
- Update the Ruff version in pre-commit, which changes a few source files
- Remove target-version from ruff configuration, since this is automatically evaluated
- Add ruff to requirements-dev.txt
- Update relevant excerpts of CONTRIBUTING.md with these changes

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,15 +8,6 @@ repos:
         -   id: check-yaml
         -   id: end-of-file-fixer
         -   id: trailing-whitespace
-    -   repo: https://github.com/psf/black-pre-commit-mirror
-        rev: 24.2.0
-        hooks:
-        -   id: black
-    -   repo: https://github.com/pycqa/isort
-        rev: 5.13.2
-        hooks:
-        -   id: isort
-            args: [setup.py, pyproj/, test/, docs/]
     -   repo: https://github.com/asottile/blacken-docs
         rev: 1.16.0
         hooks:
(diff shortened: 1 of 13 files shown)


.stickler.yml
```diff
@@ -1,6 +1,4 @@
 linters:
-  black:
-    config: ./pyproject.toml
```
snowman2, line 3: This file can be deleted.

.pre-commit-config.yaml
```diff
@@ -8,15 +8,6 @@ repos:
         -   id: check-yaml
         -   id: end-of-file-fixer
         -   id: trailing-whitespace
-    -   repo: https://github.com/psf/black-pre-commit-mirror
```
snowman2, line 11: I am okay removing isort. However, I am not sure about black. Do the formatters conflict?

mwtoews: The Ruff formatter is a drop-in replacement for Black, but much faster (100x). According to [these docs](https://docs.astral.sh/ruff/formatter/#ruff-format) "> 99.9% of lines are formatted identically", so this may explain some minor differences of formatting found in this PR.

snowman2: Sounds worth a try. Cand always swap it back later. Probably should remove the black badge in the readme.

snowman2:
Thanks @mwtoews 👍

PR #1512, opened 2025-08-05
Enable free-threading compatibility

djhoese:
Closes #1508

TODO: Look into adding free-threaded interpreter to tests

 - [ ] Closes #xxxx
 - [ ] Tests added
 - [ ] Fully documented, including `history.rst` for all changes and `api/*.rst` for new API

PR #1514, opened 2025-08-06
CI: Add 3.13 free-threading wheels

djhoese:
Continuation from #1512 

 - [ ] Closes #xxxx
 - [ ] Tests added

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -139,7 +139,6 @@ jobs:
       - name: Build wheels
         uses: pypa/cibuildwheel@v3.1.3
         env:
-          CIBW_SKIP: "pp*-win* pp31*"
           CIBW_ENABLE: cpython-freethreading
           CIBW_ARCHS: ${{ matrix.arch }}
           CIBW_ENVIRONMENT_LINUX:


djhoese:
@snowman2 PyPy is no longer built by default with cibuildwheel 3+. Do you still want those wheels built?

Edit: I see `CIBW_SKIP: "pp*-win* pp31*"` which a warning says are no longer supported. So I'm not sure what PyPy versions that leaves.

djhoese:
Alright reading the logs it looks like 3.13t wheels are being built now @alihamdan.

djhoese:
Regarding PyPy, cibuildwheel supports builds for PyPy with 3.8-3.11, but the selectors in CI were already dropping PyPy on Windows and all PyPy starting with `31*`. Kind of feels like PyPy shouldn't be built until someone comes along and needs it again.

snowman2:
> Kind of feels like PyPy shouldn't be built until someone comes along and needs it again.

That's fine with me. Mind making an issue about pypy support and pinning it?

djhoese:
Issue created but I can't pin it because there are already 3 pinned and github only allows 3.

snowman2:
> Issue created but I can't pin it because there are already 3 pinned and github only allows 3.

Much appreciated. I updated the pins.

PR #1539, opened 2025-10-23
BUG: Fix segfault when pyproj is called from threads which were not created by Python

### LINKED ISSUE #1509: Multi-threading with Qt leads to crash of Python process
When I use the multi-threading features of the Qt GUI framework (PySide6) and call Transformer.transform from pyproj inside the thread, the Python process crashes after a low number of executed threads. It always works fine the first time pyproj is used inside a thread.

Here is a minimum example using a QRunnable inside QThreadPool:
```python
from PySide6.QtCore import QRunnable, QThreadPool, QObject, Signal
from pyproj import Transformer

class WorkerSignals(QObject):
    finished = Signal(float, float)

class TransformTask(QRunnable):
    def __init__(self):
        super().__init__()
        self.signals = WorkerSignals()

    def run(self):
        lon, lat = 16.3738, 48.2082
        transformer = Transformer.from_crs("EPSG:4326", "EPSG:32633", always_xy=True)
        x, y = transformer.transform(lon, lat)
        self.signals.finished.emit(x, y)

def print_result(x, y):
    print(f"WGS84: (16.3738, 48.2082) -> UTM: ({x:.2f}, {y:.2f})")

for _ in range(100):
    task = TransformTask()
    task.signals.finished.connect(print_result)
    QThreadPool.globalInstance().start(task)
```
Interestingly, the crash does no occur when using the standard threading module:
```python
import threading
from pyproj import Transformer

def transform_coordinates():
    lon, lat = 16.3738, 48.2082
    transformer = Transformer.from_crs("EPSG:4326", "EPSG:32633", always_xy=True)
    x, y = transformer.transform(lon, lat)
    print(f"WGS84: ({lon}, {lat}) -> UTM: ({x:.2f}, {y:.2f})")

for _ in range(100):
    t = threading.Thread(target=transform_coordinates)
    t.start()
```
According to the Windows event report the crash originates from pyproj.libs\proj_9-ece80b3ce99919cf7dd31637c074a1bb.dll. The error code is 0xc0000005.

I am using PySide6 version 6.9.1. Here is the version info of my pyproj installation:
> pyproj info:
>     pyproj: 3.7.1
> PROJ (runtime): 9.5.1
> PROJ (compiled): 9.5.1
> PROJ DATA (recommended version): 1.20
> PROJ Database: 1.4
> EPSG Database: v11.022 [2024-11-05]
> ESRI Database: ArcGIS Pro 3.4 [2024-11-04]
> IGNF Database: 3.1.0 [2019-05-24]
> 
> System:
>     python: 3.13.5 | packaged by Anaconda, Inc. | (main, Jun 12 2025, 16:37:03) [MSC v.1929 64 bit (AMD64)]
>    machine: Windows-11-10.0.26100-SP0
> 
> Python deps:
>    certifi: 2025.7.14
>     Cython: None
> setuptools: 78.1.1
>        pip: 25.1

This issue did not occur using Python 3.11.9 with pyproj 3.6.1 and PySide6 6.7.2.

Any ideas?
Thanks!

dgburr:
The switch to thread-local contexts in https://github.com/pyproj4/pyproj/pull/1419 means that any QT-based apps using QThreads now segfault with pyproj >= 3.7.0 (see https://github.com/pyproj4/pyproj/issues/1509).  

This is due to the fact that threads created via QT are not created via the python thread API and therefore they do not have a python thread state (see https://docs.python.org/3/c-api/init.html#non-python-created-threads for more details).  

Therefore it is necessary to call `PyGILState_Ensure()` before calling any of the `PyThread_tss_*` functions and to release the `PyGILState_STATE` only once the pyproj context is destroyed.

 - [✅] Closes #1509 
 - [❌] Tests added

### DIFF
--- a/docs/history.rst
+++ b/docs/history.rst
@@ -6,6 +6,7 @@ Latest
 - WHL: Wheels contain PROJ 9.7.0 (pull #1529)
 - WHL: Remove MacOS 13 (X86_64) wheels (issue #1532)
 - ENH: Add :meth:`database.query_geodetic_crs_from_datum` (pull #1390)
+- BUG: Fix segfault when pyproj is called from threads which were not created by Python (issue #1509)
 
 3.7.2
 -----


djhoese:
I could be misunderstanding this (and that is very likely) but doesn't this essentially hold the GIL the entire time? Or at least introduce the GIL where it wasn't previously used?

snowman2:
> I could be misunderstanding this (and that is very likely) but doesn't this essentially hold the GIL the entire time? Or at least introduce the GIL where it wasn't previously used?

The [docs](https://docs.python.org/3/c-api/init.html#non-python-created-threads) say:

> When threads are created using the dedicated Python APIs (such as the [threading](https://docs.python.org/3/library/threading.html#module-threading) module), a thread state is automatically associated to them and the code showed above is therefore correct. However, when threads are created from C (for example by a third-party library with its own thread management), they don’t hold the [GIL](https://docs.python.org/3/glossary.html#term-GIL), because they don’t have an [attached thread state](https://docs.python.org/3/glossary.html#term-attached-thread-state).

> If you need to call Python code from these threads (often this will be part of a callback API provided by the aforementioned third-party library), you must first register these threads with the interpreter by creating an [attached thread state](https://docs.python.org/3/glossary.html#term-attached-thread-state) before you can start using the Python/C API. When you are done, you should detach the [thread state](https://docs.python.org/3/glossary.html#term-attached-thread-state), and finally free it.

Based on that, I believe this is necessary.

dgburr:
Hi @djhoese: I have done some digging into the python source and documentation and I believe that you are correct - this approach could result in one thread holding the GIL for an extended period of time.  Therefore I have come up with an alternate solution, see https://github.com/pyproj4/pyproj/pull/1540.

djhoese:
My main concern and it's possible this was never true in the first place, is that if I use pyproj from a dask-based setup where I have multiple threads processing separate numpy arrays, that they can all run simultaneously. If that requires me creating equivalent Transformer objects in each thread that is fine.

dgburr:
Hi @djhoese: I have tracked down the real root cause of the segfault, see https://github.com/pyproj4/pyproj/pull/1541

PR #1540, opened 2025-10-23
Restore option to use global pyproj context

dgburr:
Restore the `set_use_global_context()` functionality (removed in https://github.com/pyproj4/pyproj/pull/1419) so that a single PJ_CONTEXT can be used for all calls to PROJ.  This can be used as a workaround for the issue described in #1509, although the proper fix is #1541.

 - [❌] Tests added


snowman2:
This option is unsafe with multi-threading.

PR #1541, opened 2025-10-23
BUG: Clear CONTEXT_THREAD_KEY when destroying the PJ_CONTEXT

dgburr:
The switch to thread-local contexts in #1419 stores a pointer to the newly created `PJ_CONTEXT` using Thread Specific Storage.  However the value stored in TSS is not cleared when the `PJ_CONTEXT` is destroyed, meaning that a subsequent call to `pyproj_context_create()` will return a pointer to an invalid `PJCONTEXT`, leading to a segmentation fault.

My specific use case is in a multithreaded QT app with symptoms similar to https://github.com/pyproj4/pyproj/issues/1509.  However this issue can occur in any multithreaded app which performs multiple calls to `pyproj_context_create()` from the same thread.

 - [❌] Tests added

### DIFF
--- a/docs/history.rst
+++ b/docs/history.rst
@@ -6,7 +6,7 @@ Latest
 - WHL: Wheels contain PROJ 9.7.0 (pull #1529)
 - WHL: Remove MacOS 13 (X86_64) wheels (issue #1532)
 - ENH: Add :meth:`database.query_geodetic_crs_from_datum` (pull #1390)
-- BUG: Clear CONTEXT_THREAD_KEY when destroying the PJ_CONTEXT (pull #1541)
+- BUG: Clear CONTEXT_THREAD_KEY when destroying PJ_CONTEXT (pull #1541)
 
 3.7.2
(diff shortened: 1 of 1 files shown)


pyproj/_context.pyx
```diff
@@ -177,6 +177,7 @@ cdef class ContextManager:
 
     def __dealloc__(self):
         if self.context != NULL:
+            PyThread_tss_set(&CONTEXT_THREAD_KEY, NULL)
             proj_context_destroy(self.context)
```
snowman2: To be safe, add self.context = NULL here too.

dgburr:
@snowman2: should I also try to fix the linting issue in this PR, or should I create another one?

snowman2:
> @snowman2: should I also try to fix the linting issue in this PR, or should I create another one?

A separate PR would be perfect.

snowman2:
Thanks @dgburr 👍

PR #1542, opened 2025-10-24
Fix linting issues

dgburr:
Address issues preventing https://github.com/pyproj4/pyproj/pull/1541 from passing in CI


snowman2:
Thanks @dgburr 👍

PR #1556, opened 2025-12-05
fix TransformerGroup crash when proj_normalize_for_visualization() returns NULL

### LINKED ISSUE #1464: TransformerGroup does not handle cases where both the allow_superseded and always_xy user arguments is used
#### Ex. 1 TransformerGroup fails to list up available transformations when specifying both ``always_xy`` and ``allow_superseded ``

```python
from pyproj.transformer import TransformerGroup
TransformerGroup("EPSG:4230", "EPSG:4326", allow_superseded=True, always_xy=True).transformers
```

Outputs:
```python
  File "/usr/local/lib/python3.11/site-packages/pyproj/transformer.py", line 207, in __init__
    super().__init__(
  File "pyproj/_transformer.pyx", line 241, in pyproj._transformer._TransformerGroup.__init__
  File "pyproj/_transformer.pyx", line 602, in pyproj._transformer._Transformer._from_pj
  File "pyproj/_transformer.pyx", line 653, in pyproj._transformer._Transformer._init_from_crs
  File "pyproj/_transformer.pyx", line 359, in pyproj._transformer._Transformer._initialize_from_projobj
pyproj.exceptions.ProjError: Input is not a transformation.
```

#### Ex.2 TransformerGroup should return ``36`` available transformations when  ``allow_superseded `` not defined
```python
ct_lst = TransformerGroup("EPSG:4230", "EPSG:4326").transformers
print(len(ct_lst))
```

Outputs:
```python
36
```

#### Ex.3 TransformerGroup should return ``37`` available transformations when ``allow_superseded `` is set to ``True``
```python
ct_lst = TransformerGroup("EPSG:4230", "EPSG:4326", allow_superseded=True).transformers
print(len(ct_lst))
```

Outputs:
```python
37
```

ct_lst = TransformerGroup("EPSG:4230", "EPSG:4326", allow_superseded=True).transformers
print(len(ct_lst))

### Problem description
``TransformerGroup``  fails to a create a list of with all available transformation objects when the user set both ``always_xy`` and ``allow_superseded `` equals ``True``. 

### Environment Information
pyproj version 3.6.1
python 3.11

### Installation method
 pip wheel

phaarnes:
Fix bug in TransformerGroup where it crashes when `proj_normalize_for_visualization()` returns `NULL`.

This issue became visible when using both `allow_superseded=True` and `always_xy=True`, which causes the superseded concatenated transformation [ED50 to WGS 84 (15) (EPSG:8047)](https://epsg.org/concatenated-operation_8047/ED50-to-WGS-84-15.html?sessionkey=0beyllj5ob) to be added to the list. Since the two sub-steps of the transformation do not have intersecting extents, PROJ fails to perform the axis swap, and `proj_normalize_for_visualization()` returns `NULL`.

```python 
from pyproj.transformer import TransformerGroup
TransformerGroup("EPSG:4230", "EPSG:4326", allow_superseded=True, always_xy=True).transformers
```

The fix:
- In `_set_always_xy()`: Check for `NULL` return and set `self.projobj` to `NULL` before raising `ProjError`, allowing the original pointer to remain valid
- In `_TransformerGroup.__init__()`: Catch `ProjError` from `_from_pj()` and add the operation to `unavailable_operations` instead of crashing

Operations that cannot be normalized for visualization (e.g., certain concatenated operations with non-overlapping step extents) are now properly tracked in `unavailable_operations` rather than causing a crash.

**Note:**
I created a PR to PROJ (now merged to main) where the extent validation is skipped for axis-swapping operations. This is safe because the axis-swap operations added by `normalizeForVisualization()` have no spatial effect—they only reorder coordinates—so the extent check can be safely skipped.

See: https://github.com/OSGeo/PROJ/pull/4632

Fixes #1464

 - ✅ Closes #1464 
 - ✅ Tests added

### DIFF
--- a/pyproj/_transformer.pyx
+++ b/pyproj/_transformer.pyx
@@ -143,6 +143,7 @@ cdef class _TransformerGroup:
             PJ_OPERATION_FACTORY_CONTEXT* operation_factory_context = NULL
             PJ_OBJ_LIST * pj_operations = NULL
             PJ* pj_transform = NULL
+            PJ* pj_transform_normalized = NULL
             PROJ_CRS_EXTENT_USE pj_crs_extent_use
             const char* c_authority = NULL
             int num_operations = 0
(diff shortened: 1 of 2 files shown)


phaarnes:
@snowman2 

Thanks for the feedback! I have now implemented the following fixes according to your review:

1. **Call proj_normalize_for_visualization directly in _TransformerGroup** - Refactored to check proj_normalize_for_visualization directly in _TransformerGroup.__init__ before calling _from_pj. If it returns NULL, the operation is added to unavailable_operations. We then pass always_xy=False to _from_pj since normalization is already done.

2. **Check for unavailable operations** - Updated the test to verify that unavailable_operations is populated correctly.

3. **Don't set self.projobj = NULL** - This was resolved as part of the refactor - _set_always_xy now just raises the exception and lets the destructor handle cleanup.

snowman2:
Thanks @phaarnes 👍

PR #1568, opened 2026-01-12
TST: Update tests with new transformations in PROJ 9.8

phaarnes:
I found one test failing because additional transformations are available in newer PROJ versions.


snowman2:
Thanks @phaarnes 👍

PR #1583, opened 2026-03-14
docs: added windows specfic intructions for building documentation

### LINKED ISSUE #1491: DOC: add help for building the docs in windows
When working on the doc I found that building the docs was only explained for Linux users.
[Link for the conversation](https://github.com/pyproj4/pyproj/pull/1489#issuecomment-2825301296)

I think we need to add a note for windows users, explaining two things :
- the need to `pip install` the file [requirements-docs.txt](https://github.com/pyproj4/pyproj/blob/main/requirements-docs.txt) ;
- the use of `make html` instead of `make docs`.

Troppy2:
Fixes #1491

Added Windows-specific instructions for building documentation:
- Note to use `make html` instead of `make docs`
- Clarify that `requirements-docs.txt` needs to be installed

This helps Windows users who may not be familiar with the Unix make commands.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -120,6 +120,8 @@ You can build the documentation locally to see how your changes will look.
 * Build the docs: ``make docs``
 * Or, to build and open in a browser: ``make docs-browser``
 
+**On Windows:** Use ``make html`` instead of ``make docs``. Also ensure you have installed the documentation requirements with ``pip install -r requirements-docs.txt`` before building.
+
 ## Tests
 Unit tests are the lifeblood of the project, as it ensures that we can continue to add and
 change the code and stay confident that things have not broken. Running the tests requires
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -120,6 +120,8 @@ You can build the documentation locally to see how your changes will look.
 * Build the docs: ``make docs``
 * Or, to build and open in a browser: ``make docs-browser``
 
+**On Windows:** Use ``make html`` instead of ``make docs``. Also ensure you have installed the documentation requirements with ``pip install -r requirements-docs.txt`` before building.
```
snowman2:
```suggestion
**On Windows:** Use ``make -C docs html`` instead of ``make docs``. Also ensure you have installed the documentation requirements with ``pip install --group docs`` before building.
```

snowman2: #1602

snowman2:
Thanks @Troppy2 👍

Troppy2:
You're welcome. Sorry for doing it so late. I didn't receive a notification about your message.

PR #1589, opened 2026-04-14
Spelling fix in test function names

kratzert:
Hi, I had my linter pointing out a few typos. Not sure if you care about these issues for test function names, but in case you want, here is a small PR to fix two occurrences.


snowman2:
Thanks @kratzert 👍

PR #1598, opened 2026-06-16
DOC: add CPython 3.14 support trove classifier

neutrinoceros:
Python 3.14 is supported de facto on the dev branch, as wheels are built and tested.


snowman2:
Thanks @neutrinoceros 👍

PR #1600, opened 2026-06-16
BLD: add option to build future-compatible abi3 wheels

neutrinoceros:
This allows targeting CPython's LimitedAPI (and stable ABI) by setting `PYPROJ_LIMITED_API=1` at build time. Resulting wheels are future compatible with yet-unreleased Python versions.
The trade-off is that runtime performance might be impacted, though I suggest measuring it against real load-bearing applications, as I found that in many numpy-based extensions the impact to be negligible or even non-measurable. 
I could successfully build the package and run the test suite with this locally.

I suggest a couple possible approaches to leverage this:
- *append* a `cp314-abi3` target to the existing build matrix. Version-specific wheels will still be prefered by installers when available but will default to these otherwise, instead of having to, and often failing to build pyproj from source. 
- *replace* version-specific wheels with a single `cp311-abi3` wheel per platform. If there are no visible performance regressions this is definitely the way to go.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -4,6 +4,7 @@ import re
 import shutil
 import subprocess
 import sys
+import sysconfig
 from pathlib import Path
 
 from setuptools import Extension, setup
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -15,6 +16,24 @@
 PROJ_VERSION_SEARCH = re.compile(r".*Rel\.\s+(?P<version>\d+\.\d+\.\d+).*")
 VERSION_SEARCH = re.compile(r".*(?P<version>\d+\.\d+\.\d+).*")
 
+# restrict LIMITED_API usage:
+# - require an env var PYPROJ_LIMITED_API=1
+# - LIMITED_API is not compatible with free-threading (as of CPython 3.14)
```
neutrinoceros, line 21: note this comment is going to age like milk because there's a new version of the Limited API coming up in Python 3.15 that's compatible with both GIL-enabled and free-threaded builds. However there are a couple other blockers right now (Cython and setuptools both have work-in-progress patches to support this). Note that I'd be happy to revisit this later this year to add `cp315-abi3.abi3t` as a target

PR #1602, opened 2026-06-24
Use dependency-groups (PEP 735) instead of requirements files

mwtoews:
This PR uses dependency groups ([PEP 735](https://peps.python.org/pep-0735/)), which is a standard Python feature supported by modern pip frontends.

This feature replaces the need for `requirements-*.txt` files, which are removed.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -57,6 +57,38 @@ changelog = "https://pyproj4.github.io/pyproj/stable/history.html"
 [project.scripts]
 pyproj = "pyproj.__main__:main"
 
+[dependency-groups]
+dev = [
+    "cython>=0.28.4",
+    "pre-commit",
+    "setuptools",
+    {include-group = "lint"},
+    {include-group = "test"},
+    {include-group = "typing"},
+]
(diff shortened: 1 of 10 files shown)


pyproject.toml
```diff
@@ -57,6 +57,38 @@ changelog = "https://pyproj4.github.io/pyproj/stable/history.html"
...
+    "sphinx-argparse",
+]
+lint = [
+    "pylint",
+    "ruff",
+]
+test = [
+    "numpy",
+    "packaging",
+    "pandas; python_version<'3.12' and implementation_name!='pypi'",
```
snowman2: Why are there python version checks?

mwtoews: This does the same logic cleaned-up in this PR for `.github/workflows/tests.yaml` (xref d4e338f5d704d96c20bcfcf712750a27f54b7cdc). However, checking with `uv sync` (which does all compatibility checks) it seems this is unnecessary, so they can be removed.

snowman2:
Thanks @mwtoews 👍

PR #1603, opened 2026-06-24
Update and fix linting

mwtoews:
It seems that the linting job fails, possibly due to older versions of dependencies. This updates pre-commit deps and fixes any issues. A few ruff rules are ignored, they can be revised later.


snowman2:
Thanks @mwtoews 👍

PR #1606, opened 2026-06-30
ENH: add guess_wkt_version to return the WKT version of a string

### LINKED ISSUE #1031: ENH: allow to check for specific WKT flavor in the is_wkt() function
There is a `is_wkt` function that allows you to check if a given string is a WKT string:

https://github.com/pyproj4/pyproj/blob/2d3cc7e0b83f59df2c0f5d3c2a8446ce0f224700/pyproj/_crs.pyx#L32-L48

The `proj` method it is using for that (`proj_context_guess_wkt_dialect`) should also allow to check for an exact flavor. So there could be an optional `version` keyword, to allow something like `is_wkt(..., version="WKT2_2019")`?

madhavcodez:
ENH: add guess_wkt_version to return the WKT version of a string

## What

Adds `pyproj.crs.guess_wkt_version(wkt)`, which returns the
`pyproj.enums.WktVersion` of a Well-Known Text string, or `None` when the
string is not WKT.

## Why

`is_wkt` only answers whether a string is WKT (a bool). PROJ's
`proj_context_guess_wkt_dialect` already reports the specific dialect, so this
exposes that information. This is the `guess_wkt_version` API proposed by the
maintainer in #1031.

Dialect mapping:

- `PJ_GUESSED_WKT2_2019` (and its legacy alias `PJ_GUESSED_WKT2_2018`) -> `WktVersion.WKT2_2019`
- `PJ_GUESSED_WKT2_2015` -> `WktVersion.WKT2_2015`
- `PJ_GUESSED_WKT1_GDAL` -> `WktVersion.WKT1_GDAL`
- `PJ_GUESSED_WKT1_ESRI` -> `WktVersion.WKT1_ESRI`
- `PJ_GUESSED_NOT_WKT` -> `None`

## Changes

- `pyproj/_crs.pyx`: new `guess_wkt_version` function mirroring `is_wkt`.
- `pyproj/proj.pxi`: declare the current `PJ_GUESSED_WKT2_2019` enum member.
- `pyproj/_crs.pyi`: type stub.
- `pyproj/crs/__init__.py`: export the function and add it to `__all__`.
- `docs/api/crs/crs.rst`: autofunction entry.
- `docs/history.rst`: changelog bullet.
- `test/crs/test_crs.py`: per-dialect round-trip tests plus a non-WKT case.

## Test plan

- [x] Built locally against PROJ 9.8.1 (conda-forge) and ran `test/crs/test_crs.py`.
- [x] Per-dialect round-trip via `CRS.to_wkt(version=...)` returns the matching `WktVersion`.
- [x] A PROJ string returns `None`.

Closes #1031


pyproj/_crs.pyx
```diff
@@ -49,6 +49,38 @@ def is_wkt(str proj_string not None):
...
+    # PJ_GUESSED_WKT2_2018 is the legacy alias of PJ_GUESSED_WKT2_2019.
+    if dialect == PJ_GUESSED_WKT2_2019:
+        return WktVersion.WKT2_2019
+    if dialect == PJ_GUESSED_WKT2_2015:
+        return WktVersion.WKT2_2015
+    if dialect == PJ_GUESSED_WKT1_GDAL:
+        return WktVersion.WKT1_GDAL
+    if dialect == PJ_GUESSED_WKT1_ESRI:
+        return WktVersion.WKT1_ESRI
+    return None
```
djhoese, line 81: Maybe this is just the python programmer in me, but this feels like it would work better as a dictionary `.get`:

```python
return WKT_DIALECTS.get(dialect)
```

where the dictionary is a mapping of the dialect strings and the WktVersion value. I assume these constants being compared against are just strings or integers so the dictionary hashing isn't a problem. I don't feel that strongly about it, but I do think it may be cleaner and more flexible in the long run.

PR #1608, opened 2026-07-13
Add Python 3.14 to the testing

cclauss:
https://docs.python.org/3/whatsnew/3.14.html


snowman2:
Thanks @cclauss 👍

cclauss:
Some tests are failing…

snowman2:
> Some tests are failing…

That is due to EPSG upgrades from PROJ.

PR #1609, opened 2026-07-25
DOC: Update fiona CRS compatibility for fiona 1.9+

### LINKED ISSUE #1360: DOC: Converting from fiona CRS to pyproj.crs.CRS
I was struggling with finding the best way to convert between fiona CRS and pyproj CRS.

I think that the section
https://pyproj4.github.io/pyproj/dev/crs_compatibility.html#converting-from-fiona-crs-to-pyproj-crs-crs
needs to be updated with regards to 
https://github.com/Toblerity/Fiona/issues/714

Also, `dc_crs` is undefined:
https://pyproj4.github.io/pyproj/dev/crs_compatibility.html#converting-from-pyproj-crs-crs-for-fiona
https://pyproj4.github.io/pyproj/dev/crs_compatibility.html#preparing-pyproj-crs-crs-for-geopandas

HugoFara:
Closes #1360.

Fiona 1.9 added fiona.crs.CRS, so a dataset's crs attribute is a CRS object rather than a PROJ string dictionary, and the linked Fiona issue asking for a CRS class is resolved. This restructures the section to mirror the rasterio one, as suggested in the issue.

Also fixes two errors in the existing example: dc_crs was an undefined name, and WktVersion was used without being imported.

Verified by running every snippet verbatim against fiona 1.10.1 / GDAL 3.9.2, including a write-back round-trip. Built against PROJ 9.8.1. sphinx-build -b html docs/ succeeds; pre-commit passes; pytest test/ is unchanged from baseline.

 - [x] Closes #1360
 - [x] Tests added

### DIFF
--- a/docs/crs_compatibility.rst
+++ b/docs/crs_compatibility.rst
@@ -112,17 +112,19 @@ fiona
 
 https://github.com/Toblerity/Fiona
 
-Converting from `fiona` CRS to `pyproj.crs.CRS`
+Converting from `fiona.crs.CRS` to `pyproj.crs.CRS`
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Fiona currently stores the CRS as a PROJ string dictionary in the `crs`
-attribute. As such, it is best to use the `crs_wkt` attribute.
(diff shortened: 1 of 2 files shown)


snowman2:
Thanks @HugoFara 👍

PR #1610, opened 2026-07-25
BUG: Honour PROJ spherification parameters in Geod (issue #1157)

### LINKED ISSUE #1157: +R_A appears to have no effect in Geod() and CRS()
## Problem description

The `+R_A` option is not being honored in Pyproj, whereas it seems to work fine using the `geod` CLI tool.

This problem happens if I use `pyproj.Geod()` or `pyproj.CRS().get_geod()`.

## Code sample

My Python code:

```python
import logging
import os

os.environ['PROJ_DEBUG'] = '3'
logging.basicConfig(level=logging.DEBUG)

import pyproj
# DEBUG:pyproj:PROJ_DEBUG: pj_open_lib(proj.ini): call fopen(/.../lib/python3.10/site-packages/pyproj/proj_dir/share/proj/proj.ini) - succeeded

geod1 = pyproj.Geod('+proj=lonlat +ellps=WGS84')
print(geod1.f)
# 0.0033528106647474805
geod1.fwd(-70, 40, 60, 10_000)
# (-69.89851793919364, 40.04498640681145, -119.93473805953444)

geod2 = pyproj.Geod('+proj=lonlat +ellps=WGS84 +R_A')
print(geod2.f)
# 0.0033528106647474805
geod2.fwd(-70, 40, 60, 10_000)
# (-69.89851793919364, 40.04498640681145, -119.93473805953444)
```

Equivalent CLI invocation using `geod`:

```sh
export PROJ_DEBUG=3

geod -f '%0.4f' +proj=lonlat +ellps=WGS84 <<< '-70E 40N 60d 10000m'
# pj_ellipsoid - final: a=6378137.000 f=1/298.257, errno=0
# pj_ellipsoid - final:    ellps=WGS84
# 70.0447	40.2273	-119.7864m

geod -f '%0.4f' +proj=lonlat +ellps=WGS84 +R_A <<< '-70E 40N 60d 10000m'
# pj_ellipsoid - final: a=6371007.181 f=1/  0.000, errno=0
# pj_ellipsoid - final:   R_A ellps=WGS84
# 70.0448	40.2282	-119.7855m
```

## Expected Output

I expected the output from `geod1` to be different from `geod2`, and that `geod2.f` should be `0`.

I was also surprised to see that the output from the `geod` command line tool was different from the output from `pyproj.Geod.fwd`. Presumably the difference can be ascribed to different options when compiling the CLI tool (installed via MacPorts) and the Python package (installed via the wheels on PyPI).

I further expected to see "trace" logging output comparable to what the `geod` CLI tool emitted, but I suppose that's a separate issue.

## Environment Information

MacOS 12.4 on ARM64.

### Versions

```
pyproj info:
    pyproj: 3.4.0
      PROJ: 9.1.0
  data dir: /.../lib/python3.10/site-packages/pyproj/proj_dir/share/proj
user_data_dir: /.../.local/share/proj
PROJ DATA (recommended version): 1.11
PROJ Database: 1.2
EPSG Database: v10.074 [2022-08-01]
ESRI Database: ArcGIS Pro 3.0 [2022-07-09]
IGNF Database: 3.1.0 [2019-05-24]

System:
    python: 3.10.6 (main, Sep 19 2022, 14:40:49) [Clang 13.1.6 (clang-1316.0.21.2.5)]
executable: /.../bin/python
   machine: macOS-12.4-arm64-arm-64bit

Python deps:
   certifi: 2022.9.14
    Cython: None
setuptools: 65.4.1
       pip: 22.2.2
```

### Installation method

Installed using Pip in a Venv:

```
python -m venv venv
./venv/bin/pip install -U pyproj
```

HugoFara:
`Geod` parses the ellipsoid out of its init string in Python rather than handing the string to PROJ, and that parser dropped every parameter without a value:

```python
# We can ignore safely any parameter that doesn't have a value
if kvpair.find("=") == -1:
    continue
```

`+R_A` and the rest of the spherification family went through there. The result was silently wrong rather than unsupported — `Geod('+ellps=WGS84 +R_A')` returned ellipsoidal answers with no error and no warning, while PROJ's own `geod` CLI honoured the flag.

This collects those parameters before valueless ones are discarded and applies the formulas from PROJ `src/ell_set.cpp` (`ellps_spherification`), including its mutual-exclusion order and the collapse to a sphere afterwards.

Semi-major axis for `+ellps=WGS84`, against `PROJ_DEBUG=3 geod ...`:

| parameter     | before      | after       | PROJ        |
| ------------- | ----------- | ----------- | ----------- |
| `+R_A`        | 6378137.000 | 6371007.181 | 6371007.181 |
| `+R_V`        | 6378137.000 | 6371000.790 | 6371000.790 |
| `+R_a`        | 6378137.000 | 6367444.657 | 6367444.657 |
| `+R_g`        | 6378137.000 | 6367435.680 | 6367435.680 |
| `+R_h`        | 6378137.000 | 6367426.702 | 6367426.702 |
| `+R_lat_a=45` | 6378137.000 | 6378110.053 | 6378110.053 |
| `+R_lat_g=45` | 6378137.000 | 6378101.030 | 6378101.030 |
| `+R_C`        | 6378137.000 | 6356752.314 | 6356752.314 |

`Geod.fwd` now reproduces the `geod` CLI to 7 decimal places on all of them. The `+R_A` radius also matches the published WGS84 authalic sphere radius (6371007.1809 m) to 0.2 mm, so the expected values do not rest on PROJ alone.

Two deliberate limits:

- PROJ parses the `+R_lat_a`/`+R_lat_g` latitude with `proj_dmstor` and so accepts DMS; this accepts decimal degrees only and raises `GeodError` otherwise.
- `CRS('+ellps=WGS84 +R_A').get_geod()` is unchanged. `+R_A` has no representation in a CRS, and PROJ drops it there too, so honouring it would diverge from PROJ rather than match it.

Verification: 14 new tests, all failing before the change and passing after. Full suite `1047 passed`, `mypy` clean, `pre-commit` clean. Built against PROJ 9.8.1.

 - [x] Closes #1157
 - [x] Tests added

### DIFF
--- a/docs/history.rst
+++ b/docs/history.rst
@@ -16,6 +16,7 @@ Latest
 - DOC: Update fiona CRS compatibility for fiona 1.9+ (issue #1360)
 - BUG: Clear CONTEXT_THREAD_KEY when destroying PJ_CONTEXT (pull #1541)
 - BUG: Default skew angle for CF grid mapping oblique mercator to 90 (issue #1506)
+- BUG: Honour the ``+R_A``, ``+R_V``, ``+R_a``, ``+R_g``, ``+R_h``, ``+R_lat_a``, ``+R_lat_g`` and ``+R_C`` spherification parameters in :class:`pyproj.Geod` (issue #1157)
 - BLD: update build-time dependencies
 
 3.7.2
--- a/pyproj/geod.py
+++ b/pyproj/geod.py
@@ -56,6 +56,101 @@ def _params_from_ellps_map(ellps: str) -> tuple[float, float, float, float, bool
     return semi_major_axis, semi_minor_axis, flattening, eccentricity_squared, sphere
 
 
+# Spherification parameters, in the order PROJ resolves them. They are mutually
+# exclusive: PROJ applies the first one present in this order and ignores the
+# rest. See ellps_spherification in PROJ src/ell_set.cpp.
+_SPHERIFICATION_KEYS = ("R_A", "R_V", "R_a", "R_g", "R_h", "R_lat_a", "R_lat_g", "R_C")
+
+# Series coefficients for the equal-area and equal-volume spheres,
+# from PROJ src/ell_set.cpp.
+_SIXTH = 1 / 6.0
+_RA4 = 17 / 360.0
+_RA6 = 67 / 3024.0
+_RV4 = 5 / 72.0
+_RV6 = 55 / 1296.0
+
+
+def _sphere_radius_from_spherification(
+    requested: dict[str, str],
+    semi_major_axis: float,
+    semi_minor_axis: float,
+    eccentricity_squared: float,
+    lat_0: float,
+) -> float:
+    """
+    Radius of the sphere requested by a PROJ spherification parameter.
+
+    Parameter
+    ---------
+    requested: dict[str, str]
+        The spherification parameters found in the input, mapped to their
+        (possibly empty) values.
+    semi_major_axis: float
+        The semi-major axis of the ellipsoid to spherify.
+    semi_minor_axis: float
+        The semi-minor axis of the ellipsoid to spherify.
+    eccentricity_squared: float
+        The eccentricity squared of the ellipsoid to spherify.
+    lat_0: float
+        Latitude of origin in degrees, used only by ``+R_C``.
+
+    Returns
+    -------
+    float
+
+    """
+    key = next(key for key in _SPHERIFICATION_KEYS if key in requested)
+    if key == "R_A":
+        # sphere with the same surface area as the ellipsoid
+        return semi_major_axis * (
+            1.0
+            - eccentricity_squared
+            * (_SIXTH + eccentricity_squared * (_RA4 + eccentricity_squared * _RA6))
+        )
+    if key == "R_V":
+        # sphere with the same volume as the ellipsoid
+        return semi_major_axis * (
+            1.0
+            - eccentricity_squared
+            * (_SIXTH + eccentricity_squared * (_RV4 + eccentricity_squared * _RV6))
+        )
+    if key == "R_a":
+        return (semi_major_axis + semi_minor_axis) / 2
+    if key == "R_g":
+        return math.sqrt(semi_major_axis * semi_minor_axis)
+    if key == "R_h":
+        if semi_major_axis + semi_minor_axis == 0:
+            raise GeodError("Invalid ellipsoid for +R_h: a + b is zero.")
+        return (
+            2 * semi_major_axis * semi_minor_axis / (semi_major_axis + semi_minor_axis)
+        )
+    # The remaining variants are evaluated at a latitude. PROJ parses that
+    # latitude with proj_dmstor and so also accepts DMS strings; only decimal
+    # degrees are supported here.
+    raw_latitude = lat_0 if key == "R_C" else requested[key]
+    try:
+        latitude = float(raw_latitude)
+    except (TypeError, ValueError) as error:
+        raise GeodError(
+            f"Invalid latitude for +{key}: {raw_latitude!r}. "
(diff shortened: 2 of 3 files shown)


jjimenezshaw:
Hi @HugoFara 
Reading the PR description I thought that it was passing the parameters like `+R_A` to PROJ. However I see that it is re-implementing all the logic in pyproj. That is not the idea of pyproj I guess.
I think it should use the PROJ functions `pj_ell_set` or `pj_ellipsoid`.

HugoFara:
Hi @jjimenezshaw . I couldn't find a way to reach either function from pyproj, maybe you can help me here.

Both are declared in `proj_internal.h`, not `proj.h`, and that header isn't installed. PROJ ships five: `HEADERS_LIBPROJ` in `src/lib_proj.cmake` is `proj.h`, `proj_experimental.h`, `proj_constants.h`, `proj_symbol_rename.h`, `geodesic.h` (plus `apps/projapps_lib.h`). I checked the actual `proj-devel` 9.8.1 package and `proj_internal.h` isn't in it.

Only `pj_ell_set` is exported as `_Z10pj_ell_setP6pj_ctxP8ARG_listPdS3_`

`pj_ellipsoid` isn't callable from outside libproj at all. `pj_ell_set` has `PROJ_DLL` but no `extern "C"`, so what's actually exported is a C++-mangled name rather than stable API (should we build an internal `paralist`?).

Have I missed a supported route? If pyproj is expected to build against `proj_internal.h`, I'll happily redo it that way.

I also notived there is a way to let PROJ do the work through the public API only. `+proj=cart` applies spherification like any other PJ, so the ellipsoid can be read back by transforming two points: `(0, 0, 0)` gives `x = a`, and `(0, 90, 0)` gives `z = b`:

| params | a | b |
|---|---|---|
| `+ellps=WGS84` | 6378137.0000 | 6356752.3142 |
| `+ellps=WGS84 +R_A` | 6371007.1811 | 6371007.1811 |
| `+ellps=WGS84 +R_lat_a=45` | 6378110.0529 | 6378110.0529 |
| `+ellps=bessel` | 6377397.1550 | 6356078.9628 |
| `+a=6378137 +rf=298.257223563` | 6378137.0000 | 6356752.3142 |

That covers named ellipsoids, explicit `+a`/`+rf` and all eight spherification variants, with PROJ doing every bit of the math. It would remove the formulas in this PR along with the hand-rolled proj-string parser and the `pj_ellps` lookup in `_params_from_ellps_map`, so it fixes the class of bug rather than this one instance, and DMS latitudes for `+R_lat_a`/`+R_lat_g`, which this PR doesn't support, would come for free.

The trade-off is that it's a much bigger change than #1157 warrants: it moves `Geod.__init__` into the Cython layer and touches every construction path, and it's a behaviour change for anyone relying on the current parsing quirks.

Would you prefer I take this PR that way, or keep it narrow and open a separate issue for the delegation? Happy either way.

jjimenezshaw:
You are right. It is only exposed in the C++ API.

I don't know what would be the best option. Should we expose something in the C API, @rouault  @kbevers ? Is there another function that we don't see?

HugoFara:
Hi, I saw you marked the work as planned @snowman2 

I updated the branch, closed the coverage gap and fixed failing tests. Let me know if you'd need anything else.

snowman2:
The code looks good from an overview and tests standpoint. I agree with @jjimenezshaw about the logic living in PROJ in the long term. However, I don't see a function API for converting the geod string to a geod object.

@HugoFara are you comfortable enough in the PROJ codebase to attempt adding a public method to create a geod object from a geod string?

HugoFara:
I can try it. I'll require the help of my Claude to explain the specifics of the codebase, thanks for trusting it with me @snowman2 .

Issue #1588, opened 2026-04-12
TST Fix tests after PROJ 9.8.1 EPSG database revert


lpechacek:
Apologies for not following the bug report template closely but I think that this issue is trivial enough to be explained in a few sentences.

#### Problem description

I'm packaging [pyproj for openSUSE](https://build.opensuse.org/package/show/Application:Geo/python-pyproj). The tests started failing after PROJ upgrade to 9.8.0. I've ported the necessary patches (in essence, anything that has "test" in subject) from this repo and pyproj tests pass again.

However, PRO 9.8.1 has been released with a revert of some EPSG database changes. That partly restores pre-9.8.0 PROJ behavior, and pyproj tests need to be adjusted for that.

#### Proposed fix

The following crude patch serves well as an band-aid. Feel free to rework it to your standards. Thanks!
```
From b266041a437138c81ebfd565175143f2714e7911 Mon Sep 17 00:00:00 2001
From: Libor Pechacek <lpechacek@gmx.com>
Date: Sun, 12 Apr 2026 15:29:04 +0200
Subject: [PATCH] TST Fix tests after PROJ 9.8.1 EPSG database revert

PROJ reverted EPSG database from EPSG 12.054 to 12.029 with v9.8.1, so
we need to re-adjust the test for the previous behavior.
---
 test/conftest.py         |  1 +
 test/crs/test_crs.py     |  3 ++-
 test/test_transformer.py | 13 +++++++------
 3 files changed, 10 insertions(+), 7 deletions(-)

diff --git a/test/conftest.py b/test/conftest.py
index 00a85375f014..17384b9b9798 100644
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -16,6 +16,7 @@ PROJ_GTE_941 = PROJ_LOOSE_VERSION >= version.parse("9.4.1")
 PROJ_GTE_95 = PROJ_LOOSE_VERSION >= version.parse("9.5.0")
 PROJ_GTE_971 = PROJ_LOOSE_VERSION >= version.parse("9.7.1")
 PROJ_GTE_980 = PROJ_LOOSE_VERSION >= version.parse("9.8.0")
+PROJ_GTE_981 = PROJ_LOOSE_VERSION >= version.parse("9.8.1")
 
 
 def unset_data_dir():
diff --git a/test/crs/test_crs.py b/test/crs/test_crs.py
index 1eefb4cc6de6..4a077c15f2e4 100644
--- a/test/crs/test_crs.py
[...]

PR #1611, opened 2026-07-25
DOC: describe geometry_area_perimeter size limit concretely

### LINKED ISSUE #1592: geometry_area_perimeter limitations not documented properly
Essentially I would like to reopen #1400 as the documentation sentence introduced there is too vague.

The documentation [reads](https://pyproj4.github.io/pyproj/stable/api/geod.html#pyproj.Geod.geometry_area_perimeter):

> There are a few limitations :
> - only works with areas up to half the size of the globe ;
> - certain large polygons may return negative values.

As someone who would like to use this function, this doesn't give me enough information to judge whether I can use the function or not:

1. What is "large"? City size? Country size? Half the world (as possibly suggested by the bulletpoint above)?
2. In which "certain" cases does this happen? When located near the poles or maybe across the antimeridian? When the polygon has a funny shape? Simply when exceeding a certain size?

It would be nice to provide a more concrete wording, e.g. (if correct):
> large polygons over X m^2 return negative values

Htet-Kaung-San:
Closes #1592.

The note on `Geod.geometry_area_perimeter` said only:

> There are a few limitations :
> - only works with areas up to half the size of the globe ;
> - certain large polygons may return negative values.

As #1592 points out, that doesn't say what "large" or "certain" mean — whether it's city-sized or continent-sized, and whether shape, poles or the antimeridian are involved.

Both bullets turn out to be the same rule. The area is reduced modulo the total area of the ellipsoid into `(-A/2, A/2]`, so a region covering more than half the ellipsoid is reported as the negated area of its complement:

```python
from pyproj import Geod
from shapely.geometry import Polygon

geod = Geod(ellps="WGS84")
nearly_global = Polygon([(-179, -89), (179, -89), (179, 89), (-179, 89)])
print(geod.geometry_area_perimeter(nearly_global)[0])
# -2833262901265.9062
```

That polygon is wound counter-clockwise and encloses 99.4% of the globe, but comes back as roughly -0.56% — the thin strip left outside it, negated.

So the answers to the two questions in the issue are concrete:

- **"large"** means precisely "encloses more than half the ellipsoid" — about `2.55e14` m² on WGS84.
- **"certain"** isn't about shape, proximity to a pole, or crossing the antimeridian. Size alone decides it.

I checked the `(-A/2, A/2]` bound holds across a grid of boxes spanning ±10° to ±179.9° longitude and ±10° to ±89.9° latitude.

The docstring already carries a separate warning that traversal direction is signed, CCW positive. That's the other reason a result can be negative, and unlike the size limit it applies at any size, so I cross-referenced the two rather than describing them independently.

### Changes

- Replace the vague note with the modulo rule, the WGS84 numbers, and an explicit statement of what does *not* matter.
- Cross-reference the traversal-direction warning as the other, size-independent cause.
- Add doctests for the wraparound and for recovering the enclosed area by adding the total ellipsoid area back.
- Changelog entry.

Documentation only — no behavior change.

### Testing

`flake8`, `ruff` and `codespell` are clean on the changed files, and `blacken-docs --skip-errors` leaves them unchanged. The docstring's doctests pass (14 examples, 0 failures), including the two new ones. The `99.4%` figure is `99.4445%` before rounding, so it isn't sitting on a rounding boundary.

### DIFF
--- a/pyproj/geod.py
+++ b/pyproj/geod.py
@@ -1080,20 +1080,11 @@ class Geod(_Geod):
         ...         Point(-179, -89), Point(179, -89), Point(179, 89), Point(-179, 89)
         ...     ])
         ... )
-        >>> big_area, _ = geod.geometry_area_perimeter(nearly_global)
-        >>> big_area < 0
+        >>> geod.geometry_area_perimeter(nearly_global)[0] < 0
         True
 
         Its magnitude is the area of the thin strip left outside the polygon, so
-        adding the total area of the ellipsoid recovers the enclosed area:
-
-        >>> import math
-        >>> b = geod.a * (1 - geod.f)
-        >>> ecc2 = 1 - (b * b) / (geod.a * geod.a)
-        >>> ecc = math.sqrt(ecc2)
-        >>> total = 2 * math.pi * geod.a**2 * (1 + (1 - ecc2) / ecc * math.atanh(ecc))
-        >>> f"{100 * (big_area + total) / total:.1f}%"
-        '99.4%'
+        adding the total area of the ellipsoid recovers the enclosed area.
 
 
         Parameters
--- a/pyproj/geod.py
+++ b/pyproj/geod.py
@@ -1031,15 +1031,24 @@ class Geod(_Geod):
 
         .. note:: lats should be in the range [-90 deg, 90 deg].
 
-        .. note:: | There are a few limitations :
-                  | - only works with areas up to half the size of the globe ;
-                  | - certain large polygons may return negative values.
+        .. note:: The area is reduced modulo the total area of the ellipsoid,
+                  into the range ``(-A/2, A/2]``, where ``A`` is that total area
+                  (about ``5.10e14`` m\\ :sup:`2` on WGS84, so ``A/2`` is about
+                  ``2.55e14`` m\\ :sup:`2`). A region covering more than half the
+                  ellipsoid is therefore reported as the negated area of its
+                  complement: a polygon enclosing 99% of the globe returns about
+                  -1% of the total area, not +99%. Size is the only reason a
+                  correctly oriented polygon comes back negative, and "large"
+                  means precisely "more than half the ellipsoid" -- shape,
+                  proximity to a pole and crossing the antimeridian do not
+                  matter on their own.
 
         .. warning:: The area returned is signed with counter-clockwise (CCW) traversal
                      being treated as positive. For polygons, holes should use the
                      opposite traversal to the exterior (if the exterior is CCW, the
                      holes/interiors should be CW). You can use `shapely.ops.orient` to
-                     modify the orientation.
+                     modify the orientation. This is the other reason a result can be
+                     negative, and unlike the size limit above it applies at any size.
 
         If it is a Polygon, it will return the area and exterior perimeter.
         It will subtract the area of the interior holes.
@@ -1062,6 +1071,29 @@ class Geod(_Geod):
         ... )
         >>> f"{poly_area:.0f} {poly_perimeter:.0f}"
         '944373881400 3979008'
+
+        A polygon enclosing more than half the ellipsoid wraps around to a
+        negative value, even though its vertices are ordered counter-clockwise:
+
+        >>> nearly_global = Polygon(
+        ...     LineString([
+        ...         Point(-179, -89), Point(179, -89), Point(179, 89), Point(-179, 89)
+        ...     ])
+        ... )
+        >>> big_area, _ = geod.geometry_area_perimeter(nearly_global)
+        >>> big_area < 0
+        True
+
+        Its magnitude is the area of the thin strip left outside the polygon, so
+        adding the total area of the ellipsoid recovers the enclosed area:
+
+        >>> import math
+        >>> b = geod.a * (1 - geod.f)
+        >>> ecc2 = 1 - (b * b) / (geod.a * geod.a)
+        >>> ecc = math.sqrt(ecc2)
+        >>> total = 2 * math.pi * geod.a**2 * (1 + (1 - ecc2) / ecc * math.atanh(ecc))
+        >>> f"{100 * (big_area + total) / total:.1f}%"
+        '99.4%'
 
 
         Parameters
--- a/pyproj/geod.py
+++ b/pyproj/geod.py
@@ -1078,20 +1078,11 @@ class Geod(_Geod):
         ...         Point(-179, -89), Point(179, -89), Point(179, 89), Point(-179, 89)
         ...     ])
         ... )
-        >>> big_area, _ = geod.geometry_area_perimeter(nearly_global)
-        >>> big_area < 0
+        >>> geod.geometry_area_perimeter(nearly_global)[0] < 0
         True
 
         Its magnitude is the area of the thin strip left outside the polygon, so
-        adding the total area of the ellipsoid recovers the enclosed area:
-
-        >>> import math
-        >>> b = geod.a * (1 - geod.f)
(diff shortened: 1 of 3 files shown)


Htet-Kaung-San:
Gentle nudge on this one — it's a docs-only change and the workflow is sitting at "action_required" (the first-time-contributor gate), so it hasn't run since I pushed the fixup.

Two things to save a review round:

- The earlier CI failure was mine: the new examples are shapely-based, so on environments without shapely the doctest count went 6 → 10 and `test_doctest_wrapper` tripped. Fixed in 0db4de4 by trimming one example and updating the expected count to 8. Verified against a source build with PROJ 9.8.1 — passes at 8 without shapely, 0 with, and unmodified `main` still passes at 6.
- The `Conda ubuntu-latest` job fails on an EPSG-name mismatch (`RGF93 v1 to WGS 84 (1)` vs `ETRS89-FRA…`), which is the pre-existing #1588 breakage also red on `main`, not something from this PR.

Happy to rebase or adjust wording if you'd prefer.

djhoese:
I just rebased your branch (be careful pulling locally, may need `git reset --hard origin/geometry-area-perimeter-limits`. This rebase should have pulled in the latest CI fixes and kept the git history pretty clean.

djhoese (approved):
This looks reasonable to me. We'll wait for CI. Thanks!

Note: I won't pretend to fully understand the details here so if another maintainer wants to comment on this it would be appreciated.

pyproj/geod.py
```diff
@@ -1029,15 +1029,24 @@ def geometry_area_perimeter(
 
         .. note:: lats should be in the range [-90 deg, 90 deg].
 
-        .. note:: | There are a few limitations :
-                  | - only works with areas up to half the size of the globe ;
-                  | - certain large polygons may return negative values.
+        .. note:: The area is reduced modulo the total area of the ellipsoid,
```
jorisvandenbossche: I think this reads very difficult (at least for the non-native speaker that I am), what does "reduced modulo" mean?

pyproj/geod.py
```diff
@@ -1029,15 +1029,24 @@ def geometry_area_perimeter(
...
+                  into the range ``(-A/2, A/2]``, where ``A`` is that total area
+                  (about ``5.10e14`` m\\ :sup:`2` on WGS84, so ``A/2`` is about
+                  ``2.55e14`` m\\ :sup:`2`). A region covering more than half the
+                  ellipsoid is therefore reported as the negated area of its
+                  complement: a polygon enclosing 99% of the globe returns about
+                  -1% of the total area, not +99%. Size is the only reason a
+                  correctly oriented polygon comes back negative, and "large"
+                  means precisely "more than half the ellipsoid" -- shape,
```
jorisvandenbossche: I think there is no need to refer to "large" and explaining what it means, since the actual text using that word has now been removed (this now rather makes it confusing)

pyproj/geod.py
```diff
@@ -1061,6 +1070,20 @@ def geometry_area_perimeter(
...
+        A polygon enclosing more than half the ellipsoid wraps around to a
+        negative value, even though its vertices are ordered counter-clockwise:
+
+        >>> nearly_global = Polygon(
+        ...     LineString([
+        ...         Point(-179, -89), Point(179, -89), Point(179, 89), Point(-179, 89)
+        ...     ])
+        ... )
+        >>> geod.geometry_area_perimeter(nearly_global)[0] < 0
+        True
```
jorisvandenbossche: This is not actually a "nearly global" polygon in practice, because for geodesic edges, it takes the shortest path on the sphere, so this is actually connecting -179 to 179 crossing the anti-meridian, i.e. an edge of 2 degrees long, and not 358 degrees long. As a consequence of that interpretation, this polygon is actually oriented clock-wise (although shapely, which is not using geodesic logic, will indicate this is CCW), and _that_ is the reason this is giving a negative result

snowman2:
Thanks all for the input so far. @cffk, we welcome your insight if you have a moment. If you're busy, no worries.

Htet-Kaung-San:
Thanks for the careful review — pushed a commit that reworks this.

@jorisvandenbossche you're right on all three points, and the antimeridian one corrected my understanding, so thank you. I checked it before rewriting: that `nearly_global` ring has a perimeter of **~39,600 km**, which is almost exactly two pole-to-pole meridian arcs (2 × 19,781 km) — so geodesically it really is a thin ~2°-wide strip across the antimeridian, oriented clockwise, and its magnitude (0.55% of the ellipsoid) is that strip's own area, not the complement of a near-global region. Reversing the ring flips the sign, confirming it's orientation, not size. My example and its "encloses 99%" framing were wrong, so I've removed it.

What changed:
- Dropped the "reduced modulo `(-A/2, A/2]`" wording and the explanation of "large" (which no longer appears in the text).
- Removed the `nearly_global` doctest.
- Rewrote the note around orientation and the geodesic-edge behaviour — i.e. that edges take the short way round, so a ring can be clockwise on the ellipsoid even when its coordinates look counter-clockwise (the antimeridian case). That's the real answer to "when do I get a negative value".

Removing that example takes the doctest count back to 6, so `test_doctest_wrapper` is restored to its original value and the PR now only touches prose.

I've kept the new note deliberately short and I'd defer to @cffk on the precise geodesic wording if he has a moment — happy to adjust it further.

pyproj/geod.py
```diff
@@ -1029,9 +1029,16 @@ def geometry_area_perimeter(
 
         .. note:: lats should be in the range [-90 deg, 90 deg].
 
-        .. note:: | There are a few limitations :
-                  | - only works with areas up to half the size of the globe ;
-                  | - certain large polygons may return negative values.
+        .. note:: The area is *signed*: it is the area to the left of the
```
jorisvandenbossche, line 1032: The fact that it is "signed" is already in the warning below

pyproj/geod.py
```diff
@@ -1029,9 +1029,16 @@ def geometry_area_perimeter(
...
+                  boundary as it is traversed, so a clockwise ring gives a
+                  negative value (see the warning below on orientation). Two
+                  things can make the sign surprising. The edges are geodesics
+                  and each is taken the short way round the ellipsoid, so a ring
+                  can be clockwise on the ellipsoid even when its
+                  longitude/latitude coordinates look counter-clockwise -- for
+                  example when an edge spans the antimeridian. And the magnitude
+                  is at most half the ellipsoid's total area, so a region larger
+                  than that half is reported as the negated area of the rest.
```
jorisvandenbossche, line 1041: This is really a separate thing, and should be treated as such for clarity, I think (like before it was a separate bullet point)

jorisvandenbossche:
@Htet-Kaung-San it seems that you are an agent, or a human just pasting the output of the agent (and apologies if this assumption is incorrect, unfortunately this is how it is nowadays ..). 
I would encourage you to actually engage with the topic yourself and try to understand it, and rewrite the doc note yourself based on that understanding. If I am just giving feedback to an agent on the exact wording, I would rather propose a wording myself.

cffk:
My $0.02:

The documentation says it's measuring the area of a "geodesic polygon as a shapely geometry".  Because the shapely specification stipulates that the line joining two points uses linear interpolation (and this then depends on what coordinates are used), it's important to document *up front* that *this* routine connects the consecutive points with the *shortest geodesic*.  (In parentheses, add that if there are multiple shortest geodesics, the area is not uniquely defined.  In this case, an additional point can be added to make the path unique.)

Perhaps replace the discussion about lines crossing the antimeridian with the observation that a latitude-longitude shapely polygon with its linear interpolation cannot encircle a pole.  But once the vertices are connected with geodesics, pole encircling polygons are possible.

All the rest of the discussion concerning signs and reducing the area to `(-A/2, A/2]` is fine.

PR #1615, opened 2026-08-08
Fix doubled word in transformer docstring

latent-9:
Small doubled-word typo in the `Transformer` docstring in `pyproj/transformer.py`: "In those cases, you can enable enable ..." -> "you can enable".


snowman2:
Thanks @latent-9 👍

PR #1616, opened 2026-08-08
perf: improve single point transform and remove warning from numpy scalars

### LINKED ISSUE #1309: Remove special-case casting support for numpy<2.0
Once we support only Numpy 2.0+ we can remove the casting checks we have added to the Cython code for single-point transforms and Geod's fwd/inv single-point as well. Right now we make sure to throw a TypeError so we get the right conversions, but Numpy 2.0+ will do this for us automatically.

https://github.com/pyproj4/pyproj/blob/f235f9b3067e4817ec34a95cfd1717d58fc95293/pyproj/_transformer.pyx#L768-L770

greglucas:
Change from a try/except raise case for numpy array vs scalar checking to a `_is_scalar` check that also removes the numpy warning cases without any cutout needed for numpy specifically. This is both faster and cleaner. I think that before I was potentially trying two things that degraded performance: `isinstance(x, numbers.Real)` which goes back out to Python rather than Cython being able to check in C against float/int which we have now, and putting it after the type-cast cdef declarations (now the is_scalar check comes before that).

 - [x] Closes #1309

### DIFF
--- a/pyproj/geod.py
+++ b/pyproj/geod.py
@@ -300,19 +300,18 @@ class Geod(_Geod):
         scalar or array:
             Back azimuth(s) or Forward azimuth(s)
         """
-        try:
-            # Fast-path for scalar input, will raise if invalid types are input
-            # and we can fallback below
-            return self._fwd_point(
-                lons,
-                lats,
-                az,
-                dist,
-                radians=radians,
-                return_back_azimuth=return_back_azimuth,
-            )
-        except TypeError:
-            pass
+        # Fast-path for scalar input; returns None for non-scalar input
+        # and we fall back below
+        point_data = self._fwd_point(
+            lons,
+            lats,
+            az,
+            dist,
+            radians=radians,
+            return_back_azimuth=return_back_azimuth,
+        )
+        if point_data is not None:
+            return point_data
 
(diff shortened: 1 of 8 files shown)


pyproj/geod.py
```diff
@@ -300,19 +300,18 @@ def fwd(  # pylint: disable=invalid-name
...
-                az,
-                dist,
-                radians=radians,
-                return_back_azimuth=return_back_azimuth,
-            )
-        except TypeError:
-            pass
+        # Fast-path for scalar input; returns None for non-scalar input
+        # and we fall back below
+        point_data = self._fwd_point(
```
snowman2, line 305: Thoughts about using the warus operator for all of these?
```suggestion
        if (point_data := self._fwd_point(
```

greglucas: I looked at it in the call sites and I'm not sure it makes it easier to read here because there are so many nested lines and the extra parentheses (I find the walrus to be easier to read when everything is on the same line). The current path isn't too bad because it is just used as a quick return, but I can make that change if you prefer.

I did notice in the _compat.pxd that I could simplify the `getattr()` comparison though so I did make that update just now.

snowman2:
Thanks @greglucas 👍

PR #1617, opened 2026-08-09
PERF: fast paths for CRS equality and hashing

greglucas:
Comparing two CRS objects goes through PROJ's proj_is_equivalent_to, which short-circuits on matching identifiers (EPSG codes) but otherwise falls through to a full recursive structural comparison of the datum, ellipsoid, prime meridian, coordinate system and conversion. For CRS built from PROJ strings, which carry no identifier, that costs 11-15us per comparison.

Noticed this in Cartopy where we do a check to see if `self == src_crs` to try and speed up transforms, but the CRS comparison itself can be slow.
https://github.com/SciTools/cartopy/blob/bdb6cc45e27ff4fe71741b445ac15449aee35ba3/lib/cartopy/crs.py#L410

The short circuits are the simple portion `is self` and `srs1 == srs2`, but the problem is that in missing cases we will still fall back to PROJ comparisons which is why I added the cache portion here which is really to speed up the != comparisons.

 - [x] Tests added

### DIFF
--- a/pyproj/crs/crs.py
+++ b/pyproj/crs/crs.py
@@ -75,7 +75,10 @@ _COMPARISON_CACHE_LOCK = threading.Lock()
 
 
 def _cached_comparison(this: "CRS", other: "CRS", mode: str) -> bool:
-    key = (this.srs, other.srs, mode)
+    # PROJ's comparisons are symmetric, so order the srs pair to let
+    # a-vs-b and b-vs-a share an entry.
+    srs1, srs2 = this.srs, other.srs
+    key = (srs1, srs2, mode) if srs1 <= srs2 else (srs2, srs1, mode)
     try:
         return _COMPARISON_CACHE[key]
     except KeyError:
@@ -85,7 +88,8 @@ def _cached_comparison(this: "CRS", other: "CRS", mode: str) -> bool:
     else:
         result = this._crs.equals(other._crs, ignore_axis_order=mode == "ignore_axis")
     with _COMPARISON_CACHE_LOCK:
-        # bound the cache, evicting the oldest entries first
+        # bound the cache, evicting in insertion order (FIFO, not LRU: hits
+        # are served without taking the lock)
         while len(_COMPARISON_CACHE) >= _COMPARISON_CACHE_MAXSIZE:
             _COMPARISON_CACHE.popitem(last=False)
         _COMPARISON_CACHE[key] = result
--- a/pyproj/crs/crs.py
+++ b/pyproj/crs/crs.py
@@ -8,6 +8,7 @@ import json
 import re
 import threading
 import warnings
+from collections import OrderedDict
 from collections.abc import Callable
 from typing import Any, Optional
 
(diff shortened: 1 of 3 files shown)


pyproj/crs/crs.py
```diff
@@ -64,6 +64,30 @@ def __init__(self):
...
+        return _COMPARISON_CACHE[key]
+    except KeyError:
+        pass
+    if mode == "exact":
+        result = this._crs.is_exact_same(other._crs)
+    else:
+        result = this._crs.equals(other._crs, ignore_axis_order=mode == "ignore_axis")
+    if len(_COMPARISON_CACHE) >= _COMPARISON_CACHE_MAXSIZE:
+        # bound the cache, evicting the oldest entry
+        _COMPARISON_CACHE.pop(next(iter(_COMPARISON_CACHE)), None)
```
greglucas: `pop(*, None)` to protect against the thread race conditions here if another thread has popped the item first. Worst case we fall back to the slow comparison route.

pyproj/crs/crs.py
```diff
@@ -64,6 +65,33 @@ def __init__(self):
         super().__init__()
 
 
+_COMPARISON_CACHE_MAXSIZE = 256
+# Comparison results, keyed on (srs, srs, mode). PROJ's comparison recurses
+# through the datum, coordinate system and conversion of both objects and is
+# expensive, but the result only depends on the two input strings.
+_COMPARISON_CACHE: OrderedDict[tuple[str, str, str], bool] = OrderedDict()
```
snowman2, line 72: Have you already considered using cachetools for this?

pyproj/crs/crs.py
```diff
@@ -64,6 +65,33 @@ def __init__(self):
...
+# Guards mutation of _COMPARISON_CACHE
+_COMPARISON_CACHE_LOCK = threading.Lock()
+
+
+def _cached_comparison(this: "CRS", other: "CRS", mode: str) -> bool:
+    key = (this.srs, other.srs, mode)
+    try:
+        return _COMPARISON_CACHE[key]
```
snowman2, line 83: It would potentially be beneficial if the key was unique regardless of the order of comparison.

Ideas I have had:

- swap crs string order to check for existence on cache
- sort crs string tuple when creating key

Not sure if the added logic will negatively impact performance or if the gains are worth it. Just ideas at the moment.

greglucas: :+1: good idea. Sorting the crs string turned out to be a simple addition that doesn't add much overhead at all (30ns). Added in the latest commit.

pyproj/crs/crs.py
greglucas: assuming you meant functools lru_cache here (not adding a new dependency). Unfortunately, `functools.lru_cache` doesn't work either. It can't key on CRS (it hashes and `__eq__`s its keys, and `CRS.__eq__` calls equals, so it recurses back into this cache), which means keying on the srs strings and rebuilding both CRS on a miss. That's over 100 us vs 6.5us for the PROJ comparison itself.

Issue #1618, opened 2026-08-09
Ideas to reduce repository size


greglucas:
I just updated my local clone after a while and noticed that I had to pull over 500MB for the repo to get updated. Most of this is from the github-pages content that is in build artifacts so not the actual pages HTML content itself (in `build/.doctrees/` build folder which we then deploy the entire build/ tree of)

There are a few items that we could do to reduce the repo size moving forward:

1. one line `-d` flag fix to put the _doctrees outside of the build directory.

```
sphinx-build -b html -d docs/_doctrees docs/ docs/_build/
```
(Equivalently, rm -rf docs/_build/.doctrees before the deploy step.)

2. Filter out the current content
 
* Squash current commits on the `gh-pages` branch and remove that folder

or

* `git filter-repo --invert-paths --path-glob '*/.doctrees/*'` on that branch. 

Might need to open a GitHub Support ticket asking them to run git gc on the repo?

3. Switch to `actions/deploy-pages`

artifact-based, no gh-pages branch at all, https://github.com/actions/deploy-pages

snowman2:
1 & 2 sound worthwhile. 3 would remove the documentation history that is presently available.

greglucas:
@snowman2 I just opened up a PR for (1). I think you would have to do (2) because it would be rewriting history.

snowman2:
2 is complete.

PR #1620, opened 2026-08-11
CI: stop publishing Sphinx doctree cache to gh-pages

greglucas:
Removes the unused `.doctrees` sphinx content from being deployed (saves ~27MB / deploy).


greglucas:
Addresses but does not close #1618

PR #1624, opened 2026-08-21
Fix build CI pulling in free-threaded python

### LINKED ISSUE #1623: No pyproj nightly wheels uploaded in 26 days
The most recent `pyproj` nightly wheels on the [`scientific-python-nightly-wheels`](https://anaconda.org/scientific-python-nightly-wheels) channel were uploaded **26 days ago**:

- `pyproj`: 2026-07-25

Wheels are [removed after 30 days](https://github.com/scientific-python/upload-nightly-action#artifact-cleanup-policy-at-the-scientific-python-nightly-wheels-channel), so unless a new nightly is uploaded before **2026-08-24** there will be no wheels left on the channel at all, and downstream projects that test against nightlies will fail to install them.

Some things worth checking:

- Is the workflow that builds and uploads the nightly wheels failing?
- Was its scheduled run [disabled by GitHub](https://docs.github.com/en/actions/how-tos/manage-workflow-runs/disable-and-enable-a-workflow) after a period of repository inactivity?
- Has there simply been nothing to build? Consider uploading on a fixed cadence even when nothing has changed, to keep the channel populated for downstream users.

This issue was opened automatically from [scientific-python/upload-nightly-action](https://github.com/scientific-python/upload-nightly-action), and will be closed automatically once a new wheel is uploaded.

djhoese:
I can't find all the discussions on this at the moment but I think conda/conda-forge sometimes accidentally pulls in the free-threaded python build instead of the normal GIL python. This is my first guess at why `twine` is not finding `pkg_resources` which should be a dependency of a dependency iirc. This should fix #1623 at least as a quick fix.

 - [x] Closes #1623
 - [ ] Tests added

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -32,6 +32,7 @@ jobs:
           environment-name: sdist_env
           create-args: >-
             python-build
+            python-gil
             twine>=1.12
             cython
             proj=${{ env.PROJ_VERSION }}


snowman2:
Thank @djhoese 👍

PR #1626, opened 2026-09-03
BUG: Fix use-after-free when ContextManager frees a context on anothe…

### LINKED ISSUE #1622: Per-thread PJ_CONTEXT freed by threading.local destructor while OS-thread TSS keeps the pointer (use-after-free under embedders with transient thread states)
### Summary

pyproj stores its per-thread `PJ_CONTEXT*` in **OS-thread TSS** (`PyThread_tss_*`), but hands cleanup to a **`threading.local`** `ContextManager` whose `__dealloc__` calls `proj_context_destroy` (`pyproj/_context.pyx`, ~lines 178-219 in 3.7.2). Those are two different lifetime domains: `threading.local` contents die with the **`PyThreadState`**, TSS dies with the **OS thread**. Any embedder that creates and destroys thread states on long-lived native threads — the documented `PyGILState_Ensure`/`PyGILState_Release` pattern, where the outermost release deletes the thread state — frees the context between calls while TSS retains the dangling pointer. The next pyproj use **on the same OS thread** hands the freed context to `proj_create` -> segfault.

This is legal C-API embedder behavior; we hit it in production under the Sail query engine, which attaches/detaches thread states around each Python UDF call (lakehq/sail#2456, engine-side change lakehq/sail#2457).

### Native stack (identical across repeated crashes)

```
EXC_BAD_ACCESS (freed-memory address)
  libproj  getDBcontext(pj_ctx*)     <- freed PJ_CONTEXT from TSS
  libproj  proj_create
  pyproj   _crs.CRS.__init__
```

Deterministic signature: with an N-thread native pool, the crash lands on the first *reuse*, i.e. after exactly N calls — call 1 per thread creates the context, its thread state dies, call 2 dereferences the freed pointer. The lazy `_local._crs` re-materialization path (`crs.py` `_crs` property) crashes the same way once construction is cached.

### Repro

Validated end-to-end via Sail (a ~25-line script in lakehq/sail#2456: `applyInPandas` whose body only does `pyproj.CRS("EPSG:32633")`; crashes on the second wave of groups, 100% reproducible). The pure-C-API equivalent — on a native thread, loop { `PyGILState_Ensure`; call `pyproj.CRS(...)`; `PyGILState_Release` } — should reproduce without Sail; happy to build that into a self-contained reproducer if useful.

### Workaround (validated)

Pinning the `ContextManager` in a process-global dict (so `__dealloc__` never runs) eliminates the crash entirely, at the cost of one leaked context per thread.

### Suggested fix

Keep both lifetimes in one domain, e.g.:
- clear the TSS slot in `ContextManager.__dealloc__` before `proj_context_destroy`, so a later call on the same OS thread re-creates instead of reusing a freed pointer; or
- drop the TSS mirror and key the context off `threading.local` alone.

Environment: pyproj 3.7.2 (wheel, PROJ 9.5.x / `libproj...so.25.9.5.1`), CPython 3.12.3, observed on Linux x86_64 and macOS arm64.

Troppy2:
This pull request refactors thread-local storage for PROJ contexts in pyproj/_context.pyx. It removes the low-level C thread-specific storage (TSS) mirror in favor of relying solely on Python's threading.local, which keeps context lookup and context lifetime in the same domain and avoids lifetime issues with OS threads.

PR #1541 partially addressed this by clearing the TSS slot in ContextManager.__dealloc__, but that only clears the slot on the thread that happens to run __dealloc__, not necessarily the context's owning thread. When a PROJ-backed object (e.g. a CRS, Datum, or Transformer) is created on one thread and its last reference is dropped on another, __dealloc__ runs on the wrong thread: it clears that thread's TSS slot instead of the owner's, leaving the owner's slot pointing at a freed context. This PR removes the TSS mirror entirely so there is no second slot to leave dangling.

Context management changes:

Removed all usage of PyThread_tss_* functions and CONTEXT_THREAD_KEY — no more manual C-level thread-local storage.
pyproj_context_create() now looks up the context through threading.local alone.
ContextManager.__dealloc__ no longer touches TSS; it only destroys the context.

 - [x] Closes #1622
 - [ ] Tests added

### DIFF
--- a/pyproj/_context.pyx
+++ b/pyproj/_context.pyx
@@ -203,12 +203,6 @@ cdef PJ_CONTEXT* pyproj_context_create() except *:
     """
     global _CONTEXT_MANAGER_LOCAL
 
-    # The context is looked up through threading.local alone, so the
-    # lookup and the ContextManager that destroys the context share one
-    # lifetime (the PyThreadState). A previous TSS-based lookup lived in
-    # a separate lifetime (the OS thread) and could be left pointing at
-    # an already-destroyed context.
-    # see: https://github.com/pyproj4/pyproj/issues/1622
     cdef ContextManager context_manager = _CONTEXT_MANAGER_LOCAL.context_manager
     cdef PJ_CONTEXT* pyproj_context = NULL
     if context_manager is None:
--- a/pyproj/_context.pyx
+++ b/pyproj/_context.pyx
@@ -3,7 +3,6 @@ import os
 import threading
 import warnings
 
-from cpython.pythread cimport PyThread_tss_create, PyThread_tss_get, PyThread_tss_set
 from libc.stdlib cimport free, malloc
 
 from pyproj._compat cimport cstrencode
@@ -22,8 +21,6 @@ cdef str _INTERNAL_PROJ_ERROR = None
 # global variables
 cdef bint _NETWORK_ENABLED = strtobool(os.environ.get("PROJ_NETWORK", "OFF"))
 cdef char* _CA_BUNDLE_PATH = ""
-# The key to get the context in each thread
-cdef Py_tss_t CONTEXT_THREAD_KEY
 
 
 def set_use_global_context(active=None):
@@ -177,7 +174,6 @@ cdef class ContextManager:
 
     def __dealloc__(self):
         if self.context != NULL:
-            PyThread_tss_set(&CONTEXT_THREAD_KEY, NULL)
             proj_context_destroy(self.context)
             self.context = NULL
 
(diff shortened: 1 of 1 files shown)


pyproj/_context.pyx
```diff
@@ -207,17 +203,20 @@ cdef PJ_CONTEXT* pyproj_context_create() except *:
     """
     global _CONTEXT_MANAGER_LOCAL
 
-    if PyThread_tss_create(&CONTEXT_THREAD_KEY) != 0:
-        raise MemoryError("Unable to create key for PROJ context in thread.")
-    cdef const void *thread_pyproj_context = PyThread_tss_get(&CONTEXT_THREAD_KEY)
+    # The context is looked up through threading.local alone, so the
```
snowman2: I don't think this note section is needed as the previous implementation is removed.

Troppy2: Would it be better to just remove that line? I don't know if another commit would be good, but I can do it if you want, or I think you're able to remove it—but I am not sure.

snowman2:
Thanks @Troppy2 👍

PR #1638, opened 2026-09-14
DOC: Add introductory "concepts" documentation and a glossary of terms

### LINKED ISSUE #1507: DOC: Make the getting started section a bit more readable to newcomers
[The first section in gettings started](https://github.com/pyproj4/pyproj/blob/main/docs/examples.rst) uses the abbriviation CRS, it would be nice if the abbriviation was written out.
I think it would make it more readable for the unengaged (ppl like me).

So instead of:
"Using CRS"
somthing like:
"Using Coordiante Reference Systems (CRS)"

djhoese:
As mentioned in #1507, the documentation is a little lacking when it comes to explaining the basic concepts of what pyproj works with. The acronyms are especially difficult to parse if you're not already well versed in them.

I asked Claude to take a swing at resolving #1507 and it asked me how far it should take it. I told it to go far and add new documents for a full glossary and basic concepts. I think it did a really good job. I've manually reviewed everything in this pull request even though Claude wrote it all. I made some small edits myself. I also had Claude update any implicit references to PROJ documentation with the `proj:` prefix so if in the future pyproj ever adds documents or references matching the PROJ documents these references don't unintentionally start going to the pyproj docs.

Concerns Claude had about what it wrote:

* Is it write to say that WGS84 is used in "most web APIs"? I changed it to say "many web APIs".
* "+proj=latlon" is not *equivalent* to "EPSG:4326" but rather is just "similar" to. Claude went with "similar".

Concerns I have:

* Some of the low-level details about transformations versus conversions and ISO standards were all added by Claude after I complained about the strict PROJ definitions for these terms not matching how I typically think of "transformation" (converting between CRSes versus converting between datums).

CC @jonas8217, @savjaylade84, and @mikelcatania who commented on the related issue.

 - [x] Closes #1507
 - [ ] Tests added


snowman2:
I like the organization and style. The size of the MR will require some time to review.

mackenziereading19:
Thanks for putting this together - I reviewed it mainly from the newcomer/readability angle that prompted #1507. The overall structure is much clearer. I spotted three places where the wording might be worth tightening:

1. **GeoJSON / EPSG:4326 wording** - The concepts page says GeoJSON files use WGS 84 (`EPSG:4326`). That may be slightly misleading for newcomers because GeoJSON uses WGS 84 longitude/latitude ordering, whereas EPSG:4326 officially has latitude/longitude axis order. It may be clearer to separate GeoJSON from the EPSG:4326 example here.

2. **Two absolutes in “Why it matters”** - “The same numbers in two different CRSes are two different places” and “Combining data from different sources therefore always means…” both feel a little stronger than necessary. Perhaps “can represent different places” and “make sure CRSs are known and compatible, transforming to a common CRS where necessary” would preserve the teaching point without overgeneralising.

3. **UTM as ‘120 projected CRSes’** - This seems a bit too specific. UTM is more naturally described as 60 longitudinal zones with northern/southern hemisphere variants; the actual projected CRS definitions also depend on the datum. Framing it as a projection system rather than a fixed count of CRSes may be more robust.

I also checked the conversion/transformation distinction because you flagged it in the PR description; the current explanation reads clearly to me and I don’t have a substantive change to suggest there.

djhoese:
Thanks @mackenziereading19! That's exactly what I was looking for. I've made changes based on what you said. I also had Claude change the language around "most software" to "many software packages" as I don't know if anyone is in a position to say what *most* software does.

djhoese:
And yes @snowman2 this is big. Hopefully you and others can find time to review. I think this is the type of content that could have saved me a lot of confusion 15+ years ago when I started working with projections and resampling. I had to read a paper book! :open_mouth: 

Oh great:

```
 FAILED test/test_transformer.py::test_transformer_group__download_grids - urllib.error.URLError: <urlopen error [WinError 10054] An existing connection was forcibly closed by the remote host>
```

docs/examples.rst
```diff
@@ -3,38 +3,67 @@
...
+documentation see :class:`pyproj.crs.CRS`.
 
-The :class:`pyproj.crs.CRS` class can be initialized in many different ways.
-Here are some examples of initialization.
+Creating a CRS
+~~~~~~~~~~~~~~
+
+The :class:`pyproj.crs.CRS` class can be initialized from any of the
+common ways of describing a CRS. Here are some examples; all four produce
+the same CRS, WGS 84 longitude/latitude.
```
snowman2:
```suggestion
the same CRS, WGS 84 latitude/longitude.
```

docs/concepts.rst
```diff
@@ -0,0 +1,209 @@
...
+
+EPSG is not the only authority. PROJ also knows codes from ``ESRI``, ``IGNF``,
+``OGC``, and a few others; :func:`pyproj.database.get_authorities` lists them.
+In pyproj you can pass a code as an integer (``4326``, assumed to be EPSG), a
+string (``"EPSG:4326"``), or a tuple (``("EPSG", "4326")``).
+
+To find the code for a CRS you can search the PROJ database from Python with
+:func:`pyproj.database.query_crs_info` or, for UTM zones,
+:func:`pyproj.database.query_utm_crs_info`, or browse online at
+`epsg.org <https://epsg.org/>`__ or `epsg.io <https://epsg.io/>`__.
```
snowman2: `epsg.io` has historically been outdated and caused confusion.

```suggestion
`epsg.org <https://epsg.org/>`__.
```

